PR #11, opened 2023-02-15
feat: update metrics lib to track os, arch & backend type

h4ck3rk3y:


cli/cli/helpers/user_send_metrics_election/user_send_metrics_election_tracker.go
```diff
@@ -23,6 +25,23 @@ func SendAnyBackloggedUserMetricsElectionEvent() error {
...
 
+	clusterSettingStore := kurtosis_cluster_setting.GetKurtosisClusterSettingStore()
+
+	isClusterSet, err := clusterSettingStore.HasClusterSetting()
+	if err != nil {
+		return stacktrace.Propagate(err, "Failed to check if cluster setting has been set.")
+	}
+
+	var clusterType string
+	if isClusterSet {
```
Peeeekay, line 36: Nit: this is just my preference, not a huge fan of nested if conditions, it can be simplified by doing this -
```
clusterType := resolved_config.DefaultDockerClusterName
if isClusterSet {
  clusterType, err = clusterSettingStore.GetClusterSetting()
}
if err != nil {
  return stacktrace.Propagate(err, "Cluster is set but config couldn't be fetched")
}
```

Peeeekay (approved):

cli/cli/helpers/user_send_metrics_election/user_send_metrics_election_tracker.go
h4ck3rk3y: I'd prefer not having nested if conditions as well ( especially if it was further nested ) but in this case I'd rather have the error checked and returned immediately in the same block instead of closing the parent if condition, there's some risk someone comes along and starts typing in the parent if block

I've removed the else condition, and made that the default value.

PR #15, opened 2023-02-15
feat: track enclave size after run has finished

h4ck3rk3y:


core/server/api_container/server/api_container_service.go
```diff
@@ -672,7 +672,7 @@ func (apicService ApiContainerService) runStarlarkPackageSetup(packageId string,
 }
 
 func (apicService ApiContainerService) runStarlark(parallelism int, dryRun bool, packageId string, serializedStarlark string, serializedParams string, stream grpc.ServerStream) {
-	responseLineStream := apicService.startosisRunner.Run(stream.Context(), dryRun, parallelism, packageId, serializedStarlark, serializedParams)
+	responseLineStream := apicService.startosisRunner.Run(stream.Context(), dryRun, parallelism, packageId, serializedStarlark, serializedParams, apicService.metricsClient, apicService.serviceNetwork)
```
gbouv: Can't we pass those 2 clients to the `startosisRunner` upon creationg, and leave the Run function signature as it is?

gbouv (commented):
I suggest we don't pass those 2 clients to them method itself, but rather we store a ref to them in the startosis_executor object. Thoughts?

core/server/api_container/server/api_container_service.go
h4ck3rk3y: Alright, i'll keep the Run signature as is and just pass the network/client and store them as references 👍

gbouv (approved):

PR #16, opened 2023-02-15
fix: import_module and read_file should load files from kurtosis packages (kurtosis.yml must be present in the path). enforce that only kurtosis packages (directories containing kurtosis.yml) can be run.

Peeeekay:

fix: import_module and read_file should load files from kurtosis packages (kurtosis.yml must be present in the path). enforce that only remote kurtosis packages can be run.


h4ck3rk3y:
@Peeeekay can we add a conventional commit in the description that talks about how this applies to read_files as well (if it does), will begin reviewing.

core/server/api_container/server/api_container_service.go
```diff
@@ -658,6 +659,12 @@ func (apicService ApiContainerService) runStarlarkPackageSetup(packageId string,
 		return "", interpretationError
 	}
 
+	pathToKurtosisYaml := path.Join(packageRootPathOnDisk, startosis_constants.KurtosisYamlName)
+	if _, err := os.Stat(pathToKurtosisYaml); err != nil {
```
h4ck3rk3y, line 663: Looks good

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
```diff
@@ -93,9 +98,18 @@ func (provider *GitPackageContentProvider) GetModuleContents(fileInsideModuleUrl
...
 
-	// Load the file content from its absolute path
-	contents, err := os.ReadFile(pathToFile)
+	maybeKurtosisYamlPath, err := checkIfKurtosisYamlExistsInThePathProvided(pathToFile)
 	if err != nil {
+		return "", startosis_errors.WrapWithInterpretationError(err, "Error occurred while importing or reading module '%v'", fileInsideModuleUrl)
+	}
+
+	if maybeKurtosisYamlPath == filePathToKurtosisYamlNotFound {
+		return "", startosis_errors.NewInterpretationError("kurtosis.yml is not found in the path of '%v'; files can only be imported or read from kurtosis packages. For more information, go to: %v", fileInsideModuleUrl, howImportWorksLink)
```
h4ck3rk3y: lets use the constant that you declared for "kurtosis.yml"?

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
```diff
@@ -93,9 +98,18 @@ func (provider *GitPackageContentProvider) GetModuleContents(fileInsideModuleUrl
 		return "", interpretationError
 	}
 
-	// Load the file content from its absolute path
-	contents, err := os.ReadFile(pathToFile)
+	maybeKurtosisYamlPath, err := checkIfKurtosisYamlExistsInThePathProvided(pathToFile)
 	if err != nil {
+		return "", startosis_errors.WrapWithInterpretationError(err, "Error occurred while importing or reading module '%v'", fileInsideModuleUrl)
```
h4ck3rk3y: "importing module or reading file", I guess your current framing is fine too. I wonder whether we can handle this better and talk about the specific case by getting a generic error wrapped in the read_file / import_module definitions?

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
h4ck3rk3y: kurtosis -> Kurtosis

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
```diff
@@ -278,3 +292,26 @@ func getReferenceName(repo *git.Repository, parsedURL *ParsedGitURL) (plumbing.R
...
+func checkIfKurtosisYamlExistsInThePathProvided(absPathToFile string) (string, *startosis_errors.InterpretationError) {
+	return checkIfKurtosisYamlExistsInThePathProvidedInternal(absPathToFile, os.Stat)
+}
+
+func checkIfKurtosisYamlExistsInThePathProvidedInternal(absPathToFile string, stat func(string) (os.FileInfo, error)) (string, *startosis_errors.InterpretationError) {
+	removeTrailingPathSeparator := strings.Trim(absPathToFile, string(os.PathSeparator))
+	dirs := strings.Split(removeTrailingPathSeparator, string(os.PathSeparator))
+	logrus.Debugf("Found directories: %v", dirs)
+
+	maybePackageRootPath := "/"
```
h4ck3rk3y: can we use the os.PathSeparator here?

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
```diff
@@ -278,3 +292,26 @@ func getReferenceName(repo *git.Repository, parsedURL *ParsedGitURL) (plumbing.R
...
+	for _, dir := range dirs[:len(dirs)-1] {
+		maybePackageRootPath = path.Join(maybePackageRootPath, dir)
+		pathToKurtosisYaml := path.Join(maybePackageRootPath, startosis_constants.KurtosisYamlName)
+		if _, err := stat(pathToKurtosisYaml); err == nil {
+			logrus.Debugf("Found root path: %v", maybePackageRootPath)
+			return pathToKurtosisYaml, nil
+		} else if !errors.Is(err, os.ErrNotExist) {
+			return filePathToKurtosisYamlNotFound, startosis_errors.WrapWithInterpretationError(err, "An error occurred while locating kurtosis.yml in the path of '%v'", absPathToFile)
```
h4ck3rk3y: lets use the constant you created?

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
```diff
@@ -278,3 +292,26 @@ func getReferenceName(repo *git.Repository, parsedURL *ParsedGitURL) (plumbing.R
 
 	return "", false, nil
 }
+
+func checkIfKurtosisYamlExistsInThePathProvided(absPathToFile string) (string, *startosis_errors.InterpretationError) {
+	return checkIfKurtosisYamlExistsInThePathProvidedInternal(absPathToFile, os.Stat)
+}
+
+func checkIfKurtosisYamlExistsInThePathProvidedInternal(absPathToFile string, stat func(string) (os.FileInfo, error)) (string, *startosis_errors.InterpretationError) {
```
h4ck3rk3y: Pausing review here, as this function seems a bit complicated and makes me wonder if we can achieve the same guarantees by checking if the kurtosis.yml exists when the ClonePackage or StorePackageContents call is done, can we just loudly fail and say that the package doesn't contain the Kurtosis.yml?

In clone package, we have the full package id (even nested), full url, and we can just check that sub path for the kurtosis.yml?

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
```diff
@@ -278,3 +292,26 @@ func getReferenceName(repo *git.Repository, parsedURL *ParsedGitURL) (plumbing.R
...
+	removeTrailingPathSeparator := strings.Trim(absPathToFile, string(os.PathSeparator))
+	dirs := strings.Split(removeTrailingPathSeparator, string(os.PathSeparator))
+	logrus.Debugf("Found directories: %v", dirs)
+
+	maybePackageRootPath := "/"
+	for _, dir := range dirs[:len(dirs)-1] {
```
h4ck3rk3y, line 329: we are doing all but the last path, as that can be a file?

h4ck3rk3y: so we start at /kurtosis-data/ and then slowly append every subpath till we find a path with .yml in it?

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
h4ck3rk3y: lets put a big description on top on how this guy works

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
h4ck3rk3y: worried how this is dependent on APIC having access to every directory from root,

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
Peeeekay: We are calling this method in `GetModuleContent` which is used while importing or reading files, and to make sure that `kurtosis.yml` exists. 

`ClonePackage` or `StorePackageContents` on the other hand only is for package that we want to run  - and it does loudly fail in the `api_container_service` at the moment; local actually fails much earlier in the cli itself.

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
Peeeekay: yeah, this is the dirty way to do it. I believe APIC would have the access to `/kurtosis-data` as all the package, even the package which is being run is stored there? 

yes, we can make improvements - by letting users define the dependencies in `kurtosis.yml` which isn;t super critical for today.

Peeeekay: also currently - when we import packages or read file -- all that data goes to `/kurtosis-data`; so i would imagine APIC must have access to that folder

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
Peeeekay: so `checkIfKurtosisYamlExistsInThePathProvidedInternal` is solely being used to check if `kurtosis.yml` is present while importing packages in `GetModuleContent` -- let me change the name to reflect that as well.

h4ck3rk3y: What about having this check in `atomicClone` that gets called by everything? You have the parsed URL, post clone you can check whether its a Kurtosis Package or not.

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
h4ck3rk3y: Hmm, how about going the other way around, 

/apple/bat/cat/doll/foo.star

try

/apple/bat/cat/doll/kurtosis.yml
/apple/bat/cat/kurtosis.yml
/apple/bat/kurtosis.yml

More likely we get the right Kurtosis YML and we don't hit permission issues. Though there can only be one Kurtosis YML as nesting isn't allowed (as we don't enforce it)

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
Peeeekay: i do not want to do it in `atomicClone` because that method should only be responsible for cloning the contents, not doing this validation whether `kurtosis.yml` exists.. 

also this method isn't needed for when we are running packages, because we know the file-path ( which is a user input) where kurtosis.yml needs to exists -- so no need to looping through the directories to find one.

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
Peeeekay: yeah i can do that, i was following an assumption, most likely than not we will find it in like upper subdirectories.. but hard to justify that assumption. 

If permission error you are worried about, what if I start with `/apple/bat` i.e (`/kurtosis-data/startosis-packages)` onwards -- and not `/kurtosis-data`.

No nested packages enforcement,  that is a good point, i can take a stab in one of my next prs, i can just refactor the method to enforce that as well.

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
h4ck3rk3y: fair enough, atomicClone I guess doesn't need to do that (it would be doing too much ?) also it doesn't have any packageId information, it just has the parsed URL so its quite useless to do it there.

Hmm, so when we have proper declarations, I reckon we can rewrite GetModuleContents to take in the packageId. The main problem right now I guess for which you have to do your thing is that we don't know if the package is the root repo, sub dir etc

Can we stick a TODO to clean up the logic when we have proper dependency management? We would only clone/fetch contents for things listed inside Kurotis.yml, and as we'd know the exact package id, we can just start fetching there and this best effort search would go away.

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
h4ck3rk3y: hmm, `/kurtosis-data/startosis-packages` this is fair enough. Agree can't tell if the upper/lower sub dir has it. We will also clean it all up when we have proper dependencies.

> No nested packages enforcement, that is a good point, i can take a stab in one of my next prs, i can just refactor the method to enforce that as well

perhaps create a separate ticket for this :-) No need to club the two!

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
```diff
@@ -278,3 +292,26 @@ func getReferenceName(repo *git.Repository, parsedURL *ParsedGitURL) (plumbing.R
 
 	return "", false, nil
 }
+
+func checkIfKurtosisYamlExistsInThePathProvided(absPathToFile string) (string, *startosis_errors.InterpretationError) {
```
h4ck3rk3y: lets rename this function to `checkIfFileIsInAValidPackage` or `checkIfKurtosisYmlExistsInAnyParentDir` etc

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
h4ck3rk3y: i'd change this too, following the rename of the `checkIfKurtosisYamlExistsInThePathProvided`

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
```diff
@@ -184,3 +185,71 @@ func Test_getPathToPackageRoot(t *testing.T) {
 	actual = getPathToPackageRoot(parsedGitUrl)
 	require.Equal(t, "sample/sample-package", actual)
 }
+
+func Test_checkIfKurtosisYamlExistsInThePathProvided_somewhereInTheMiddle(t *testing.T) {
```
h4ck3rk3y: rename these if you rename the function!

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
```diff
@@ -184,3 +185,71 @@ func Test_getPathToPackageRoot(t *testing.T) {
...
+		return nil, filePathAndMockReturnMap[filePath]
+	}
+
+	filePath := "/root/subdir/subdir1/folder/some_file.txt"
+	actual, err := checkIfKurtosisYamlExistsInThePathProvidedInternal(filePath, mockStatMethod)
+	require.Nil(t, err)
+	require.Equal(t, "/root/subdir/subdir1/kurtosis.yml", actual)
+}
+
+func Test_checkIfKurtosisYamlExistsInThePathProvided_packageIsSameAsWhereTheFileIs(t *testing.T) {
```
h4ck3rk3y: beautiful stuff! 👍

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
```diff
@@ -184,3 +185,71 @@ func Test_getPathToPackageRoot(t *testing.T) {
...
+			"/root/subdir/subdir1/kurtosis.yml": os.ErrNotExist,
+		}
+
+		return nil, filePathAndMockReturnMap[filePath]
+	}
+
+	filePath := "/root/subdir/some_file.txt"
+	_, err := checkIfKurtosisYamlExistsInThePathProvidedInternal(filePath, mockStatMethod)
+	require.NotNil(t, err)
+	require.ErrorContains(t, err, fmt.Sprintf("An error occurred while locating kurtosis.yml in the path of '%v'", filePath))
```
h4ck3rk3y: lets use the constant,

h4ck3rk3y (approved):

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
h4ck3rk3y: just curious if we could do something like

	mockStatMethod := func(filePath string) (os.FileInfo, error) {
		filePathAndMockReturnMap := map[string]error{
			"/root/kurtosis.yml":                os.ErrNotExist,
			"/root/subdir/kurtosis.yml":         nil,
			"/root/subdir/subdir1/kurtosis.yml": os.ErrNotExist,
		}

               pathWithYml, found := filePathAndMockReturnMap[filePath]
               if !found {
                     return stacktrace.New("tried a path that we didnt account for"), ""
               }

		return nil, pathWithYml
	}

Peeeekay: i like this, will update the code!!

PR #17, opened 2023-02-15
feat!: this a breaking change where we are deprecating PacketDelay to introduce latency in favour of PacketDelayDistribution. Instead of using packet delay, use UniformPacketDelayDistribution for constant delays or NormalPacketDelayDistribution for normally distributed latencies

Peeeekay:


core/server/api_container/server/startosis_engine/kurtosis_types/connection_config.go
```diff
@@ -13,16 +12,15 @@ import (
...
 )
 
 var (
 	PreBuiltConnectionConfigs = &starlarkstruct.Module{
 		Name: "connection",
 		Members: starlark.StringDict{
-			"BLOCKED": NewConnectionConfig(100, nil, NoPacketDelay),
-			"ALLOWED": NewConnectionConfig(0, nil, NoPacketDelay),
+			"BLOCKED": NewConnectionConfig(100, nil),
+			"ALLOWED": NewConnectionConfig(0, nil),
```
gbouv, line 23: is it the same to use constant packet delay with 0 instead of nil here? I'd say we should try to avoid using `nil` where we can, because they expose us to NPE

gbouv (approved):

core/server/api_container/server/startosis_engine/kurtosis_types/connection_config.go
Peeeekay: I currently have `if checks` to make sure that we don't see nil pointer error - used nil initially so that i can avoid printing the struct if user hasn;t passed an input.. though I can refactor it to avoid using `nil` as a FLUP though; wdyt?  @gbouv

Peeeekay:
closes: https://app.zenhub.com/workspaces/engineering-636cff9fc978ceb2aac05a1d/issues/gh/kurtosis-tech/kurtosis-private/1072

PR #20, opened 2023-02-16
docs: added more information regarding port_id

Peeeekay:
This PR closes the docs part of:  https://app.zenhub.com/workspaces/engineering-636cff9fc978ceb2aac05a1d/issues/gh/kurtosis-tech/kurtosis-private/1005

docs: added more information regarding port_id

### DIFF
--- a/docs/docs/reference/starlark-types.md
+++ b/docs/docs/reference/starlark-types.md
@@ -83,6 +83,35 @@ get_request_recipe = GetHttpRequestRecipe(
 )
 ```
 
+:::info
+Important - `port_id` field accepts user defined ID assinged to a port in service's port map while defininig `ServiceConfig`. For example, we have a service config with following port map:
+
+```
+    test-service-config = ServiceConfig(
+        ports = {
+            // "port_id": port_number
+            "http": 5000,
+            "grpc": 3000
+            ...
+        }
+        ...
+    )
+```
+
+The user defined port IDs in above port map are: `http` and `grpc`. These can be passed to create http request recipes (`GET` OR `POST`) such as:
+
+```
+    recipe = GetHttpRequestRecipe(
+        port_id = "http",
+        service_name = "service-using-test-service-config",
+        endpoint = "/ping"
+        ...
+    )
+```
+
+This above recipe when used with `request` or `wait` instruction, will make a `GET` request to a service with name `service-using-test-service-config` on port `5000` with the path `/ping`.
+:::
+
 #### PostHttpRequestRecipe
 
 The `PostHttpRequestRecipe` can be used to make `POST` requests.


docs/docs/reference/starlark-types.md
```diff
@@ -54,6 +54,34 @@ exec_recipe = ExecRecipe(
...
+        ports = {
+            "http": 5000,
+            "port_id": 3000
+            ...
+        }
+        ...
+    )
+```
+
+The user defined port IDs in above port map are: `grpc` and `port_id`. These can be passed to create http request recipes (`GET` OR `POST`) such as:
```
adschwartz: did you mean `http` rather than `grpc`?

docs/docs/reference/starlark-types.md
```diff
@@ -54,6 +54,34 @@ exec_recipe = ExecRecipe(
 
 The `HttpRequestRecipe` is used to make `HTTP` requests to an endpoint. Currently, we support `GET` and `POST` requests. (see [request][starlark-instructions-request] or [wait][starlark-instructions-wait]).
 
+:::info
+Important - `port_id` field accepts user defined ID assinged to a port in service's port map while defininig `ServiceConfig`. For example, we have a service config with following port map:
```
adschwartz: Consider this rewrite and I fixed some typos:

Note, the `port_id` field in the Recipe accepts a user defined ID (i.e. the `port_id` is not the actual port number but rather the key used to lookup the port number). The port ID to port number mappings are defined in the `ServiceConfig`. For instance, in this example we have a service config with following port map:

adschwartz: I could be overlooking something here, but does it make sense to talk about the `port_id` field when it it's not introduced until when the `GetHttpRequestRecipe` is introduced on line 74?

Peeeekay: @adschwartz - yeah was debating myself - to put either before (get and post) or after(get and post). We could put in the middle (after get, and before post); to think more about this - i could put his in the middle. wdyt

docs/docs/reference/starlark-types.md
```diff
@@ -83,6 +83,34 @@ get_request_recipe = GetHttpRequestRecipe(
 )
 ```
 
+:::info
```
leoporoli, line 86: WDYT if we move this before the `GetHttpRequestRecipe` title? because important for both types

docs/docs/reference/starlark-types.md
```diff
@@ -83,6 +83,34 @@ get_request_recipe = GetHttpRequestRecipe(
...
+        ports = {
+            "http": 5000,
+            "port_id": 3000
+            ...
+        }
+        ...
+    )
+```
+
+The user defined port IDs in above port map are: `http` and `port_id`. These can be passed to create http request recipes (`GET` OR `POST`) such as:
```
leoporoli: I suggest not use `port_id` as an example of a key because it's not semantic, I mean what is a service's port portId is it the grpc port? is it the telnet port? 
I sugget to replace this with:
```
ports = {
            // "port_id": port_number 
            "http": 5000,
            "grpc": 3000
            ...
        }
```

leoporoli (approved):

docs/docs/reference/starlark-types.md
Peeeekay: i will keep this for now; as anders mentioned we don;t mention port id until `GetHttpRequestRecipe` - so just did in between.

PR #33, opened 2023-02-21
fix: fix error message on importing/reading a package instead of a module

h4ck3rk3y:


core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
```diff
@@ -293,7 +293,8 @@ func getReferenceName(repo *git.Repository, parsedURL *ParsedGitURL) (plumbing.R
 	return "", false, nil
 }
 
-/**
+/*
+*
```
gbouv: Is that a legit go comment formatting here?? 🤔 
I personally prefer what we had before...

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
```diff
@@ -67,7 +67,7 @@ func (provider *GitPackageContentProvider) GetOnDiskAbsoluteFilePath(fileInsideP
 		return "", interpretationError
 	}
 	if parsedURL.relativeFilePath == "" {
-		return "", startosis_errors.NewInterpretationError("The relative path to file is empty for '%v'", fileInsidePackageUrl)
+		return "", startosis_errors.NewInterpretationError("Path to import '%v' needs to point to a specific file but didn't. Users can only read or import specific files and not entire packages.", fileInsidePackageUrl)
```
gbouv: nit: the phrasing looks weird to me. Though I am not a native english speaker. Something like "_The path_ '%s' needs to point to a specific file but _it_ didn't" resonates better to me

gbouv (approved):

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
h4ck3rk3y: That was auto, but reverted.

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
h4ck3rk3y: Agreed, fixed.

PR #34, opened 2023-02-21
feat: added Kurtosis Docs command

h4ck3rk3y:


cli/cli/commands/docs/docs.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/kurtosis-tech/kurtosis/cli/cli/user_support_constants"
+	"github.com/kurtosis-tech/stacktrace"
+	"os/exec"
+	"runtime"
+)
+
+const (
+	linuxOSName   = "linux"
+	macOSName     = "darwin"
+	windowsOSName = "windows"
```
gbouv, line 18: We don't communicate in our docs on how to install kurtosis CLI on windows. And as of today I don't think it is possible to do so. I'd suggest we remove windows support here to avoid confusion

cli/cli/commands/docs/docs.go
```diff
@@ -0,0 +1,50 @@
...
+// TODO refactor this to use the multi_os_command_executor after https://github.com/kurtosis-tech/kurtosis/pull/28/files is merged
+func run(_ context.Context, _ *flags.ParsedFlags, _ *args.ParsedArgs) error {
+	var args []string
+	switch runtime.GOOS {
+	case linuxOSName:
+		args = []string{"xdg-open", user_support_constants.DocumentationUrl}
+	case macOSName:
+		args = []string{"open", user_support_constants.DocumentationUrl}
+	case windowsOSName:
+		args = []string{"rundll32", "url.dll,FileProtocolHandler", user_support_constants.DocumentationUrl}
```
gbouv, line 42: Trusting you on those, it's hard to test I guess. Have you tried them on each relevant arch?

cli/cli/commands/root.go
```diff
@@ -99,6 +100,7 @@ func init() {
 	RootCmd.AddCommand(analytics.AnalyticsCmd.MustGetCobraCommand())
 	RootCmd.AddCommand(files.FilesCmd)
 	RootCmd.AddCommand(discord.DiscordCmd.MustGetCobraCommand())
+	RootCmd.AddCommand(docs.DocsCmd.MustGetCobraCommand())
```
gbouv, line 96: Suuuper nit here, but should we sort those lexicographically? 👼

cli/cli/commands/docs/docs.go
```diff
@@ -0,0 +1,50 @@
...
+	default:
+		return stacktrace.NewError("Unsupported operating system")
+	}
+	command := exec.Command(args[0], args[1:]...)
```
gbouv, line 46: What's the difference here with `exec.Command(args...)`?

cli/cli/commands/docs/docs.go
```diff
@@ -0,0 +1,50 @@
...
+	if err := command.Start(); err != nil {
+		return stacktrace.Propagate(err, "An error occurred while opening the Docs page '%v'", user_support_constants.DocumentationUrl)
```
gbouv: nit: s/Docs/docs

gbouv (approved):

cli/cli/commands/docs/docs.go
h4ck3rk3y: We have windows support on the `kurtosis discord` command, though in fairness I am the one who added it. Let's see what our Slack discussion resolves to?

cli/cli/commands/docs/docs.go
h4ck3rk3y: Not windows but otherwise yes. This is the same as `discord`

cli/cli/commands/docs/docs.go
h4ck3rk3y: That doesn't compile with `Cannot use 'args' (type []string) as the type [](string, string) or string`

cli/cli/commands/docs/docs.go
h4ck3rk3y: @gbouv I'll stick in a TODO that says "remove Windows" support if not needed and proceed ahead?

h4ck3rk3y: I have put in a TODO, I am going with the merge, will remove if we choose to Drop windows support.

PR #39, opened 2023-02-22
refactor: added name-generator to generate random nature theme names.

Peeeekay:

refactor: added name-generator to generate random nature theme names for us


common/name_generator/data_provider/nature_theme_data.go
```diff
@@ -0,0 +1,257 @@
+package data_provider
```
Peeeekay: The idea is everyone can eventually contribute to -  nature theme `nouns` and `adjectives`

Peeeekay: In next PR, will divide them into places, objects etc. as per project requirements 
https://kurtosistech.slack.com/archives/C04QGC9EAE4/p1676492287388239

common/name_generator/nature_theme_names.go
```diff
@@ -0,0 +1,24 @@
+package name_generator
+
+import "github.com/kurtosis-tech/kurtosis/common/name_generator/data_provider"
+
+// GenerateNatureThemeNameForArtifacts - generate a nature theme name
+// we read the data everytime we call this method without consistent ordering
+// TODO: use singleton pattern to access data from data provider once
```
Peeeekay: ATM: every time this method is called the order of data may change.

Peeeekay: this method will be called from `default_service_network`

common/name_generator/data_provider/nature_theme_data.go
```diff
@@ -0,0 +1,257 @@
...
+		"warm":       true,
+		"watchful":   true,
+		"weathered":  true,
+		"whispering": true,
+		"wild":       true,
+		"winter":     true,
+		"wispy":      true,
+		"yearning":   true,
+		"young":      true,
+		"zealous":    true,
```
gbouv: Oh wow... Let's extract those huge lists into separate resource files please

common/name_generator/data_provider/nature_theme_data.go
```diff
@@ -0,0 +1,257 @@
+package data_provider
+
+// GetNatureThemeNounsData it is a set so that we do not have duplicate keys
+// TODO: filter out the nouns into places and objects and animals
+// TODO: remove potential insensitive words
```
gbouv: How are we generating enclave names today? Do we have similar sets somewhere in the codebase?

common/name_generator/data_provider/nature_theme_data.go
```diff
@@ -0,0 +1,257 @@
...
+func GetNatureThemeNounsData() []string {
+	data := map[string]bool{
+		"aurora":     true,
```
gbouv: Did you take those words from somewhere? Or how did you generate them?

common/name_generator/data_provider/nature_theme_data.go
```diff
@@ -0,0 +1,257 @@
...
+		"weathered":  true,
+		"whispering": true,
+		"wild":       true,
+		"winter":     true,
+		"wispy":      true,
+		"yearning":   true,
+		"young":      true,
+		"zealous":    true,
+	}
+	return convertMapSetToStringArray(data)
```
gbouv: What's the purpose of declaring a local set and immediately converting it to a list here? Can't we use a list straight?

common/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/kurtosis-tech/kurtosis/common
+
+go 1.18
+
+require github.com/stretchr/testify v1.8.1
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
```
gbouv: I don't see any import of those packages. Did you run `go mod tidy`? Do we need all those?

common/name_generator/name_generator.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+var randomNameGenerator *RandomNameGenerator
+var once sync.Once
```
gbouv: Generally speaking global variables are an anti-pattern. Those are useful to static objects, which I think is what you're doing here, but do we really need a static object for this?

I don't see why using a regular object type here would not work

common/name_generator/nature_theme_names.go
```diff
@@ -0,0 +1,24 @@
...
+// TODO: add nature theme names for enclaves etc.
+func GenerateNatureThemeNameForArtifacts() string {
```
gbouv: Why do we need those as separate objects? Can we clarify how this entire name generator will be used? It seems overly complex right, with the static object pattern, with `GeneratorArgs` which are passed as params to GenerateName, with this GenerateNatureThemeNameForArtifacts function apparently non deterministic... 
I'm a bit confused.
For a v0 of this thing, I would expect a simple object with a simple constructor `NewRandomNameGenerator()`, and a single interface like `GetRandomName() string`. That's it. I might be missing something, but it is not clear reading this PR on why we need something so complex

gbouv (commented):
Before we go that route, can we:
- Clearly define our needs and goals for this library?
- Where will it be used concretely?
- How will it be used? Do we have concurrency constraints?
Kurtosis dev moto is "flexible minimalism", let's not over engineer something if we don't need it

Taking apart the implementation comments:
- I don't like the package is named `common`
- If we show that such a library is really needed, I actually think it should not be a package, but rather its own repo

common/name_generator/data_provider/nature_theme_data.go
Peeeekay: It is done in engine module:- https://www.notion.so/kurtosistech/Autogenerate-Files-Artifact-Names-97ec8e3556764e04b9da74578b92e6fa .. I explained it in the doc. 

I decided to use a common a repo because we would need name-generator in both core and engine modules, and did not wanted to like depend on the entire engine module just to be able to generate names. I see it more as what we have errors repo. @gbouv 

I agree overall that the common repo can grow big and maybe misused; but at the same time it does play an important role from preventing us from unnecessary dependencies.

common/name_generator/nature_theme_names.go
Peeeekay: https://www.notion.so/kurtosistech/Autogenerate-Files-Artifact-Names-97ec8e3556764e04b9da74578b92e6fa  - The reason I pass arrays is because, I would need to have like GenerataArtifactName, GenerateEncalveName etc.. which uses different set of seed data to return a name for us. 

I was favouring resource files with go embed as well - but Kevin suggested to not use and use go set ( will confirm on the rationale why when he's back).. In the meanwhile, I can revert back to resource files approach. @gbouv

name-generator/adjectives.go
```diff
@@ -0,0 +1,124 @@
...
+	"wild":       true,
+	"winter":     true,
+	"wispy":      true,
+	"yearning":   true,
+	"young":      true,
+	"zealous":    true,
+}
+
+// singleton pattern - where convertMapSetToStringArray will be called only once.
+func getAdjectives() []string {
```
Peeeekay: as discussed; this make the code deterministic for both nouns and adjectives

name-generator/name_generator.go
```diff
@@ -0,0 +1,40 @@
+package name_generator
```
Peeeekay: removed commons in favour of name-generator

common/name_generator/data_provider/nature_theme_data.go
Peeeekay: discussed offline; creating a separate module `name_generator`

common/name_generator/data_provider/nature_theme_data.go
Peeeekay: discussed offline; this is to yell at us if there duplicates -- added singleton pattern to make it deterministic

common/go.mod
Peeeekay: yes

common/name_generator/nature_theme_names.go
Peeeekay: discussed offline;

common/name_generator/name_generator.go
Peeeekay: I am using a singleton pattern which I believe is good enough - as we only need one instance of `randomNameGenerator` because we should seed random number just once for pseudo-random number:- https://stackoverflow.com/questions/12321133/how-to-properly-seed-random-number-generator

In future we can use crypto library if need be, as suggested in this doc

name_generator/adjectives.go
```diff
@@ -0,0 +1,124 @@
...
+	"winter":     true,
+	"wispy":      true,
+	"yearning":   true,
+	"young":      true,
+	"zealous":    true,
+}
+
+// singleton pattern - where convertMapSetToStringArray will be called only once.
+func getAdjectives() []string {
+	once.Do(func() {
```
gbouv: Are you using the `once` object from name_generator here? Does it work? I.e. can you use the same once object for 2 different function??

name_generator/adjectives.go
```diff
@@ -0,0 +1,124 @@
...
+		adjectives = convertMapSetToStringArray(adjectivesSet)
+	})
+	return adjectives
```
gbouv: Okay, definitely not a huge fan of using a set and converting it to a list. I've been thinking about this yesterday in my evening, and what do you think of this approach:
1. We use a simple list for adjective and nouns. No convertMapToList function anywhere, no singleton pattern. Simple list, that's it
2. We ensure uniqueness within the list by having a simple unit test validating each adj / noun is present only once
3. We stick a comment at the top of each adj / noun list saying that uniqueness is ensured at testing level.

Basically using a set adds a bunch of complexity with your singleton pattern, and it seems it's not worth it at this level. Curious to have your thoughts

name_generator/name_generator.go
```diff
@@ -0,0 +1,40 @@
...
+type randomNameGenerator struct {
+	random *rand.Rand
+}
+
+type generatorArgs struct {
+	adjectives []string
+	nouns      []string
+}
+
+// getNameGenerator we use once to seen random generator once
```
gbouv: nit: s/seen/seed
Maybe expand a little bit on the singleton pattern here. Why we use it, what value it provides, pretty much what we discusseed yesterday

name_generator/name_generator_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestRandomNameGenerator_GetNameGenerator(t *testing.T) {
+	nameGenerator := getNameGenerator()
+	nameGenerator2 := getNameGenerator()
+	// testing the memory address equality
+	require.True(t, nameGenerator == nameGenerator2)
+}
```
gbouv, line 13: Can you quickly test generateName as well here? It's very simple, I'm not interested in validating it does what it is expected to do, but a unit test will ensure that future changes don't break the behaviour unexpectedly.
For example, let's say tomorrow I go to this file and I mistakenly change `%v-%v` with a typo `%v-ewnvbqourghqefk%v` and the reviewer don't catch it. It would be pretty bad for the users.
A simple test would catch those things for us immediately

name_generator/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/kurtosis-tech/kurtosis/name-generator
+
+go 1.18
+
+require github.com/stretchr/testify v1.8.1
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
```
gbouv, line 9: double tapping here, is that really used anywhere?

gbouv (commented):
Looks good! However (because there's always a however, right?), because of your ajective and noun set, you're forced to use 2 stacked singleton patterns. I think using regular lists of adjective and nouns would reaaaaally simplify the code here. I suggest an approach to ensure uniqueness in one of my comment. curious to hear your thoughts here.

name_generator/adjectives.go
Peeeekay: Wholeheartedly agreeed - was thinking about this approach too; you are correct because too many `once` probably makes the code complicated w/o much value. Great mind thinks alike :P

name_generator/go.mod
Peeeekay: When I do `go mod tidy` it adds these packages... probably used by testify; not entirely sure.

name_generator/name_generator_test.go
Peeeekay: Previously; i was doing over abstraction... you are right instead i can just have the test the `generateName` directly instead of testing the methods in `name_theme_names`

name_generator/go.mod
gbouv: Weird... Okay, np, as long as we're not pulling unnecessary libs, all good

name_generator/adjectives_test.go
```diff
@@ -0,0 +1,23 @@
...
+	for _, adj := range NOUNS {
+		freq, found := adjectivesHash[adj]
+		if found && freq == 1 {
+			duplicateArr = append(duplicateArr, adj)
+		}
+
+		adjectivesHash[adj] = adjectivesHash[adj] + 1
+	}
+
+	require.Len(t, duplicateArr, 0, "Duplicate Error: found %v twice in ADJECTIVES", duplicateArr)
```
gbouv: nit: `s/twice/multiple times`

name_generator/adjectives_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+// check for duplicates for nouns and adjectives
+func Test_noDuplicatesInAdjectives(t *testing.T) {
+	var duplicateArr []string
+	adjectivesHash := map[string]int{}
+
+	for _, adj := range NOUNS {
```
gbouv: ⚠️ ⚠️ `s/NOUNS/ADJECTIVES`

name_generator/adjectives_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+// check for duplicates for nouns and adjectives
+func Test_noDuplicatesInAdjectives(t *testing.T) {
+	var duplicateArr []string
+	adjectivesHash := map[string]int{}
```
gbouv: nit: what do you mean by `Hash`? I don't think we hash anything, this more adjective frequency map. Anyway, no need to change, was just curious what you call Hash

name_generator/nouns_test.go
```diff
@@ -0,0 +1,23 @@
+package name_generator
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+// check for duplicates for nouns and adjectives
+func Test_noDuplicatesInNouns(t *testing.T) {
```
gbouv: nit: maybe put those two test into the same file and use a common function to DRY this out a little. Not the end of the world though

gbouv (approved):
Perfect !

PR #40, opened 2023-02-22
feat: added boilerplate method to generate unique file artifact name

Peeeekay:

feat: added boilerplate method to generate unique file artifact name

### DIFF
--- a/core/server/api_container/server/service_network/default_service_network.go
+++ b/core/server/api_container/server/service_network/default_service_network.go
@@ -56,10 +56,6 @@ const (
 	exactlyOneShortenedUuidMatch = 1
 
 	singleServiceStartupBatch = 1
-
-	// TODO: this is something we can take a look in detail
-	// but we with random numbers as suffix, we should always be able to have some unique name available
-	maxFileArtifactNameRetries = 5
 )
 
 var (
@@ -898,22 +894,16 @@ func (network *DefaultServiceNetwork) GetExistingAndHistoricalServiceIdentifiers
 	return network.allExistingAndHistoricalIdentifiers
 }
 
-// GenerateUniqueFileArtifactName
+// GetUniqueNameForFileArtifact
 // TODO: It is not wired with the flow, in next PR it will be called from Interpret methods
 //  from starlark instructions like upload_file, render_templates
 // and will return unique artifact name after 5 retries, same as enclave id generator
-func (network *DefaultServiceNetwork) GenerateUniqueFileArtifactName() (string, error) {
+func (network *DefaultServiceNetwork) GetUniqueNameForFileArtifact() (string, error) {
 	filesArtifactStore, err := network.enclaveDataDir.GetFilesArtifactStore()
 	if err != nil {
 		return "", stacktrace.Propagate(err, "An error occurred while getting files artifact store")
 	}
-
-	// TODO: this will be removed in the next PR and actual name generator method will be passed in
-	generateNameClosure := func() string {
-		return ""
-	}
-
-	return generateUniqueNameForFileArtifact(filesArtifactStore, generateNameClosure, maxFileArtifactNameRetries), nil
+	return filesArtifactStore.GenerateUniqueNameForFileArtifact(), nil
 }
 
 // ====================================================================================================
@@ -1782,30 +1772,4 @@ func renderTemplateToFile(templateAsAString string, templateData interface{}, de
 		return stacktrace.Propagate(err, "An error occurred while writing the rendered template to destination '%v'", destinationFilepath)
 	}
 	return nil
-}
-
-// generateUniqueNameForFileArtifact - this method returns unique file artifact name after x max retries
-func generateUniqueNameForFileArtifact(fileArtifactStore *enclave_data_directory.FilesArtifactStore, generateNatureThemeName func() string, maxRetry int) string {
-	var maybeUniqueName string
-
-	// try to find unique nature theme random generator
-	for maxRetry >= 0 {
-		maybeUniqueName = generateNatureThemeName()
-		maxRetry = maxRetry - 1
-
-		if !fileArtifactStore.CheckIfArtifactNameExists(maybeUniqueName) {
-			return maybeUniqueName
-		}
-	}
-
-	// if unique name not found, append a random number after the last found random name
-	additionalSuffix := 1
-	maybeUniqueNameWithRandomNumber := fmt.Sprintf("%v-%v", maybeUniqueName, additionalSuffix)
-	for fileArtifactStore.CheckIfArtifactNameExists(maybeUniqueNameWithRandomNumber) {
-		additionalSuffix = additionalSuffix + 1
-		maybeUniqueNameWithRandomNumber = fmt.Sprintf("%v-%v", maybeUniqueName, additionalSuffix)
-	}
-
-	logrus.Warnf("Cannot find unique name generator, therefore using a name with a number %v", maybeUniqueNameWithRandomNumber)
-	return maybeUniqueNameWithRandomNumber
 }
--- a/core/server/api_container/server/service_network/default_service_network.go
+++ b/core/server/api_container/server/service_network/default_service_network.go
@@ -55,7 +55,10 @@ const (
 	emptyCollectionLength        = 0
 	exactlyOneShortenedUuidMatch = 1
 
-	singleServiceStartupBatch  = 1
+	singleServiceStartupBatch = 1
+
+	// TODO: this is something we can take a look in detail
+	// but we with random numbers as suffix, we should always be able to have some unique name available
 	maxFileArtifactNameRetries = 5
 )
 
@@ -910,7 +913,7 @@ func (network *DefaultServiceNetwork) GenerateUniqueFileArtifactName() (string, 
 		return ""
 	}
 
-	return generateUniqueNameForFileArtifact(filesArtifactStore.CheckIfArtifactExists, generateNameClosure, maxFileArtifactNameRetries)
+	return generateUniqueNameForFileArtifact(filesArtifactStore, generateNameClosure, maxFileArtifactNameRetries), nil
 }
 
 // ====================================================================================================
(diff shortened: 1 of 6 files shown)


core/server/commons/enclave_data_directory/files_artifact_store.go
```diff
@@ -116,6 +116,13 @@ func (store FilesArtifactStore) ListFiles() map[string]bool {
 	return artifactNameSet
 }
 
+// CheckIfArtifactExists
+// TODO: add necessary mutexes to avoid race conditions in next PR
```
Peeeekay: i need good night sleep to figure out mutexes -- trying to rationalize some edge cases -- trying to understand whether read lock will read stale data or na.

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -894,6 +895,24 @@ func (network *DefaultServiceNetwork) GetExistingAndHistoricalServiceIdentifiers
 	return network.allExistingAndHistoricalIdentifiers
 }
 
+// GenerateUniqueFileArtifactName
+// TODO: It is not wired with the flow, in next PR it will be called from Interpret methods
+// TODO: from starlark instructions like upload_file, render_templates
```
gbouv: nit: you don't need to add TODO as a prefix here, you just need to indent it with a spcae:
```
// TODO: It is not wired with the flow, in next PR it will be called from Interpret methods
//  from starlark instructions like upload_file, render_templates
```
and your golang should colour the second line as part of the above TODO

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1761,3 +1780,23 @@ func renderTemplateToFile(templateAsAString string, templateData interface{}, de
 	}
 	return nil
 }
+
+// generateUniqueNameForFileArtifact - this method returns unique file artifact name after x max retries
+func generateUniqueNameForFileArtifact(
+	checkIfFileArtifactNameExists func(artifactName string) bool,
+	generateNatureThemeName func() string,
+	maxRetry int) (string, error) {
```
gbouv: Do we need to pass in all those functions as params? I like functional programming, but we've not been using those patterns a lot so far, so I'd rather stick to a more classical logic here. Why not calling `network.enclaveDataDir.GetFilesArtifactStore()` straight here?  I know your answer will be "for unit testing", but my answer to your answer will be: "Can we use use mocks"?

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1761,3 +1780,23 @@ func renderTemplateToFile(templateAsAString string, templateData interface{}, de
...
+
+	maybeUniqueName := generateNatureThemeName()
+
+	for !checkIfFileArtifactNameExists(maybeUniqueName) && maxRetry > 0 {
```
gbouv: How did you end up choosing 5 as the max retry here? If we have a Go for the hyperscaled enclave, we might have tens of thousands of files. Although I think Waku names all their artifact, setting 5 here seems a bit arbitrary. Not a huge fan of it TBH, unless you did the math behind and can prove that 5 is the right number here.

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1761,3 +1780,23 @@ func renderTemplateToFile(templateAsAString string, templateData interface{}, de
...
+		maybeUniqueName = generateNatureThemeName()
+		maxRetry = maxRetry - 1
+	}
+
+	if maxRetry == 0 {
```
gbouv: I don't think we can fail hard here if the max retry is hit. Having a nice and clear name for artifact it seems like a "nice to have" feature that is not worth failing the entire `kurtosis run` just because we were unable to generate a name.
I think we should pick a maxRetry appropriate to our goals here. IDK how the random name generator works, but let's target for example 100K artifacts in total, and pick a maxRetry such that it won't be hit in 99% of the cases. Something like this.
And then when it's hit, let's log a big warning, but fallback to a random UUID, or something we _know_ will work, to avoid failing the entire run.

core/server/commons/enclave_data_directory/files_artifact_store.go
gbouv: Yes, let's make sure it's thread safe, in particular as we're parallelising the deployment of services. Happy to chat about the difference between a Mutex and a RWMutex if you're not clear on those, it is essential knowledge.
IDK what we're using in this artifact store, but we surely need a RWMutex. And then just use it in read mode in this function, that should be it

core/server/api_container/server/service_network/default_service_network.go
Peeeekay: discussed offline, will stick to 5 for now; as i have added additional random numbers. this is something to think about in upcoming PRs.

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1761,3 +1783,29 @@ func renderTemplateToFile(templateAsAString string, templateData interface{}, de
 	}
 	return nil
 }
+
+// generateUniqueNameForFileArtifact - this method returns unique file artifact name after x max retries
+func generateUniqueNameForFileArtifact(fileArtifactStore *enclave_data_directory.FilesArtifactStore, generateNatureThemeName func() string, maxRetry int) string {
+	var maybeUniqueName string
+
+	// try to find unique nature theme random generator
+	for maxRetry >= 0 {
```
gbouv: nit: use `for i := 0; i < maxRetry; i++ {}` instead, it avoids you having to decrease `maxRetry` in the loop.
BTW changing the value of a parameter is a bit of an anti-pattern in general. Because it needs you to think: does go passes a copy of the value of the actual value pointer to make sure the upstream value is not going to be changed behind the curtain.

Also, using the for loop above, you don't need to pass maxRetry as a param of this function, you can read it straight from the constant

core/server/commons/enclave_data_directory/files_artifact_store.go
```diff
@@ -116,6 +116,15 @@ func (store FilesArtifactStore) ListFiles() map[string]bool {
 	return artifactNameSet
 }
 
+// CheckIfArtifactNameExists - It checks whether the FileArtifact with a name exists or not
+func (store FilesArtifactStore) CheckIfArtifactNameExists(artifactName string) bool {
```
gbouv, line 149: Alternative option here: rather than adding this `CheckIfArtifactNameExists` here, what about we implement `GetNewArtifactName` at this level instead, and just call it from service_network?
With the current implementation it seems that we're leaking a bit of the file artifact mechanism into the default_service_network, for which it is not its job to manage files. Thoughts?

core/server/commons/enclave_data_directory/files_artifact_store.go
```diff
@@ -193,3 +202,18 @@ func (store FilesArtifactStore) removeFileUnlocked(filesArtifactUuid FilesArtifa
 
 	return nil
 }
+
+// method needed for testing
+func NewFilesArtifactStoreForTesting(
```
gbouv: Lets put this constructor next the the other one.

gbouv (commented):
Thoughts on moving all the logic at the fileArtifactStore and simplifying default_service_network? Otherwise things look good 👍

core/server/commons/enclave_data_directory/files_artifact_store.go
Peeeekay: Awesome suggestion!!! I should have seen this before - did some refactoring to promote S in SOLID principle and fileArtifactStore is responsible for handling nitty gritty details for handling the generation.

I am still assigning a global maxRetry variable to struct field - it just helps in unit testing  as I don't have rely on global variable and I can pass in any value for maxRetry while testing. @gbouv

core/server/commons/enclave_data_directory/files_artifact_store.go
```diff
@@ -16,21 +18,48 @@ import (
 const (
 	artifactExtension                     = "tgz"
 	maxAllowedMatchesAgainstShortenedUuid = 1
+	// TODO: this is something we can take a look in detail
+	// but we with random numbers as suffix, we should always be able to have some unique name available
+	maxFileArtifactNameRetries = 5
```
gbouv: Let's call this one `maxFileArtifactNameRetriesDefault` and the on in the struct `maxFileArtifactNameRetries` so that it's clear they are linked together, and one is the default of a value that can be customized when needed

core/server/commons/enclave_data_directory/files_artifact_store.go
```diff
@@ -16,21 +18,48 @@ import (
...
-		mutex:                      &sync.RWMutex{},
-		artifactNameToArtifactUuid: make(map[string]FilesArtifactUUID),
-		shortenedUuidToFullUuid:    make(map[string][]FilesArtifactUUID),
+		fileCache:                       newFileCache(absoluteDirpath, dirpathRelativeToDataDirRoot),
+		mutex:                           &sync.RWMutex{},
+		artifactNameToArtifactUuid:      make(map[string]FilesArtifactUUID),
+		shortenedUuidToFullUuid:         make(map[string][]FilesArtifactUUID),
+		maxRetriesToGetFileArtifactName: maxFileArtifactNameRetries,
+		//TODO: in next pr will assign this to name-generator method
+		generateNatureThemeName: nil,
```
gbouv, line 43: Curious to see how you pipe everything together here, but mayyyybe we don't have to store this as a struct attribute? let's see

core/server/commons/enclave_data_directory/files_artifact_store.go
```diff
@@ -116,6 +145,38 @@ func (store FilesArtifactStore) ListFiles() map[string]bool {
 	return artifactNameSet
 }
 
+// CheckIfArtifactNameExists - It checks whether the FileArtifact with a name exists or not
+func (store FilesArtifactStore) CheckIfArtifactNameExists(artifactName string) bool {
+	store.mutex.RLock()
+	defer store.mutex.RUnlock()
```
gbouv: @Peeeekay I believe this `CheckIfArtifactNameExists` function can be private now right? I would even say that you can inline it inside `GenerateUniqueNameForFileArtifact` directly. Basically you take the mutex, do the for loop until max retry is reached, and release the mutex.
Because here you're interacting with the mutex inside a loop, seems unnecessary

gbouv (approved):
Love it ❤️

core/server/commons/enclave_data_directory/files_artifact_store.go
Peeeekay: just allows me to mock the method for unit testing.

PR #44, opened 2023-02-22
ci: breaking change PRs get conventional titles for free

h4ck3rk3y:


.github/workflows/breaking-change-description-title-update.yml
```diff
@@ -0,0 +1,59 @@
...
+
+# cancel redundant builds
+concurrency:
+  group: "update-breaking-change-${{ github.head_ref }}"
+  cancel-in-progress: true
+
+jobs:
+  update_pr:
+    # we want this to only run on breaking-release PRs
+    if: contains(github.head_ref, 'release-please') != true && contains(github.head_ref, 'breaking-release')
```
gbouv: Do we need `== true` at the end? Or maybe it's implicit, but in this case we should use either all implicit or all explicit, a mix of the two is confusing IMO

gbouv (approved):

.github/workflows/breaking-change-description-title-update.yml
h4ck3rk3y: it is implicit, we don't have to have an `== true`. Perhaps I could just `!` the other conditions instead of match with `!=` true. Let me try.

h4ck3rk3y: !contains doesn't work, that needs to be compared to != true, contains is still implicit == true, and we have that throughout. I can compare == true to be consistent.

PR #46, opened 2023-02-22
refactor: move metrics to the CLI

h4ck3rk3y:
fixes - https://github.com/kurtosis-tech/kurtosis-private/issues/1121


cli/cli/commands/enclave/add/add.go
```diff
@@ -125,6 +126,21 @@ func run(cmd *cobra.Command, args []string) error {
 		enclaveName = enclaveIdStr
 	}
 
+	metricsClient, metricsClientCloser, err := metrics_client_factory.GetMetricsClient()
```
gbouv: Interesting, so the EnclaveAdd command is a raw cobra command. Should we make it an EngineConsumingCommand like rm and others? That would avoid this metric client creation block here I guess, but maybe there's something I'm missing

cli/cli/commands/enclave/rm/rm.go
```diff
@@ -89,7 +91,7 @@ func run(
 
 	enclaveDestructionErrorStrs := []string{}
 	for _, enclaveId := range enclaveIdsToDestroy {
-		if err := destroyEnclave(ctx, kurtosisCtx, enclaveId, shouldForceRemove); err != nil {
+		if err := destroyEnclave(ctx, kurtosisCtx, enclaveId, shouldForceRemove, metricsClient); err != nil {
```
gbouv: Any opinion on adding the call to metrics client in the run function, instead of down in `destroyEnclave`?

cli/cli/helpers/metrics_client_factory/metrics_client_factory.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/kurtosis-tech/kurtosis/cli/cli/kurtosis_config/resolved_config"
+	"github.com/kurtosis-tech/kurtosis/kurtosis_version"
+	metrics_client "github.com/kurtosis-tech/metrics-library/golang/lib/client"
+	"github.com/kurtosis-tech/metrics-library/golang/lib/source"
+	"github.com/kurtosis-tech/stacktrace"
+)
+
+const (
+	shouldFlushMetricsClientQueueOnEachEvent          = false
+	defaultSendingUserMetricsIfConfigurationIsntFound = true
```
gbouv: nit: `s/Isnt/IsNot`

cli/cli/helpers/metrics_client_factory/metrics_client_factory.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+type doNothingMetricsClientCallback struct{}
+
+func (d doNothingMetricsClientCallback) Success()          {}
+func (d doNothingMetricsClientCallback) Failure(err error) {}
```
gbouv: Out of curiosity what's this thing? Reading its name, I'm just wondering if it could be worth logging a little something in the failure case 🤷‍♂️

cli/cli/helpers/metrics_client_factory/metrics_client_factory.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/kurtosis-tech/metrics-library/golang/lib/source"
+	"github.com/kurtosis-tech/stacktrace"
+)
+
+const (
+	shouldFlushMetricsClientQueueOnEachEvent          = false
+	defaultSendingUserMetricsIfConfigurationIsntFound = true
+)
+
+type doNothingMetricsClientCallback struct{}
```
gbouv: My Java-brain tells me that this should leave in a separate file. Also, this is in sync with `File often` style guide rule.  I know it's super small, but maybe we should stick to it?

cli/cli/helpers/metrics_client_factory/metrics_client_factory.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	var sendUserMetrics bool
+	if hasConfig {
+		kurtosisConfig, err := kurtosisConfigStore.GetConfig()
+		if err != nil {
+			return nil, nil, stacktrace.NewError("An error occurred while fetching stored configuration")
+		}
+		sendUserMetrics = kurtosisConfig.GetShouldSendMetrics()
+	} else {
+		sendUserMetrics = defaultSendingUserMetricsIfConfigurationIsntFound
```
gbouv: Is the "default send metric = true" behaviour hardcoded in a different place or just here? I'm wondering if that's somethiing we should place in the library itself as a global constant accessible from everywhere. Rational is that at something point if we change this approach, it will be a pain to track down all places where we have this value. And that's the kind of thing we don't want to fail at because telling users "we don't send metric" when in fact we do send some is a pretty bad thing to do.

cli/cli/helpers/metrics_client_factory/metrics_client_factory.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	metricsClient, metricsClientCloseFunc, err := metrics_client.CreateMetricsClient(
+		source.KurtosisCLISource,
+		kurtosis_version.KurtosisVersion,
+		metricsUserId,
+		clusterType,
+		sendUserMetrics,
+		shouldFlushMetricsClientQueueOnEachEvent,
+		doNothingMetricsClientCallback{},
```
gbouv: We should have a constructor for this thing here. It adds a reason for putting it into in a separate file

gbouv (approved):
Few nits but overall looks good! Things might need discussions / further tracking:
- Should we move add to an EngineConsumingCommand?
- Should we have the "defaultSendMetric = true" constant in a centralized place rather than adhoc in the file constants? Rational explained below.

cli/cli/commands/enclave/add/add.go
h4ck3rk3y: @gbouv I agree that we should. I wanted to do it in a different PR as I already had so many changes here. Do you want me to

- do it in this PR
- ticket it and do it in the next PR?

cli/cli/helpers/metrics_client_factory/metrics_client_factory.go
h4ck3rk3y: I just copied the ones from the apic/engine, we use a different call back for the install-consent as we want to clear the backlog but normally we don't want it to do anything.

I have added a comment on top.

cli/cli/helpers/metrics_client_factory/metrics_client_factory.go
h4ck3rk3y: The library itself produces two different clients a do nothing client / a segment client based on the passed boolean. For now the CLI is the only customer of the Library. I'll make that boolean available in a variable within the CLI and stick in a TODO to move it to the CLI?

cli/cli/commands/enclave/add/add.go
h4ck3rk3y: Added.

PR #47, opened 2023-02-22
fix: better errors when enclave cleaning fails

h4ck3rk3y:


engine/server/engine/enclave_manager/enclave_manager.go
```diff
@@ -290,49 +291,31 @@ func (manager *EnclaveManager) Clean(ctx context.Context, shouldCleanAll bool) (
 	// TODO: Refactor with kurtosis backend
 	resultSuccessfullyRemovedArtifactsIds := map[string]map[string]bool{}
```
leoporoli: I know this comment is out of this scope, but I don't understand why it is not just `map[string]bool{}` it seems that the `phase` part is not in use

leoporoli: do you think that it's worth it to change it on this PR ?

leoporoli (approved):

engine/server/engine/enclave_manager/enclave_manager.go
h4ck3rk3y: good shout will fix

PR #55, opened 2023-02-24
feat: print users analytics id

h4ck3rk3y:


leeederek:
@h4ck3rk3y with this PR, a user can simply run `kurtosis analytics id` to print their metric user id? One of the commits says that there is an option to create an analytics id. Can you elaborate on what that flow looks like?

h4ck3rk3y:
@leeederek this is simply printing analytics id, that commit is wrong!

h4ck3rk3y:
closing this in favor of @leoporoli working on this

cc @leeederek

PR #57, opened 2023-02-24
docs: add docs about `kurtosis analytics`

leeederek:
To be merged alongside #55

docs: add docs about


docs/docs/reference/cli/analytics-disable.md
```diff
@@ -0,0 +1,13 @@
+---
+title: analytics disable
```
h4ck3rk3y, line 2: curious why you chose two different pages instead of one?

docs/docs/reference/cli/analytics-id.md
```diff
@@ -0,0 +1,15 @@
+---
+title: analytics id
+sidebar_label: analytics id
+slug: /analytics-id
+---
+
+In addition to log outputs and screenshots, [product analytics metrics](../../explanations/metrics-philosophy.md) may be useful for our team when debugging issues and shipping fixes for bugs.
```
h4ck3rk3y, line 7: do we want to do it this way or a different way (kurtosis id?)

if we are doing it this way this is arlight.

h4ck3rk3y (approved):

docs/docs/reference/cli/analytics-id.md
h4ck3rk3y: pending https://github.com/kurtosis-tech/kurtosis/pull/55 getting merged

docs/docs/reference/cli/analytics-disable.md
leeederek: Good question. I re-organized the CLI docs from a single page with 10+ commands to a directory with sub-pages for _each_ command. Doing so helps with readability and link-ability (we can now link to specific commands directly rather than just a single large page). 

Since `kurtosis analytics disable` and `kurtosis analytics enable` are two distinct commands, I've broken them apart into 2 different pages to keep this consistent with the other commands.

Here is what the new CLI docs look like: https://docs.kurtosis.com/cli

docs/docs/reference/cli/analytics-id.md
leeederek: Sorry, I don't understand what "way" you are referring to when you say the following:
> do we want to do it this way or a different way (kurtosis id?)

I care about a few things, hopefully sharing this will help add context on why I framed the docs this way:
1. That a user can easily print their metrics ID (blocked on #55)
2. That a user understands what their metrics ID is used for and why its important (docs to the metrics-philosophy)
3. That a user knows why it can be beneficial (for us and for them) to share this information with us when submitting feedback and that it is optional for them to do so

I believe that what I wrote here in the docs help us with achieving (2) and (3) above.

mieubrisse (approved):
Really nice job Derek

PR #59, opened 2023-02-25
feat: introduce nature themed name for enclaves

Peeeekay:

feat: added nature theme name for engines


engine/server/engine/enclave_manager/enclave_name_generator.go
```diff
@@ -20,50 +19,44 @@ var (
...
 	retries uint16,
-) (string, error) {
+) string {
 	retriesLeft := retries - 1
 
-	randomEnclaveName := enclaveIdGenerator.generator.Generate()
+	randomEnclaveName := enclaveIdGenerator.generateNatureThemeName()
+	isIdInUse := isEnclaveNameInUse(randomEnclaveName, allCurrentEnclaves)
 
-	validationError := validateEnclaveName(randomEnclaveName)
```
Peeeekay, line 45: Don't think so we need this anymore, as we are using custom name generation - which will be valid.

engine/server/engine/enclave_manager/enclave_name_generator.go
```diff
@@ -20,50 +19,44 @@ var (
...
 type enclaveNameGenerator struct {
-	generator namegenerator.Generator
+	generateNatureThemeName func() string
 }
 
 func GetEnclaveNameGenerator() *enclaveNameGenerator {
-	// NOTE: We use a 'once' to initialize the enclaveNameGenerator because it contains a seed,
-	// and we don't ever want multiple enclaveNameGenerator instances in existence
+	// NOTE: Do not think we still need multiple instances of this struct; though
+	// open not having once here as well.
```
gbouv: @Peeeekay do we? I believe the singleton pattern you implemented in the name generator should be enough here right?
Pushing it even further, why do we even need the `enclaveNameGenerator` struct _at all_? Can't we just call the static function `GenerateNatureThemeNameForEngine` straight from a static `GetRandomEnclaveNameWithRetries` and forget about this once.Do?

engine/server/engine/enclave_manager/enclave_name_generator.go
```diff
@@ -20,50 +19,44 @@ var (
...
 func (enclaveIdGenerator *enclaveNameGenerator) GetRandomEnclaveNameWithRetries(
 	allCurrentEnclaves map[enclave.EnclaveUUID]*enclave.Enclave,
 	retries uint16,
-) (string, error) {
+) string {
 	retriesLeft := retries - 1
 
-	randomEnclaveName := enclaveIdGenerator.generator.Generate()
+	randomEnclaveName := enclaveIdGenerator.generateNatureThemeName()
+	isIdInUse := isEnclaveNameInUse(randomEnclaveName, allCurrentEnclaves)
```
gbouv, line 17: You're mixing the concept of ID and NAME here. pick one and stick to it

engine/server/engine/enclave_manager/enclave_name_generator.go
gbouv: Agreed, especially as we're validating it downstream in the codebase anyway

engine/server/engine/enclave_manager/enclave_name_generator.go
```diff
@@ -20,50 +19,44 @@ var (
...
 	}
 
-	return randomEnclaveName, nil
+	// if no retry is left, then use the last nature theme name with NUMBER at the end
+	number := 1
+	randomEnclaveNameWithNumber := fmt.Sprintf("%v-%v", randomEnclaveName, number)
+	for isEnclaveNameInUse(randomEnclaveNameWithNumber, allCurrentEnclaves) {
+		number = number + 1
+		randomEnclaveNameWithNumber = fmt.Sprintf("%v-%v", randomEnclaveName, number)
+	}
```
gbouv: You can avoid the repetition of the Sprintf here I think:
```go
randomEnclaveNameWithNumber := randomEnclaveName
for int i := 1; isEnclaveNameInUse(randomEnclaveNameWithNumber, allCurrentEnclaves); i++ {
    randomEnclaveName = fmt.Sprintf("%v-%v", randomEnclaveName, i)
}
return randomEnclaveNameWithNumber
```
Downside is that you test that the first randomEnclaveNameWithNumber twice, but I think it's worth it in this case

engine/server/engine/enclave_manager/enclave_name_generator_test.go
```diff
@@ -7,13 +7,56 @@ import (
...
+	mockNatureThemeName := func() string {
+		timesCalled++
+		// found unique on the last try
+		if timesCalled == 4 {
+			return "unique-name"
+		}
 
-	randomEnclaveId, err := enclaveIdGeneratorObj.GetRandomEnclaveNameWithRetries(noCurrentEnclave, retries)
-	require.NoError(t, err)
+		if timesCalled%2 == 0 {
```
gbouv: Can you add a comment explaining how this mocked function behaves?

gbouv: Is that `%2` necessary here? Since you're increasing timesCalled at the beginning, this check will be run against `timesCalled = 1 | 2 | 3` because at 4 it will return in the above if. So, basically the `%2` is hit only for `2`. So, I believe you can just do `== 2` here. This is also simpler to understand

engine/server/engine/enclave_manager/enclave_name_generator_test.go
```diff
@@ -7,13 +7,56 @@ import (
...
+			return "name-exists"
+		}
+
+		return "not-unique"
+	}
+
+	enclaveIdGeneratorObj := enclaveNameGenerator{generateNatureThemeName: mockNatureThemeName}
+	randomEnclaveId := enclaveIdGeneratorObj.GetRandomEnclaveNameWithRetries(currentEnclavePresent, retries)
+	require.NotEmpty(t, randomEnclaveId)
+	require.Equal(t, "unique-name", randomEnclaveId)
```
gbouv: Maybe put all those string into const at the top. Esp. as you're using them in different places

name_generator/nouns.go
```diff
@@ -1,105 +1,135 @@
...
+	"atoll",
+	"cove",
+	"sinkhole",
+	"beach",
+	"reef",
+	"outcrop",
+	"sand",
+}
+
+var FILE_ARTIFACT_NOUNS = []string{
```
gbouv, line 80: Since you added a `// places` comment for the other, mind adding one for this one?

name_generator/nouns.go
```diff
@@ -1,105 +1,135 @@
 package name_generator
 
-var NOUNS = []string{
-	"fern",
-	"night",
-	"shark",
+// places
+var ENGINE_NOUNS = []string{
```
gbouv: Same here, why `ENGINE` here? Isn't it for enclaves?

name_generator/nature_theme_names.go
```diff
@@ -6,7 +6,21 @@ package name_generator
...
-		nouns:      NOUNS,
+		nouns:      FILE_ARTIFACT_NOUNS,
+	}
+
+	nameGenerator := getNameGenerator()
+	natureThemeName := nameGenerator.generateName(args)
+	return natureThemeName
+}
+
+// GenerateNatureThemeNameForEngine - generates nature theme name for file artifacts
```
gbouv: "GenerateNatureThemeNameForEngine - generates nature theme name for file artifacts"
1. I'm confused by `Engine`. Isn't it for enclaves?
2. ... generates nature theme name for ... enclaves?

gbouv (approved):
Some minor comments on details. High level code looks good!

name_generator/nature_theme_names.go
Peeeekay: :( good catch.

engine/server/engine/enclave_manager/enclave_name_generator.go
Peeeekay: mhm; yeah you are right on this. No need for this struct - static method is good enough.

PR #61, opened 2023-02-27
fix: preserve cli provided ordering of completions throughout shells

h4ck3rk3y:
we sort the completions, Shell resorts it for us with the new version and this directive we can keep our ordering (names followed by uuids)


h4ck3rk3y:
Closes kurtosis-tech/kurtosis-private/issues/958

h4ck3rk3y:
@gbouv the last cobra release was on the 22nd of October, not sure when the next version comes out so using latest commit.

gbouv (approved):

PR #62, opened 2023-02-27
feat: fix the output of kurtosis enclave dump

h4ck3rk3y:
feat: make the output dir optional on kurotsis enclave dump

### DIFF
--- a/container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go
+++ b/container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go
@@ -208,10 +208,8 @@ func (provider *dockerEnclaveObjectAttributesProviderImpl) ForUserServiceContain
 	privatePorts map[string]*port_spec.PortSpec,
 ) (DockerObjectAttributes, error) {
 	name, err := provider.getNameForUserServiceContainer(
-		[]string{
-			string(serviceName),
-			string(serviceUuid),
-		},
+		serviceName,
+		serviceUuid,
 	)
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "An error occurred creating the user service Docker container name object")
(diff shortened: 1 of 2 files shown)


h4ck3rk3y:
image

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go
```diff
@@ -440,6 +439,19 @@ func (provider *dockerEnclaveObjectAttributesProviderImpl) getNameForEnclaveObje
 	return name, nil
 }
 
+// Gets the name of the service container (service_name--service_uuid)
+func (provider *dockerEnclaveObjectAttributesProviderImpl) getNameForUserServiceContainer(elems []string) (*docker_object_name.DockerObjectName, error) {
```
gbouv: Don't we want to replace `elems []string` with `serviceName string, serviceUuid ServiceUUID` here to have a more explicit function signature?

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go
```diff
@@ -203,14 +202,14 @@ func (provider *dockerEnclaveObjectAttributesProviderImpl) ForApiContainer(
 }
 
 func (provider *dockerEnclaveObjectAttributesProviderImpl) ForUserServiceContainer(
-	serviceId service.ServiceName,
+	serviceName service.ServiceName,
```
gbouv, line 205: Since you're updating this file, would you mind updating the outdated variable names below as well? I see a bunch of serviceGUIDs and so on

gbouv (approved):
We will still be able to link together a user service and its sidecar right? I guess so since I see service UUID being in both places, but just want to make sure. Would be nice to keep the user service and its sidecar close to each other, but I don't think it's the case right now anyway so I think it's fine

h4ck3rk3y:
@gbouv I think they're linked & also the network partitioning test continues to pass.

PR #74, opened 2023-02-28
fix: stamp enclave uuid at the end of enclave objects

h4ck3rk3y:
this allows for the kurtosis-api to follow scheme similar to services (service_name--service_uuid) but as kurtosis-api doesn't have a service uuid we do kurtosis-api--enclave_uuid


h4ck3rk3y:
image

h4ck3rk3y:
inspired by https://kurtosistech.slack.com/archives/C032ZB39AHH/p1677518658633209

gbouv (approved):

PR #77, opened 2023-02-28
refactor: Cleanup unused code and refactor Starlark engine

gbouv:


core/server/api_container/server/startosis_engine/startosis_interpreter.go
```diff
@@ -281,5 +281,6 @@ func newStarlarkThread(threadName string) *starlark.Thread {
 		Print:      makePrintFunction(),
 		Load:       makeLoadFunction(),
 		OnMaxSteps: nil,
+		Steps:      0,
```
h4ck3rk3y, line 284: nit: should we take this constant out?

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/subnetwork_tuple_parser.go
```diff
@@ -0,0 +1,21 @@
...
+	"go.starlark.net/starlark"
+)
+
+func ParseSubnetworks(subnetworkArgName string, subnetworksTuple starlark.Tuple) (service_network_types.PartitionID, service_network_types.PartitionID, *startosis_errors.InterpretationError) {
+	subnetworksStr, interpretationErr := kurtosis_types.SafeCastToStringSlice(subnetworksTuple, subnetworkArgName)
+	if interpretationErr != nil {
+		return "", "", interpretationErr
+	}
+	if len(subnetworksStr) != 2 {
+		return "", "", startosis_errors.NewInterpretationError("Subnetworks tuple should contain exactly 2 subnetwork names. %d was/were provided", len(subnetworksStr))
```
h4ck3rk3y: nit: maybe just make it "were" or handle the ==1 case and make it was?

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/subnetwork_tuple_parser.go
```diff
@@ -0,0 +1,21 @@
...
+	}
+	subnetwork1 := service_network_types.PartitionID(subnetworksStr[0])
```
h4ck3rk3y, line 22: nit: constants

core/server/api_container/server/startosis_engine/kurtosis_instruction/kurtosis_print/kurtosis_print.go
```diff
@@ -105,7 +110,13 @@ func (instruction *PrintInstruction) Execute(_ context.Context) (*string, error)
 }
 
 func (instruction *PrintInstruction) String() string {
-	return shared_helpers.CanonicalizeInstruction(PrintBuiltinName, instruction.args, instruction.getKwargs())
+	var stringifiedArguments []string
+	for _, arg := range instruction.args {
+		stringifiedArguments = append(stringifiedArguments, builtin_argument.StringifyArgumentValue(arg))
+	}
+	stringifiedArguments = append(stringifiedArguments, fmt.Sprintf("%s=%q", endArgName, instruction.end))
```
h4ck3rk3y, line 117: can we put a tiny comment on top to explain what this does and what the separatorArgName and endArgName are

h4ck3rk3y (approved):

core/server/api_container/server/startosis_engine/startosis_interpreter.go
gbouv: I don't think so. This constant is just the default for struct instantiation to make the linter happy about exhaust struct. We have a bunch of places where we instantiate struct with inline constants like this.
I think if the constant was != 0, we could potentially discuss putting it a named constant with an explicit name on why we're not using default, but here I think it's not necessary

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/subnetwork_tuple_parser.go
gbouv: I don't really treat those as constants neither TBH. There are indices, and I don't think putting them into their own constant like "sliceFirstElementIndex" would really improve code robustness here. Happy to discuss more in depth if you feel we should extract those to constants, otherwise I'll leave them as they are

core/server/api_container/server/startosis_engine/kurtosis_instruction/kurtosis_print/kurtosis_print.go
gbouv: This goes away entirely in https://github.com/kurtosis-tech/kurtosis/pull/80. Will skip the comment

PR #78, opened 2023-02-28
docs: included the service_name parameter for the exec, request, and wait instructions

leoporoli:
@leeederek Context here: https://www.notion.so/kurtosistech/Design-Embedding-wait-in-the-add_service-instruction-79cc8089f00d4b6a96e31611cfa3a500#e5361bfb861b4d5c81b31c7f0cfb4e7f


docs/docs/reference/starlark-instructions.md
```diff
@@ -158,18 +158,31 @@ plan.assert(
 
 The `exec` instruction on the [`plan`][plan-reference] object executes commands on a given service as if they were running in a shell on the container.
 
+:::caution
+
+The previous `exec` instruction signature `exec(recipe=recipe)` is also accepted but will be deprecated soon, we suggest users to use this new one.
```
leeederek:
```suggestion
The previous `ExecRecipe` type `ExecRecipe(service_name = "my_service", command = ["echo", "Hello, world"])` is also accepted but will be deprecated soon, we suggest users to use this new one where the `service_name` argument is passed in via the `execRecipe` instruction signature instead. 
```

I think this makes more sense to me. What do you think?

leeederek (commented):
Just a bit confused about the wording around ExecRecipe. The caution tip about the wait instruction makes sense to me though!

docs/docs/reference/starlark-instructions.md
leoporoli: yeap! makes sense

leeederek (approved):
Makes more sense to me now. Thanks for taking the time to clarify for me what is actually happening with this change.

PR #81, opened 2023-02-28
feat: added the `id` flag in the `analytics` CLI command which allow users to get the `analytics ID` in an easy way

leoporoli:
fixes https://github.com/kurtosis-tech/kurtosis-private/issues/1130


docs/docs/reference/cli/cli.md
```diff
@@ -20,6 +20,7 @@ Kurtosis supports tab-completion, and we strongly recommend [installing it][addi
 
 ### Toggle Analytics
 On installation Kurtosis enables anonymized [analytics][metrics-philosophy-reference] by default. You can toggle this functionality simply by running:
+We identify every user's machines with a hash value of the `unique machine id of most host OS's`, which we call the `user's analytics ID`, and this allow us to analyze the user experience flow.
```
leeederek:
```suggestion
A metrics user ID is a unique hash generated for every Kurtosis user.
```

I would simplify it to this, and use: `metrics user ID` instead of `user's analytics iD`

docs/docs/reference/cli/cli.md
```diff
@@ -33,6 +34,13 @@ kurtosis analytics disable
 
 if you would prefer not to.
 
+```bash
+kurtosis analytics id
+```
+
+if you want to see your  `user's analytics ID`
```
leeederek:
```suggestion
if you want to see your  `metrics user ID`
```

Peeeekay (approved):

PR #82, opened 2023-03-01
feat: integrate nature theme based name to cli (render template and store service) for file artifacts

Peeeekay:

feat: integrate nature theme named to render_template and store_service

### DIFF
--- a/cli/cli/commands/files/storeservice/storeservice.go
+++ b/cli/cli/commands/files/storeservice/storeservice.go
@@ -35,17 +35,20 @@ const (
 	kurtosisBackendCtxKey = "kurtosis-backend"
 	engineClientCtxKey    = "engine-client"
 
-	starlarkTemplate = `
-CURRENT_TIME_STR = str(time.now().unix)
-ARTIFACT_NAME = "cli-stored-artifact-" + CURRENT_TIME_STR
+	starlarkTemplateWithArtifactName = `
 def run(plan, args):
-	name = ARTIFACT_NAME
-	if args.name != "":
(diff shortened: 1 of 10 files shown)


Peeeekay:
I am basically integrating the name themed auto-generation of file-artifacts name to the CLI.

Let me know if you have any questions, but I was stuck between either updating the current response from enclaveCtx methods like `UploadFiles` etc, because we only return UUIDs ( was reluctant to break the contract for SDK apis) or create new a RPC method which I can call to get a unique nature themed name.

I went with the latter as it I wanted to make this PR backward compatible as this particular feature should not be breaking imo opinion.

gbouv (commented):
Not a big fan of that implementation TBH. I don't think we want to expose such a specialized and narrow functionality to our API. What about instead we make the artifactName arg option on our UploadFiles API function, and handle the mechanics of generating a unique one entirely in the backend?

cli/cli/commands/files/storeservice/storeservice.go
```diff
@@ -35,17 +35,20 @@ const (
 	kurtosisBackendCtxKey = "kurtosis-backend"
```
gbouv, line 35: nit: can we rename this file and package `storeservicefile` maybe?

gbouv (approved):

cli/cli/commands/files/storeservice/storeservice.go
Peeeekay: i will leave this for now, as the command itself is just: `storeservice`

PR #89, opened 2023-03-02
feat!: integrate nature theme name to upload command cli

Peeeekay:

feat!: integrate nature theme name to upload command cli
refactor!: Migrate Kurtosis Print instruction to Starlark framework. This restrict the use of `print` to a single argument only. (#80)


Peeeekay:
This is the breaking change for `upload` cli command. Did not do it for storeweb file command as it's needs to be deprecated: https://github.com/kurtosis-tech/kurtosis-private/issues/837

api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -215,20 +215,20 @@ func (enclaveCtx *EnclaveContext) GetServices() (map[services.ServiceName]servic
...
 	}
 
 	args := binding_constructors.NewUploadFilesArtifactArgs(content, artifactName)
 	response, err := enclaveCtx.client.UploadFilesArtifact(context.Background(), args)
 	if err != nil {
-		return "", stacktrace.Propagate(err, "An error was encountered while uploading data to the API Container.")
+		return "", "", stacktrace.Propagate(err, "An error was encountered while uploading data to the API Container.")
 	}
-	return services.FilesArtifactUUID(response.Uuid), nil
+	return services.FilesArtifactUUID(response.Uuid), services.FileArtifactName(response.GetName()), nil
```
gbouv: The discrepancy between `response.Uuid` and `response.GetName()` looks weird here. to my understanding the main difference is that getting the attribute can return nil if the attribute is optional, whereas the function can't. Here since both are mandatory attributes, I'd say we don't really care, so maybe just call the functions?

api/protobuf/core/api_container_service.proto
```diff
@@ -434,6 +434,7 @@ message UploadFilesArtifactArgs {
 message UploadFilesArtifactResponse {
   // UUID of the files artifact, for use when referencing it in the future
   string uuid = 1;
+  string name = 2;
```
gbouv, line 439: nit: Maybe add a quick comment similar to what we have above. It's our API after all

cli/cli/commands/files/upload/upload.go
```diff
@@ -93,19 +89,16 @@ func run(
 	if err != nil {
 		return stacktrace.Propagate(err, "An error occurred getting the name to be given to the produced artifact")
 	}
-	if artifactName == defaultName {
-		artifactName = fmt.Sprintf(artifactNamePrefix, time.Now().Unix())
-	}
 
-	filesArtifactUuid, err := enclaveCtx.UploadFiles(path, artifactName)
+	filesArtifactUuid, fileArtifactName, err := enclaveCtx.UploadFiles(path, artifactName)
```
gbouv, line 93: nit: I'd name the argument and the response the same to emphasize the fact that they represent the same thing. `fileArtifactName` seems better to me

api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -215,20 +215,20 @@ func (enclaveCtx *EnclaveContext) GetServices() (map[services.ServiceName]servic
...
+func (enclaveCtx *EnclaveContext) UploadFiles(pathToUpload string, artifactName string) (services.FilesArtifactUUID, services.FileArtifactName, error) {
 	content, err := shared_utils.CompressPath(pathToUpload, ensureCompressedFileIsLesserThanGRPCLimit)
 	if err != nil {
-		return "", stacktrace.Propagate(err,
+		return "", "", stacktrace.Propagate(err,
 			"There was an error compressing the file '%v' before upload",
 			pathToUpload)
 	}
 
 	args := binding_constructors.NewUploadFilesArtifactArgs(content, artifactName)
```
gbouv, line 226: Hmm, so the `artifactName` attribute on `args` is now optional. Here you're simply using the fact that empty string = no attribute. But since protobuf gives us a nice way to differentiate between the two, should we change the artifactName here to be optional? I'd do it, to be explicit at the API level. Wdyt?

core/server/api_container/server/api_container_service.go
```diff
@@ -369,14 +369,18 @@ func (apicService ApiContainerService) GetExistingAndHistoricalServiceIdentifier
 	return &kurtosis_core_rpc_api_bindings.GetExistingAndHistoricalServiceIdentifiersResponse{AllIdentifiers: allIdentifiers}, nil
 }
 
-func (apicService ApiContainerService) UploadFilesArtifact(ctx context.Context, args *kurtosis_core_rpc_api_bindings.UploadFilesArtifactArgs) (*kurtosis_core_rpc_api_bindings.UploadFilesArtifactResponse, error) {
+func (apicService ApiContainerService) UploadFilesArtifact(_ context.Context, args *kurtosis_core_rpc_api_bindings.UploadFilesArtifactArgs) (*kurtosis_core_rpc_api_bindings.UploadFilesArtifactResponse, error) {
+	maybeArtifactName := args.GetName()
```
gbouv, line 373: And here I'd call `args.Name` and below have something like `if maybeArtifactName == nil || maybeArtifactName == ""`.
Doesn't change the world, but this way by just looking at the API someone will immediately know that the artifact name is optional. Whereas in the current state we need to look into the code to discover that it's indeed optional

gbouv (approved):
Looks good. I'd just update the API to be extra explicit that artifactName as a param to UploadFile is now optional.

core/server/api_container/server/api_container_service.go
Peeeekay: i don;t see a big value in adding optional to be honest - i will stick what we currently have as I don;t need to extra `nil` check.

api/golang/core/lib/enclaves/enclave_context.go
Peeeekay: it would require another if statement, if artifact == "" - pass in nil -- i think its not needed and unnecessary imo

cli/cli/commands/files/upload/upload.go
Peeeekay: mhm: `fileArtifactName` is of type `services.FileArtifactName` and `artifactName` is of type `string` .. I will leave it like this for now.

PR #93, opened 2023-03-03
feat!: This is a breaking change where we are removing the `ExecRecipe.service_name`, `GetHttpRequestRecipe.service_name`, and `PostHttpRequestRecipe.service_name` fields, we suggest users pass this value as an argument in the `exec`, `request` and `wait` instructions where this type is currently used. We are also deprecating the previous `exec`, `request`, and `wait` instructions signature that haven't the `service_name` field, users must add this field on these instructions call. Another change is that now the `service_name` field on the `exec`, `request`, and `wait` instructions is mandatory.

leoporoli:
Fixes https://github.com/kurtosis-tech/kurtosis-private/issues/1128


leoporoli:
@gbouv I'm adding the docs changes also here because basically, I'm removing content instead of adding new redactions, so I think is not necessary someone from product to check this changes, but if we think that this decision is not good I can split these on two

core/server/api_container/server/startosis_engine/startosis_interpreter_test.go
```diff
@@ -64,14 +64,13 @@ func TestStartosisInterpreter_DefineFactAndWait(t *testing.T) {
...
 	get_recipe = GetHttpRequestRecipe(
-		service_name = "web-server",
 		port_id = "http-port",
 		endpoint = "?input=output",
 		extract = {
 			"input": ".query.input"
 		}
 	)
-	response = plan.wait(get_recipe, "code", "==",  200, timeout="5m", interval="5s")
+	response = plan.wait(get_recipe, "code", "==",  200, timeout="5m", interval="5s",service_name = "web-server")
```
gbouv: nit: blank space before service_name

gbouv (commented):
@leoporoli the code looks good but should we be merging this to the breaking change branch instead of main?

leoporoli:
> @leoporoli the code looks good but should we be merging this to the breaking change branch instead of main?

@gbouv you are totally right! I forgot this, it seems that the regular habit of sending every PR to the main pushed me to do it. I'll change the destination, good catch! thanks

leoporoli:
@gbouv I changed the destination branch. Could you review it, please? thanks

gbouv (approved):

PR #97, opened 2023-03-06
refactor: Migrate ConnectionConfig to Starlark Framework

gbouv:


core/server/api_container/server/startosis_engine/kurtosis_types/connection_config/connection_config.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+type ConnectionConfig struct {
+	*kurtosis_type_constructor.KurtosisValueTypeDefault
+}
+
+func CreateConnectionConfig(packetLossPercentage starlark.Float) (*ConnectionConfig, *startosis_errors.InterpretationError) {
+	args := []starlark.Value{
+		packetLossPercentage,
+		nil, // no delay distribution as we don't need it
```
Peeeekay, line 64: Is this because delay distribution is optional and will be extracted out in toKurtosisType() method?

core/server/api_container/server/startosis_engine/kurtosis_types/connection_config/connection_config.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	packetLossPct := float32(packetLossPctStarlark)
+
+	packetDelayDistributionStarlark, found, interpretationErr := kurtosis_type_constructor.ExtractAttrValue[packet_delay_distribution.PacketDelayDistribution](
+		connectionConfig.KurtosisValueTypeDefault, PacketDelayDistributionAttr)
+	if interpretationErr != nil {
+		return nil, interpretationErr
+	}
+	var packetDelayDistribution partition_topology.PacketDelayDistribution
+	if found {
```
Peeeekay, line 92: nit 
``` 
packetDelayDistribution := partition_topology.NewUniformPacketDelayDistribution(0)
if found {
  packetDelayDistributionPtr, interpretationErr := packetDelayDistributionStarlark.ToKurtosisType()
  if interpretationErr != nil {
    return nil, interpretationErr
  }
  packetDelayDistribution = *packetDelayDistributionPtr
}
```

Peeeekay (approved):

core/server/api_container/server/startosis_engine/kurtosis_types/connection_config/connection_config.go
gbouv: It's just because we don't need in the go codebase to create a ConnectionConfig with a non default delay distribution. So, leaving it as it is for now. We can add it as a param in the future if we need to

core/server/api_container/server/startosis_engine/kurtosis_types/connection_config/connection_config.go
gbouv: Same remark as in the other PR, I tend to prefer final variables for readability reasons, even though sometimes, like here, it adds an if clause to initialise it. If you don't feel too strongly about it I'll leave it this way here, happy to chat more about it if you have concerns and I can eventually submit a FLUP PR to change it

PR #99, opened 2023-03-06
docs: Add LICENSE

victorcolombo:


.github/workflows/change-versions.yml
```diff
@@ -21,6 +21,7 @@ jobs:
         run: | 
           api/scripts/update-package-versions.sh "$(cat version.txt)"
           api/scripts/update-own-version-constants.sh "$(cat version.txt)"
+          scripts/update-license-version.sh "$(cat version.txt)"
```
h4ck3rk3y, line 24: i was wondering if .github/ is a better place for this; but this is fine too

scripts/update-license-version.sh
```diff
@@ -0,0 +1,56 @@
...
+
+if ! $(kudet update-version-in-file "${to_update_abs_filepath}" "${REPLACE_PATTERN_YEAR}" "${current_year}"); then
+    echo "Error: An error occurred setting year '${current_year}' in constants file '${constant_file_abs_filepath}' using pattern '${REPLACE_PATTERN_YEAR}'" >&2
+    exit 1
+fi
+
+if ! $(kudet update-version-in-file "${to_update_abs_filepath}" "${REPLACE_PATTERN_CHANGE_DATE}" "${license_change_date}"); then
+    echo "Error: An error occurred setting change date '${license_change_date}' in constants file '${constant_file_abs_filepath}' using pattern '${REPLACE_PATTERN_CHANGE_DATE}'" >&2
+    exit 1
+fi
```
h4ck3rk3y: nit: new line

scripts/update-license-version.sh
```diff
@@ -0,0 +1,56 @@
...
+
+current_year=$(date +"%Y")
+
+license_change_date=$(date -v +4y +"%Y-%m-%d")
+
+# ==================================================================================================
+#                                             Main Logic
+# ==================================================================================================
+to_update_abs_filepath="${root_dirpath}/${PACKAGE_JSON_FILEPATH}"
+if ! $(kudet update-version-in-file "${to_update_abs_filepath}" "${REPLACE_PATTERN_VERSION}" "${new_version}"); then
```
h4ck3rk3y, line 43: why use kudet and not sed?

h4ck3rk3y (approved):

scripts/update-license-version.sh
victorcolombo: This will be part of the greater effort on a subsequent PR

.github/workflows/change-versions.yml
victorcolombo: I might take a look at how we do/store scripts when deprecating kudet

PR #101, opened 2023-03-06
feat: enclave clean has both name and uuid

### LINKED ISSUE #38: [FR]: Print enclave name when running `kurtosis clean -a` (in addition to the UUID)
## Observed Behavior
When I run `kurtosis clean -a` with a running or stopped enclave, Kurtosis will remove all artifacts and prints out the enclave UUID.

image

## Expected Behavior
I'd like to be able to see the UUID and the names of the artifacts that were cleaned. 

## Why
Printing both the UUID and the name of the artifacts that were cleaned helps with readability and makes the print outputs more consistent with other commands, such as `kurtosis enclave ls`. `kurtosis service add` and `kurtosis enclave inspect` which prints out both UUID and names for enclaves and containers.

Example output of `kurtosis enclave inspect`: 
image

Example output of `kurtosis enclave ls`:
image

Example output of `kurtosis service add`:
image

h4ck3rk3y:


h4ck3rk3y:
image

victorcolombo (approved):

PR #106, opened 2023-03-07
feat: docs are versioned

h4ck3rk3y:


h4ck3rk3y:
image

after the first release is done; this is what it would look like

Peeeekay (approved):

PR #110, opened 2023-03-08
docs: Simplify the docs landing page with latest "distribution application development platform" language

mieubrisse:


victorcolombo:
@mieubrisse Are we propagating this change across the website as well? I would like to have both the README and Docs to have similar messaging

victorcolombo (approved):

PR #122, opened 2023-03-09
fix: Pass correct latest tag to GoReleaser CLI build

victorcolombo:


cli/cli/scripts/build.sh
```diff
@@ -92,7 +92,7 @@ fi
     else
         goreleaser_verb_and_flags="build --rm-dist --snapshot --single-target"
     fi
-    if ! goreleaser ${goreleaser_verb_and_flags}; then
+    if ! GORELEASER_CURRENT_TAG=$(cat $root_dirpath/version.txt) goreleaser ${goreleaser_verb_and_flags}; then
```
h4ck3rk3y, line 95: LGTM assuming that bash expansion works1

h4ck3rk3y (approved):

cli/cli/scripts/build.sh
victorcolombo: this is run both locally and on CI

PR #124, opened 2023-03-09
build: Stop publishing go module to a separate repository

victorcolombo:


.circleci/config.yml
```diff
@@ -809,42 +806,6 @@ jobs:
   # -- End build jobs ---------------------------------------
 
   # -- Publish jobs --------------------------------------------
-  publish_kurtosis_api_source:
```
h4ck3rk3y, line 812: On thumbs but you need to remove this from the workflows section below as well

h4ck3rk3y: oh it got removed

h4ck3rk3y (approved):

PR #129, opened 2023-03-09
refactor: Remove kudet as a dependency for building Kurtosis

victorcolombo:


scripts/get-docker-tag.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
tedim52, line 1: Is this script grabbing the most recent tag on the branch? Right now, it doesn't seem like it, but could be wrong. For context: here is the spec I implemented for `kudet get-docker-tag` along with the [PR](https://github.com/kurtosis-tech/kudet/pull/10).

```
repo = get git repo

dockerTag = get the tag on the most recent commit if it exists

if tag doesn't exist: 
	dockerTag = get the most recent commit hash

if working tree is dirty
	 dockerTag = append dirty to the tag
	else
	 dockerTag = append nothing

dockerTag = sanitize dockerTag

return docker tag
```

victorcolombo: Not really, but getting the latest tag is not needed for any build step, only for publish step. In that case, I switched to using `version.txt` file: https://github.com/kurtosis-tech/kurtosis/blob/main/.circleci/config.yml#L899

victorcolombo: It seems to be across the board that all scripts take either a semver version (used to be generated by kudet, now we depend on `version.txt`), or they call kudet (now, the new script) for git tag

tedim52: From my understanding when initially devving `kudet get-docker-tag` the desired outcome was that even during build/development, the created docker images have tags corresponding with the version (coming from the latest git tag) of Kurtosis (eg. `0.68.6`, `0.68.6-dirty`) - and if those aren't available commit hash tag will do (eg.`asfd2-dirty`). If that's not the desired outcome anymore, then definitely agree that it's not needed. Going to approve jus flagging, may be worth clarifying if it's still a desired outcome

tedim52 (approved):

scripts/get-docker-tag.sh
tedim52: For more context, I think that desired outcome came from wanting to use created docker images to help with local Kurtosis development, and knowing the version(as opposed to a commit hash or smth more random) is useful for testing changes, but it could be the case that Kurtosis dev's don't use docker images to dev anymore.

PR #131, opened 2023-03-09
docs: create guide on installing historical versions

tedim52:


docs/docs/guides/installing-historical-versions.md
```diff
@@ -0,0 +1,67 @@
...
+
+<details>
+<summary>Homebrew</summary>
+
+1. Uninstall your current version of Kurtosis
+    ```
+    brew uninstall kurtosis-tech/tap/kurtosis-cli
+    ```
+
+2. Install an earlier version of Kurtosis
```
h4ck3rk3y, line 20: Do we have to uninstall beofre we install?

its alright if we do; just confirming

h4ck3rk3y (approved):

docs/docs/guides/installing-historical-versions.md
tedim52: Technically, you don't but it's much simpler to just uninstall (otherwise you have to do some brew linking commands).

Issue #135, opened 2023-03-10
[FR]: Sort services in `enclave inspect` by service creation time


mieubrisse:
### Background and motivation

Right now the current `enclave inspect` output sorts services on their UUID, which leads to crazy sorts:

<img width="460" alt="Screen Shot 2023-03-09 at 15 56 12" src="https://user-images.githubusercontent.com/1142185/224197062-2551c572-8578-4400-a6ad-81076e7dbeb7.png">

It would be reasonable to sort based on the service creation time

### Current behavior

Sort by UUID

### Desired behavior

Sort by service creation time

### How important is this feature or improvement to you?

Nice to have, this feature would make using Kurtosis more enjoyable.

leeederek:
Sorting by name (#139) is likely good enough here and I'm in favor of closing this out. @mieubrisse, do you agree that the changes made in #139 address this pain point, even though the implementation is different (sorting by name instead of creation time).

mieubrisse:
Fine for now, though I do think sorting by creation time will become more important when we have bigger enclaves

PR #139, opened 2023-03-10
fix: sort services by name

h4ck3rk3y:
we currently sort by uuid, so my-nginx-0, my-nginx-2 and my-nginx-1 might get sorted in the wrong order. This sorts by name which is similar to how things were sorted before we went from GUID (service-id-timestamp) to UUID (hex string)

Note, services aren't still sorted by creation time as they never were. That would be a result of #135


h4ck3rk3y:
image

gbouv (approved):

PR #151, opened 2023-03-13
feat: Added `ReadyConditions` type which is used in `ServiceConfig` for defining how to check services readiness 

leoporoli:


core/server/api_container/server/startosis_engine/kurtosis_types/service_config/ready_conditions.go
```diff
@@ -0,0 +1,213 @@
...
+
+// ReadyConditions is a starlark.Value that holds all the information needed for ensuring service readiness
+type ReadyConditions struct {
+	*kurtosis_type_constructor.KurtosisValueTypeDefault
+	recipe    recipe.Recipe
+	field     string
+	assertion string
+	target    starlark.Comparable
+	interval  time.Duration
+	timeout   time.Duration
```
gbouv: Out of curiosity why do you need to store those fields? They are already stored inside the KurtosisValueTypeDefault and you can extract them on demand, maybe?

core/server/api_container/server/startosis_engine/kurtosis_types/service_config/ready_conditions.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func (readyConditions *ReadyConditions) ValidateAndFillReadyConditions() *startosis_errors.InterpretationError {
```
gbouv: Why to you have this logicin a separate function? Can't that happen at interpretation? Or even better when you actually need to extract those values?

core/server/api_container/server/startosis_engine/kurtosis_types/service_config/ready_conditions.go
```diff
@@ -0,0 +1,213 @@
...
+		found             bool
+		httpRecipe        *recipe.HttpRequestRecipe
+		execRecipe        *recipe.ExecRecipe
+		interpretationErr *startosis_errors.InterpretationError
+	)
+
+	httpRecipe, found, interpretationErr = kurtosis_type_constructor.ExtractAttrValue[*recipe.HttpRequestRecipe](readyConditions.KurtosisValueTypeDefault, RecipeAttr)
+	genericRecipe = httpRecipe
+	if interpretationErr != nil || !found {
+		execRecipe, found, interpretationErr = kurtosis_type_constructor.ExtractAttrValue[*recipe.ExecRecipe](readyConditions.KurtosisValueTypeDefault, RecipeAttr)
```
gbouv: FWIW we should rework the recipe types to inherit a single common type, this will avoid the double parsing here. We do that for the distribution objects. Do you mind adding a TODO here?

core/server/api_container/server/startosis_engine/kurtosis_types/service_config/ready_conditions.go
```diff
@@ -0,0 +1,213 @@
...
+		genericRecipe     recipe.Recipe
+		found             bool
+		httpRecipe        *recipe.HttpRequestRecipe
+		execRecipe        *recipe.ExecRecipe
+		interpretationErr *startosis_errors.InterpretationError
+	)
+
+	httpRecipe, found, interpretationErr = kurtosis_type_constructor.ExtractAttrValue[*recipe.HttpRequestRecipe](readyConditions.KurtosisValueTypeDefault, RecipeAttr)
+	genericRecipe = httpRecipe
+	if interpretationErr != nil || !found {
```
gbouv: `interpretationErr  != nil` is really different than `found` here. If `found == true`, it's pointless to try to parse it as a ExecRecipe as the argument simply doesn't exist at all. I think this would improve readability:
```
if !found {
    return startosis_errors.NewInterpretationError("Required attribute '%s' could not be found on type '%s'",
				RecipeAttr, ReadyConditionsTypeName)
}
if interpretationErr != nil {
    ... try parse it as an ExecRecipe
}
```

core/server/api_container/server/startosis_engine/kurtosis_types/service_config/ready_conditions.go
```diff
@@ -0,0 +1,213 @@
...
+					Name:              IntervalAttr,
+					IsOptional:        true,
+					ZeroValueProvider: builtin_argument.ZeroValueProvider[starlark.String],
+					Validator:         nil,
+				},
+				{
+					Name:              TimeoutAttr,
+					IsOptional:        true,
+					ZeroValueProvider: builtin_argument.ZeroValueProvider[starlark.String],
+					Validator:         nil,
```
gbouv: You probably want to add validators here to fail as early as possible when the string passed is not a valid duration. Wdyt?

core/server/api_container/server/startosis_engine/kurtosis_types/service_config/service_config.go
```diff
@@ -268,6 +275,22 @@ func (config *ServiceConfig) ToKurtosisType() (*kurtosis_core_rpc_api_bindings.S
 	return builder.Build(), nil
 }
 
+func (config *ServiceConfig) GetReadyConditions() (*ReadyConditions, *startosis_errors.InterpretationError) {
```
gbouv, line 278: Where is this called?

By doing that, you're breaking the framework idea of "fail as early as possible". By having this `ValidateAndFillReadyConditions` in the ReadyConditions object, you're basically postponing parsing of the argument to a later code execution. For example if we pass in an invalid duration string, it currently doesn't fail at interpretation, which is sad because it could very well do.
Therefore in here you have to create this `GetReadyConditions` method, which IDK where it's being used ATM but I basically you're propagating the same behaviour: interpretation of ServiceConfig will not fail even if ReadyConditions is invalid.

I think the expected way to implement this is:
- Do not store any of the timeout, intrval, etc inside ReadyConditions
- Add validator to each field, such that interpretation will validate the fields
- Do the cast inside each Getter you have in ReadyConditions.

If you do all this, I don't think it will be more code, and potential user input failures will happen earlier.
Happy to chat if you have concerns, maybe I'm missing something

gbouv (commented):
Maybe add some tests for ReadyConditions?

core/server/api_container/server/startosis_engine/kurtosis_types/service_config/ready_conditions.go
leoporoli: I wanted to do it this way because it's necessary to validate all of them at the same time during the interpretation and then I save them in the object because of this method:
image

That will be called then here:
image

image


And the object it's needed here:
image

.github/workflows/pr-description-update.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           delimiter="Changelog picked up from commits here:"
           cleaned_body="${pr_body%${delimiter}*}"
           cleaned_body_without_trailing_space="$(python .github/remove_trailing_space.py "${cleaned_body}")"
-          conventional_commits="$(git log --pretty=format:"%B"  origin/main..HEAD | grep -Pio '^(build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test|¯\\_\(ツ\)_\/¯)(\(\w+\)?((?=:\s)|(?=!:\s)))?(!)?(:\s.*)?(\n\n.*)?$' | grep ':' || true)"
+          conventional_commits="$(git log --pretty=format:"%B"  origin/${{github.base_ref}}..HEAD | grep -Pio '^(build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test|¯\\_\(ツ\)_\/¯)(\(\w+\)?((?=:\s)|(?=!:\s)))?(!)?(:\s.*)?(\n\n.*)?$' | grep ':' || true)"
```
leoporoli, line 36: I'm adding this because otherwise was picking all the conventional commits from the main branch and in this case we have the `breaking-release` as a base branch

leoporoli:
@gbouv this guy is ready for a second review, thanks

.github/workflows/pr-description-update.yml
gbouv: Nice!

core/server/api_container/server/startosis_engine/kurtosis_types/service_config/ready_conditions.go
```diff
@@ -0,0 +1,231 @@
...
+	*kurtosis_type_constructor.KurtosisValueTypeDefault
+}
+
+func (readyConditions *ReadyConditions) GetRecipe() (recipe.Recipe, *startosis_errors.InterpretationError) {
+	var (
+		genericRecipe     recipe.Recipe
+		found             bool
+		httpRecipe        *recipe.HttpRequestRecipe
+		execRecipe        *recipe.ExecRecipe
+		interpretationErr *startosis_errors.InterpretationError
```
gbouv: Do you need to declare all those upfront?

core/server/api_container/server/startosis_engine/kurtosis_types/service_config/ready_conditions.go
```diff
@@ -0,0 +1,231 @@
...
+		//TODO we should rework the recipe types to inherit a single common type, this will avoid the double parsing here.
+		_, ok := value.(*recipe.ExecRecipe)
+		if !ok {
+			return startosis_errors.NewInterpretationError("The '%s' attribute is not a Recipe (was '%s').", RecipeAttr, reflect.TypeOf(value))
+		}
+	}
+	return nil
+}
+
+func validateDuration(value starlark.Value, attributeName string) *startosis_errors.InterpretationError {
```
gbouv: We have a file where we implement generic validator, maybe you could add this one there, seems fairly generic. At least we will need that when we migrate recipe object types

core/server/api_container/server/startosis_engine/recipe/http_request_recipe.go
```diff
@@ -27,11 +27,11 @@ const (
 
 	statusCodeKey    = "code"
 	bodyKey          = "body"
-	extractKeyPrefix = "extract"
+	ExtractKeyPrefix = "extract"
```
gbouv, line 30: Yup, we need to migrate that one to the framework. I believe the recipe types are the only ones left

gbouv (approved):
Looks really good!

leeederek:
@leoporoli are we tracking to add docs about this too?

PR #196, opened 2023-03-14
feat: Add library to manage context configurations

gbouv:
## Description:
see PR title

NO


context-config-store/api/golang/kurtosis_context_visitor_test.go
```diff
@@ -0,0 +1,66 @@
...
+	contextName = "context-name"
+
+	visitorTestResult = "pong"
+)
+
+func TestVisitorCoversAllFields(t *testing.T) {
+	tmp := new(isKurtosisContext_KurtosisContextInfo)
+	elem := reflect.ValueOf(tmp).Elem()
+	typ := reflect.TypeOf(*tmp)
+	fmt.Println(typ)
```
h4ck3rk3y: should this be an assertion/ require instead?

context-config-store/api/protobuf/context_config.proto
```diff
@@ -0,0 +1,55 @@
...
+}
+
+message ContextUuid {
+  string value = 1;
+}
+
+message LocalOnlyContextV0 {
+}
+
+message RemoteContextV0 {
```
h4ck3rk3y: can we add some comments on fields of this type

context-config-store/scripts/build.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+# 2021-07-08 WATERMARK, DO NOT REMOVE - This script was generated from the Kurtosis Bash script template
+
+set -euo pipefail   # Bash "strict mode"
+script_dirpath="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+root_dirpath="$(dirname "${script_dirpath}")"
+
+go_mod_file_name="go.mod"
+protobuf_generate_script="${root_dirpath}/../api/scripts/protobuf-bindings-generator.sh"
```
h4ck3rk3y: nit: run dirname on root_dirpath and call it "monorepo_root_dirpath" or something

context-config-store/store/context_config_store.go
```diff
@@ -0,0 +1,40 @@
...
+
+type ContextConfigStore interface {
+	// GetAllContexts returns the list of currently saved contexts.
+	GetAllContexts() ([]*api.KurtosisContext, error)
+
+	// GetCurrentContext returns the current context information.
+	GetCurrentContext() (*api.KurtosisContext, error)
+
+	// SwitchContext switches to the context passed as an argument.
+	// It throws an error is the contextUuid does not.
```
h4ck3rk3y: typo

context-config-store/store/context_config_store.go
```diff
@@ -0,0 +1,40 @@
...
+	SwitchContext(contextUuid *api.ContextUuid) error
+
+	// AddNewContext adds a new context to the store.
+	// If the context already exists, it does nothing. Note that context equality is computed on
+	// context UUID, nothing else. To update a context, consider removing it and adding it again.
+	AddNewContext(contextToAdd *api.KurtosisContext) error
+
+	// RemoveContext removes the contexts passed as an argument.
+	// It does nothing is the contextUuid does not point to any known context.
```
h4ck3rk3y: s/is the/ if the/g

context-config-store/scripts/build.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+# 2021-07-08 WATERMARK, DO NOT REMOVE - This script was generated from the Kurtosis Bash script template
```
h4ck3rk3y: this doesn't work for me at all I get

running root scripts/build.sh
```
ok  	github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/database_accessors/enclave_db/partition_topology_db/service_partitions	(cached)
?   	github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/database_accessors/test_helpers	[no test files]
?   	github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/network_helpers	[no test files]
ok  	github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/operation_parallelizer	(cached)
ok  	github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/uuid_generator	(cached)
?   	github.com/kurtosis-tech/kurtosis/context-config-store/api/golang	[no test files]
# github.com/kurtosis-tech/kurtosis/context-config-store/store/persistence
store/persistence/default_context_config.go:6:32: undefined: golang.NewContextUuid
store/persistence/default_context_config.go:8:27: undefined: golang.NewLocalOnlyContext
store/persistence/default_context_config.go:10:32: undefined: golang.NewKurtosisContextConfig
FAIL	github.com/kurtosis-tech/kurtosis/context-config-store/store [build failed]
FAIL	github.com/kurtosis-tech/kurtosis/context-config-store/store/persistence [build failed]
FAIL
Error: Build script '/Users/gyanendramishra/work/kurtosis/context-config-store/scripts/build.sh' failed
```

Can we add this to CI as well so that we know whether the scripts/tests run or break? 

```
# gyanendramishra @ Gyanendras-Laptop in ~/work/kurtosis on git:gbouv/context-config-store o [10:57:11] C:130
$ ./context-config-store/scripts/build.sh
?   	github.com/kurtosis-tech/kurtosis/context-config-store/api/golang	[no test files]
# github.com/kurtosis-tech/kurtosis/context-config-store/store/persistence
store/persistence/default_context_config.go:6:32: undefined: golang.NewContextUuid
store/persistence/default_context_config.go:8:27: undefined: golang.NewLocalOnlyContext
store/persistence/default_context_config.go:10:32: undefined: golang.NewKurtosisContextConfig
FAIL	github.com/kurtosis-tech/kurtosis/context-config-store/store [build failed]
FAIL	github.com/kurtosis-tech/kurtosis/context-config-store/store/persistence [build failed]
FAIL
```

context-config-store/scripts/build.sh
```diff
@@ -0,0 +1,24 @@
...
+proto_files_dir="${root_dirpath}/api/protobuf/"
+golang_api_dir="${root_dirpath}/api/golang"
+golang_language_name="golang"
+
+# ==================================================================================================
+#                                             Main Logic
+# ==================================================================================================
+cd "${root_dirpath}"
+
+# generate protobuf golang files using protobuf-bindings-generator script
```
h4ck3rk3y: this breaks pattern in the sense that we normally split building & protobuf generation.

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+
+	currentContextUuid := contextConfig.GetCurrentContext()
+	var knownContexts []string
+	for _, kurtosisContext := range contextConfig.GetContexts() {
+		knownContexts = append(knownContexts, kurtosisContext.GetUuid().GetValue())
+		if kurtosisContext.GetUuid().GetValue() == currentContextUuid.GetValue() {
+			return kurtosisContext, nil
+		}
+	}
+	return nil, stacktrace.NewError("Unable to find current context info in context config file. "+
```
h4ck3rk3y: this error assumes that the store is the file_backed_config store, while you allow for a generic type.

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+		"Current context is set to '%s' but known contexts are: '%s'",
+		currentContextUuid.GetValue(),
+		strings.Join(knownContexts, ", "))
```
h4ck3rk3y: super nit: put this delimiter in a constant

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+	defer store.Unlock()
+
+	contextConfig, err := store.storage.LoadContextConfig()
+	if err != nil {
+		return stacktrace.Propagate(err, "Unable to load current context config")
+	}
+
+	var doesExist bool
+	for _, kurtosisContext := range contextConfig.GetContexts() {
+		doesExist = doesExist || kurtosisContext.GetUuid().GetValue() == contextUuid.Value
```
h4ck3rk3y: this seems like a clever way of doing this but is 

```
if kurtosisContext.GetUuid().Value() == contextUuid.value { 
     doesExist = true
     break
}
```

more readable? maybe that's just a pattern I am used too

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	if !doesExist {
+		return stacktrace.NewError("Context with UUID '%s' does not exist yet.", contextUuid.GetValue())
```
h4ck3rk3y: nit: period at tend of error which we normally don't do

h4ck3rk3y: s/exist yet./exist in store/g

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	newContextConfig := api.NewKurtosisContextConfig(contextUuid, contextConfig.GetContexts()...)
+	if err = store.storage.PersistContextConfig(newContextConfig); err != nil {
+		return stacktrace.Propagate(err, "Unable to persist new current context to store")
```
h4ck3rk3y: are we persisting new context to store or a new config?

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+		RWMutex: &sync.RWMutex{},
+		storage: storage,
+	}
+}
+
+func (store *contextConfigStoreImpl) GetAllContexts() ([]*api.KurtosisContext, error) {
+	store.RLock()
+	defer store.RUnlock()
+
+	contextConfig, err := store.storage.LoadContextConfig()
```
h4ck3rk3y: these names at first are a bit tough to read / reason with

now I understand that a ContextConfig has a lot of contexts in it and is not the config of a given context. Can we name better?

context-config-store/store/context_config_store_impl.go
h4ck3rk3y: i think this comes from the naming confusion, as far as I understand there is no new context here just the config that keeps the contexts has changed

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return stacktrace.Propagate(err, "Unable to load current context config")
+	}
+
+	newContextUuid := contextToAdd.GetUuid()
+	for _, kurtosisContext := range contextConfig.GetContexts() {
+		if kurtosisContext.GetUuid().GetValue() == newContextUuid.GetValue() {
+			return nil // doing nothing as context already exists
```
h4ck3rk3y: interesting that we choose to return nil instead of erroring (fail loudly) like a dictionary. can this have side effects? will the person assume that the context they are storing got stored while the previous one stays persisted?

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+
+	var newContexts []*api.KurtosisContext
+	for _, kurtosisContext := range contextConfig.GetContexts() {
+		newContexts = append(newContexts, kurtosisContext)
+	}
+	newContexts = append(newContexts, contextToAdd)
+
+	newContextConfig := api.NewKurtosisContextConfig(contextConfig.GetCurrentContext(), newContexts...)
+	if err = store.storage.PersistContextConfig(newContextConfig); err != nil {
+		return stacktrace.Propagate(err, "Unable to persist new context config to store")
```
h4ck3rk3y: in this case both a new context config (as a new config has been added) and a new context are being stored.

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+		if kurtosisContext.GetUuid().GetValue() == newContextUuid.GetValue() {
+			return nil // doing nothing as context already exists
+		}
+	}
+
+	var newContexts []*api.KurtosisContext
+	for _, kurtosisContext := range contextConfig.GetContexts() {
+		newContexts = append(newContexts, kurtosisContext)
+	}
+	newContexts = append(newContexts, contextToAdd)
```
h4ck3rk3y: are all of these the new context or is only the contextToAdd the one?

context-config-store/api/protobuf/context_config.proto
```diff
@@ -0,0 +1,55 @@
...
+
+// Regenerate with the following command ran from repo root
+//    protoc -I=context-config-store/api/protobuf/ --go_out=context-config-store/api/golang/ --go_opt=module=github.com/kurtosis-tech/kurtosis/context-config-store/api/golang context-config-store/api/protobuf/context_config.proto
+
+// NOTE: It sucks that we have this Go-specific logic inside this file (which should be language-agnostic). However, the Protobuf team have
+// taken a hard stance on this being the way it should be done, so we have to do it this way.
+option go_package = "github.com/kurtosis-tech/kurtosis/context-config-store/api/golang";
+
+message KurtosisContextConfig {
+  ContextUuid currentContext = 1;
```
h4ck3rk3y: can we rename this to currentContextUuid

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+
+	newContextConfig := api.NewKurtosisContextConfig(contextConfig.GetCurrentContext(), newContexts...)
```
h4ck3rk3y: here the first item is the uuid an the rest are actual contexts 👍

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+	return nil
+}
+
+func (store *contextConfigStoreImpl) RemoveContext(contextUuid *api.ContextUuid) error {
+	store.Lock()
+	defer store.Unlock()
+
+	contextConfig, err := store.storage.LoadContextConfig()
+	if err != nil {
+		return stacktrace.Propagate(err, "Unable to load current context config")
```
h4ck3rk3y: what does "current" mean here?

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	foundContextToRemove := false
+	var newContexts []*api.KurtosisContext
+	for _, kurtosisContext := range contextConfig.GetContexts() {
```
h4ck3rk3y: nit: s/kurtosisContext/kurtosisContextInStore/

context-config-store/store/context_config_store_impl_test.go
```diff
@@ -0,0 +1,157 @@
...
+	// Setup storage mock
+	storage := persistence.NewMockConfigPersistence(t)
+	contextConfig := api.NewKurtosisContextConfig(otherContextUuid, localContext) // unknown current context UUID
+	storage.EXPECT().LoadContextConfig().Return(contextConfig, nil)
+
+	// Run test
+	testContextConfigStore := NewContextConfigStore(storage)
+	result, err := testContextConfigStore.GetCurrentContext()
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "Unable to find current context info in context config file. Current context is set to 'other-context-uuid' but known contexts are: 'context-uuid'")
```
h4ck3rk3y: nice!

h4ck3rk3y: super nit: use the consts above

context-config-store/store/context_config_store_impl_test.go
```diff
@@ -0,0 +1,157 @@
...
+	storage := persistence.NewMockConfigPersistence(t)
+	contextConfig := api.NewKurtosisContextConfig(contextUuid, localContext, otherLocalContext)
+	storage.EXPECT().LoadContextConfig().Return(contextConfig, nil)
+
+	// Run test
+	testContextConfigStore := NewContextConfigStore(storage)
+	err := testContextConfigStore.AddNewContext(otherLocalContext)
+	require.NoError(t, err)
+
+	// Need to check the method exist first because if the method name changes in the future this test would do nothing
```
h4ck3rk3y: oh wow so thorough

context-config-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,139 @@
...
+		} else {
+			foundContextToRemove = true
+		}
+	}
+
+	if !foundContextToRemove {
+		return nil
+	}
+
+	newContextConfig := api.NewKurtosisContextConfig(contextConfig.GetCurrentContext(), newContexts...)
```
h4ck3rk3y: how does this behave if we remove the current context? should we throw or do we let it get into an inconsistent state?

context-config-store/store/persistence/config_persistence.go
```diff
@@ -0,0 +1,16 @@
+package persistence
+
+import (
+	api "github.com/kurtosis-tech/kurtosis/context-config-store/api/golang"
+)
+
+type ConfigPersistence interface {
+	// init initializes the persistence. It does nothing it's already initialized
```
h4ck3rk3y: sensing a typo here It does nothing it's already initialized

context-config-store/api/golang/context_config.pb.go
```diff
@@ -0,0 +1,610 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.28.1
+// 	protoc        v3.21.12
+// source: context_config.proto
+
+package golang
```
h4ck3rk3y: we are using PROTO here because we want to send it over the wire?

context-config-store/store/persistence/file_backed_config_persistence.go
```diff
@@ -0,0 +1,128 @@
...
+			return stacktrace.Propagate(err, "Unexpected error checking if context config file exists at '%s'",
+				configFilePath)
+		}
+	} else {
+		// file already exists, no need to create it
+		persistence.backingFilePath = configFilePath
+		return nil
+	}
+
+	persistence.Lock()
```
h4ck3rk3y: can this just be at the top like other methods?

context-config-store/store/persistence/file_backed_config_persistence.go
```diff
@@ -0,0 +1,128 @@
...
+	defer persistence.Unlock()
+
+	persistence.backingFilePath = configFilePath
+	if err = persistence.persistContextConfigInternal(defaultContextConfig); err != nil {
+		return stacktrace.Propagate(err, "Unable initialize context config file writing default config to it")
```
h4ck3rk3y: there are some typos and missing words here

context-config-store/store/persistence/file_backed_config_persistence.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return nil
+}
+
+func (persistence *FileBackedConfigPersistence) persistContextConfigInternal(contextConfig *api.KurtosisContextConfig) error {
+	serializedConfig, err := serializer.Marshal(contextConfig)
+	if err != nil {
+		return stacktrace.Propagate(err, "Unable to serialize content of context config object")
```
h4ck3rk3y: .. to json

context-config-store/store/persistence/file_backed_config_persistence.go
```diff
@@ -0,0 +1,128 @@
...
+	"google.golang.org/protobuf/encoding/protojson"
+	"os"
+	"path"
+	"sync"
+)
+
+const (
+	applicationDirname = "kurtosis" // TODO: this is common to Kurtosis config, should refactor
+
+	fileName        = "context-config.json"
```
h4ck3rk3y: can we just write Proto to disk? or do we want the user to be able to read the file

context-config-store/store/persistence/file_backed_config_persistence_test.go
```diff
@@ -0,0 +1,71 @@
...
+      },
+      "name": "context-name",
+      "localOnlyContextV0": {}
+    }
+  ]
+}`
+)
+
+func TestPersistConfig(t *testing.T) {
+	tempFile, err := os.CreateTemp(tempFileDir, tempFileNamePattern)
```
h4ck3rk3y: should we defer delete this file?

context-config-store/store/persistence/file_backed_config_persistence_test.go
```diff
@@ -0,0 +1,71 @@
...
+	err = storage.PersistContextConfig(contextConfig)
+	require.Nil(t, err)
+
+	fileContent, err := os.ReadFile(tempFile.Name())
+	require.Nil(t, err)
+	require.Equal(t, serializedContextConfig, string(fileContent))
+}
+
+func TestLoadConfig(t *testing.T) {
+	tempFile, err := os.CreateTemp(tempFileDir, tempFileNamePattern)
```
h4ck3rk3y: should we defer delete this file, "picking up your own toys"

context-config-store/store/persistence/default_context_config.go
```diff
@@ -0,0 +1,10 @@
+package persistence
+
+import api "github.com/kurtosis-tech/kurtosis/context-config-store/api/golang"
+
+var personalContextName = "personal"
+var personnalContextUuid = api.NewContextUuid("00000000-0000-0000-0000-000000000000")
```
gbouv: I suggest we change that for `default` and also planning to using our own uuid generator, so UUID will be `00000000000000000000000000000000` instead

gbouv: Updated the notion doc already, will vet that to Kevin just in case

context-config-store/store/persistence/file_backed_config_persistence_test.go
```diff
@@ -0,0 +1,71 @@
...
+	require.Nil(t, err)
+	_, err = tempFile.Write([]byte(serializedContextConfig))
+	require.Nil(t, err)
+
+	storage := newFileBackedConfigPersistenceForTesting(tempFile.Name())
+	result, err := storage.LoadContextConfig()
+	require.Nil(t, err)
+	require.Equal(t, contextConfig.GetCurrentContext().GetValue(), result.GetCurrentContext().GetValue())
+	require.Len(t, result.GetContexts(), len(contextConfig.GetContexts()))
+	require.Equal(t, contextConfig.GetContexts()[0].GetUuid().GetValue(), result.GetContexts()[0].GetUuid().GetValue())
```
h4ck3rk3y: can we  check contextConfig.GetContexts() == result.GetContexts()?

h4ck3rk3y: ```require.True(t, proto.Equal(contextConfig.GetContexts()[0], result.GetContexts()[0]))```

context-config-store/scripts/build.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
h4ck3rk3y: can we stick a readme here

context-config-store/api/golang/kurtosis_context_visitor_test.go
gbouv: Huh I tried to implement a test to automatically check that all sub types are part of the visitor, but it was becoming too complex and I gave up. I initially didn't want to commit that `TestVisitorCoversAllFields` test, will remove. Sorry for the confusion

context-config-store/api/protobuf/context_config.proto
gbouv: Yup, added a bunch of comments in there

context-config-store/store/context_config_store_impl.go
gbouv: 👍 , I'll rephrase

context-config-store/store/context_config_store_impl.go
gbouv: No strong opinion, both works with me. Will change!

context-config-store/store/context_config_store_impl.go
gbouv: Yes you're right, I think it's probably safer to throw loudly here. If users want to add a context that already exists, we can suggest them to remove it first and then adding it again. I'll throw here with a clear message, good suggestion!

context-config-store/store/context_config_store_impl.go
gbouv: contextToAdd is the only new one. I changed the variable name to be `updatedContextsList`, hopefully it helps

context-config-store/store/context_config_store_impl.go
gbouv: Updated it to "Unable to load the list of contexts currently stored", hope it helps

context-config-store/store/context_config_store_impl.go
gbouv: Changed here and everywhere

context-config-store/store/context_config_store_impl.go
gbouv: Arf, good point, I did check for that in my prototype but forgot to add it here :( I'll add, and also add a test for this. Good flag!

context-config-store/store/persistence/config_persistence.go
gbouv: I think I might be alittle dyslexic when it comes to if/it/is

context-config-store/store/persistence/file_backed_config_persistence.go
gbouv: I initially wrote the proto to disk straight, but I think writing JSON doesn't add a ton of complexity and is easier to look at (even though we don't expect anyone to look at this file, better safe than sorry)

context-config-store/store/persistence/file_backed_config_persistence_test.go
gbouv: Yup good point, will add

context-config-store/store/persistence/file_backed_config_persistence_test.go
gbouv: Oh nice, didn't know about his proto.Equal! Good suggestion!

context-config-store/scripts/build.sh
gbouv: Yeah, I didn't account for the fact that our own proto generation script would wipe the directory before updating the bindings... This should be fixed now, and I added a step in the CI to build it there as well

context-config-store/scripts/build.sh
gbouv: Extracted proto generation outside of the build.sh

context-config-store/store/context_config_store_impl.go
gbouv: Yeah good point, I agree `config` is a bit overloaded here... I changed a bunch of names now, hopefully that makes things clearer?? Let me know what you think

context-config-store/api/golang/context_config.pb.go
gbouv: Hmm, I don't think we will send this over the wire, at least not now. We could have used raw go types, but then it's a _real_ pain to deal with union types, proto has that built-in. We could have used another type generation language as well, but I don't see a compelling argument for bringing another tooling in our repo where proto meets the requirements IMO. Also, this is not user facing at all. Let me know what you think

context-config-store/scripts/build.sh
gbouv: You mean in the build.sh script or in the sub folder? To explain what this store is doing?

context-config-store/store/persistence/file_backed_config_persistence.go
gbouv: Technically the above logic does not need to happen behind a lock. Even more behind a Write-Lock as it's just checking the existence of the file and nothing else. What I did is that I moved the `persistence.Lock()` call inside the persistsContextsStateInternal function, it's nicer this way and with that we don't need to use the lock in this function at all :)

context-config-store/api/golang/context_config.pb.go
h4ck3rk3y: I don't see the the value as keeping something as proto and then storing it in JSON; maybe the marshal/unmarshal story is less complex with a normal Go struct but you don't get unions with it I guess.

Can we stick a comment on top saying that this isn't sent over the wire and the only reason its Proto is that we want to suppport unions?

context-config-store/store/persistence/file_backed_config_persistence.go
h4ck3rk3y: if our workflows don't involve people looking at the file on disk perhaps just keep it proto and the code becomes simpler?

YAGNI. No strong opinions though. Using proto for unions and then storing it as JSON seems complex. Maybe it can be simpler.

context-config-store/scripts/build.sh
h4ck3rk3y: not for the script but in the package / subfolder explaining the store

scripts/build.sh
```diff
@@ -13,6 +13,7 @@ root_dirpath="$(dirname "${script_dirpath}")"
 BUILD_SCRIPT_RELATIVE_FILEPATHS=(
     "scripts/generate-kurtosis-version.sh"
     "container-engine-lib/scripts/build.sh"
+    "context-config-store/scripts/build.sh"
```
h4ck3rk3y: this refers to the previous folder!

contexts-state-store/api/golang/kurtosis_context_visitor.go
```diff
@@ -0,0 +1,22 @@
...
+
+type KurtosisContextVisitor[ResultType any] struct {
+	VisitLocalOnlyContextV0 func(localContext *generated.LocalOnlyContextV0) (*ResultType, error)
+
+	VisitRemoteContextV0 func(localContext *generated.RemoteContextV0) (*ResultType, error)
+}
+
+func Visit[ResultType any](kurtosisContext *generated.KurtosisContext, visitor KurtosisContextVisitor[ResultType]) (*ResultType, error) {
+	if kurtosisContext.GetLocalOnlyContextV0() != nil {
+		return visitor.VisitLocalOnlyContextV0(kurtosisContext.GetLocalOnlyContextV0())
```
h4ck3rk3y: why do we have the V0 suffix?

contexts-state-store/store/context_config_store_impl.go
```diff
@@ -0,0 +1,157 @@
...
+
+	currentContextUuid := contextsState.GetCurrentContextUuid()
+	var contextUuidsInStore []string
+	for _, kurtosisContextInStore := range contextsState.GetContexts() {
+		contextUuidsInStore = append(contextUuidsInStore, kurtosisContextInStore.GetUuid().GetValue())
+		if kurtosisContextInStore.GetUuid().GetValue() == currentContextUuid.GetValue() {
+			return kurtosisContextInStore, nil
+		}
+	}
+	return nil, stacktrace.NewError("Unable to find current context info in currently stored contexts config. "+
```
h4ck3rk3y: config? state?

contexts-state-store/store/context_config_store.go
```diff
@@ -0,0 +1,39 @@
...
+	// GetCurrentContext returns the current context information.
+	GetCurrentContext() (*generated.KurtosisContext, error)
+
+	// SwitchContext switches to the context passed as an argument.
+	// It throws an error if the contextUuid does not point to any known context.
+	SwitchContext(contextUuid *generated.ContextUuid) error
+
+	// AddNewContext adds a new context to the store.
+	// It throws an error if a context with the same UUID already exists
+	AddNewContext(newContext *generated.KurtosisContext) error
```
h4ck3rk3y: should the store be providing the id or should the user?

contexts-state-store/store/context_config_store_impl_test.go
```diff
@@ -0,0 +1,187 @@
...
+	expectContextsStateAfterRemoval := api.NewKurtosisContextsState(contextUuid, localContext)
+	storage.EXPECT().PersistContextsState(expectContextsStateAfterRemoval).Return(nil)
+
+	// Run test
+	testContextConfigStore := NewContextConfigStore(storage)
+	err := testContextConfigStore.RemoveContext(otherContextUuid)
+	require.NoError(t, err)
+}
+
+func TestRemoveContext_FailureCurrentContext(t *testing.T) {
```
h4ck3rk3y: Neat!

contexts-state-store/store/persistence/config_persistence.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	api "github.com/kurtosis-tech/kurtosis/contexts-state-store/api/golang/generated"
+)
+
+type ConfigPersistence interface {
+	// init initializes the persistence. It does nothing if it is already initialized
+	init() error
+
+	// PersistContextsState writes the context config to the persistence
```
h4ck3rk3y: state / config?

contexts-state-store/store/persistence/file_backed_config_persistence.go
```diff
@@ -0,0 +1,124 @@
...
+func (persistence *FileBackedConfigPersistence) PersistContextsState(newContextsState *generated.KurtosisContextsState) error {
+	if err := persistence.init(); err != nil {
+		return stacktrace.Propagate(err, "Unable to initialize context config persistence")
+	}
+	return persistence.persistContextsStateInternal(newContextsState)
+}
+
+func (persistence *FileBackedConfigPersistence) LoadContextsState() (*generated.KurtosisContextsState, error) {
+	if err := persistence.init(); err != nil {
+		return nil, stacktrace.Propagate(err, "Unable to initialize context config persistence")
```
h4ck3rk3y: config/ state ?

contexts-state-store/store/persistence/file_backed_config_persistence_test.go
```diff
@@ -0,0 +1,73 @@
...
+        "value":  "context-uuid"
+      },
+      "name":  "context-name",
+      "localOnlyContextV0":  {}
+    }
+  ]
+}`
+)
+
+func TestPersistConfig(t *testing.T) {
```
h4ck3rk3y: config/state?

contexts-state-store/api/golang/context_config_constructors.go
```diff
@@ -0,0 +1,26 @@
+package golang
+
+import "github.com/kurtosis-tech/kurtosis/contexts-state-store/api/golang/generated"
```
h4ck3rk3y: rename file if we go with state

contexts-state-store/api/protobuf/context_config.proto
```diff
@@ -0,0 +1,62 @@
+syntax = "proto3";
+package context_config_store;
```
h4ck3rk3y: perhaps rename this if we go with state

contexts-state-store/store/context_config_store.go
```diff
@@ -0,0 +1,39 @@
+package store
```
h4ck3rk3y: perhaps rename these files to have state instead of store.

h4ck3rk3y (approved):
@gbouv have some more minor comments. Name wise I think I prefer "Config" more than I prefer "State"

Earlier I think the function was called "LoadContextConfig", which feels like config of a single context if we use plurals "LoadContextsConfig" maybe its easier to grasp. I don't want to block you on this and I'll leave the decision to you here.

gbouv:
Sounds good Gyani I can roll back to config, I was trying to be creative but maybe Config with plural "contexts" will work just fine

contexts-state-store/store/context_config_store.go
gbouv: Probably neither. In the real world, to connect to a remote context, the connection file will provide it. Once we start supporting multiple local context, the CLI can generate one on the fly if it helps. In any case, I'd like to restrict the store to its only purpose: storing. At least for now, if that works for you

contexts-state-store/api/golang/kurtosis_context_visitor.go
gbouv: To have a V1 in the near future?? :P I imagine those objects are going to evolve, and we can do so very easily by adding a new member to the union type. If we want to make a breaking change to `LocalOnlyContextV0`, we can introduce `LocalOnlyContextV1` by adding it to the oneof proto file

context-config-store/api/golang/context_config.pb.go
gbouv: How do you handle union types with go JSON serialization?

context-config-store/store/persistence/file_backed_config_persistence.go
gbouv: I don't follow you on this... The only thing that changes is `proto.Marshal` becomes `protojson.Marshal` and same for `Unmarshal` (I removed the multiline thing). I mean, I'm happy to serialize the proto straight to disk, it's just that doing it in JSON doesn't add _any_ complexity to the code _at all_, and if at some point, for whatever reason, who knows, we want to look into this file, we keep this door open. Basically here cost = 0, value = 0 but mayyyybe value could be 0.5 who knows. So, really unless I'm missing something I don't understand your point here sorry

contexts-state-store/store/context_config_store.go
h4ck3rk3y: Yeah sounds good to me

contexts-state-store/api/golang/kurtosis_context_visitor.go
h4ck3rk3y: feels a bit like future proofing but we already have it so keep it :) Resolving this.

context-config-store/store/persistence/file_backed_config_persistence.go
h4ck3rk3y: Agree; the added complexity is very very little and compensates more than enough for the value it provides. 

Two different formats made me uneasy but it was a knee jerk.

context-config-store/scripts/build.sh
gbouv: Added a REAME to the contexts-config-store module :)

contexts-config-store/store/persistence/default_contexts_config.go
```diff
@@ -0,0 +1,28 @@
...
+
+func NewDefaultContextsConfig() (*generated.KurtosisContextsConfig, error) {
+	// This logic is similar to uuid_generate.GenerateUUIDString, but we don't want to pull container engine lib here
+	// just for this.
+	// TODO: extract the UUID generator outside of container engine lib maybe
+	randomUuid, err := uuid.NewRandom()
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "Unable to generate a random UUID for the default context")
+	}
+	newContextUuidStr := strings.Replace(randomUuid.String(), "-", "", -1)
```
h4ck3rk3y, line 23: perhaps take these magic constants out

h4ck3rk3y: if we are in the business of generating uuids should we generate uuids for other additions too?

weird that default would look like "aedf1230...." and someone can just put in "hello"

PR #197, opened 2023-03-14
fix: wait instruction hanging forever when `service_name` field is not passed

### LINKED ISSUE #174: [bug]: `wait` hangs forever, rather than throwing an error, when you don't specify the service to wait upon

0.68.8

Painful, this is causing significant friction in my workflow.

Write a `wait` without specifying the service:
```
    plan.wait(
        ExecRecipe(command = psql_command_prefix + ["-c", "\\l"]),
        "code",
        "==",
        0,
    )
```
Run, and observe it hang forever without throwing an error.

### Relevant log output

_No response_

### Desired behavior

An error should be thrown, because we're not actually specifying where we're running the wait.

leoporoli:
## Description:
Were are moving the validation to the interpretation layer to detect the missing field before executing the instructions.

fixes #174

Now if the  argument is not specified this error will be thrown:
image


Peeeekay (approved):

leeederek:
hey @leoporoli , should the service_name field be mandatory? The docs still say optional.

leoporoli:
@leeederek the mandatory change and the doc's update are in the breaking-change branch now, it will release this week, details [here](https://github.com/kurtosis-tech/kurtosis/blob/2023-03-22-breaking-release/core/server/api_container/server/startosis_engine/kurtosis_instruction/wait/wait.go#L81) and [here](https://github.com/kurtosis-tech/kurtosis/blob/2023-03-22-breaking-release/docs/docs/reference/starlark-instructions.md#wait)

PR #204, opened 2023-03-15
fix: Add back fetch depth to change version GH action

victorcolombo:
## Description:

During the process of removing , an accidental line was removed from the change version Github Action: https://github.com/kurtosis-tech/kurtosis/commit/d5d58c3c666bd4da9d886c9f777e9fc6b14fb57d#diff-39713e0e880264021f8f50fa74fc6f9a0fd40e39a82f233d0f46ea12adb9799dL15

* [ ] Yes
* [X] No

fix: Add back fetch depth to change version GH action


tedim52 (approved):
LGTM.

PR #206, opened 2023-03-15
fix: Tag docker images correctly after Kudet removal

victorcolombo:

* [ ] Yes
* [X] No


mieubrisse (approved):

mieubrisse:
I temporarily disabled branch protection in order to merge this, because the last release (0.68.9) was actually broken so old_enclave_continuity_test was failing and I don't think there was a way to make it pass. Branch protection now back on.

PR #216, opened 2023-03-15
fix: collapse current behavior into background+motivation

leeederek:
## Description:
We received feedback that some users felt like they were repeating themselves when filling out details for the  the "Current Behavior" section and "Background and Motivation" sections. This PR collapses those two together to make submitting FRs easier and smoother.

* [X] Yes
* [ ] No


yarn.lock
leoporoli, line 1: what are you changing in this PR? I'm seeing only the yarn.lock file

leoporoli (approved):

leoporoli (approved):

PR #223, opened 2023-03-15
feat: integrate lsp with monrepo

Peeeekay:

* [ ] Yes
* [x] No


Peeeekay:
This is the PR for me to the sanity check - and verify whether integration will work kurtosis; and it does. kurtosis lsp starts the lsp server. I am not going to merge this PR but will create separate smaller PRs for easy review. However, would be awesome if you can just look at the comments and let me know if the plan sounds good! @h4ck3rk3y @victorcolombo CC: @adschwartz

.circleci/config.yml
```diff
@@ -361,23 +361,24 @@ jobs:
 
   build_cli:
     docker:
-      - image: "cimg/go:<< pipeline.parameters.cli-go-version >>"
+      - image: "docker/tilt-releaser"
```
Peeeekay, line 383: This image contains necessary tools for cross compilation; I do not personally want to use tilt-releaser as it is.. I am thinking for v0 -- I can just use `kurtosis-releaser` with base image being `tilt-releaser`.

One thing I noticed is that this image is 2 GB whereas previous image was just 145mb and we CircleCI doesn't cache the layers -- as a FLUP we should do it. ( didn't look into how to do it - let me know if you guys did before?)

cli/cli/.goreleaser.yml
```diff
@@ -5,24 +5,34 @@ before:
...
     goos:
-      - linux
       - darwin
+      - linux
     goarch:
       - amd64
       - arm64
+    env:
+      - CGO_ENABLED=1
+      - CC={{ if index .Env "CC_WITH_DEFAULT" }}{{ .Env.CC_WITH_DEFAULT }}{{ else }} clang {{ end }}
```
Peeeekay, line 20: Using an environment variable to update the compilers for arm64. This is because I haven't figure out a way to download o64h-clang locally - need to investigate image details further to fully understand how to download - it requires (osxcross). I feel like because this is open source repo -- i am reluctant to introduce new tooling for local development. I can maybe take a look it as a FLUP ( but imo it's not high priority)

starlark-lsp/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
Peeeekay: I will update the READ.me to be more professional in a separate PR; will mention about it being a fork from starlark-lsp and remove the references from it as well. 

I will change the folder name and everything to `klsp` -- unless you different suggestion?

Peeeekay: There's another alternative where I can have `lsp` in a different repo and use it as package in `cli/cli/lsp.go`; If you think that's better way to organize the code - i can do that too.

Peeeekay:
There are still few things missing such as:
1. tests for extensions
2. logging ( currently LSP server does not respect the flags we input via cli)  - so will make it a little hard to debug initially but i have ways around it.

I will add these as FLUPS -- ( or side quests)

.circleci/config.yml
Peeeekay: if we don't want to do this ; the alternative is to have starlark-lsp as it's separate binary and exists in a separate repository. Users would need to install and run it separately, or we do that automatically via kurtosis cli. Ideally we could potentially install a separate binary from the plugin ( but it might add delay to this project)

gbouv: Do you have concrete data on the downsides here? Is build being slower? If yes how many times slower? Is it going to be much more expensive? If yes by how much. IDK if you were waiting for input on this, but if it's the case I'd suggest we get more concrete data points as it's difficult to take a decision on something vague like this.

cli/cli/.goreleaser.yml
gbouv: Yeah I agree, for those things makes sure you stick comments everywhere explaining what they do and why they are needed, they look pretty cryptic but if that's what we need to get it to work, IMO no need to invest extra time just to make this cleaner

starlark-lsp/LICENSE
gbouv: What's the cost of having it in a separate repo? (this includes future costs as well, in case there are some). I'd personally bias toward extracting it in its own repo if we can. It's quite far from the core business of what the CLI is currently doing, and I'm usually in favour of splitting responsibility clearly when it does not incur additional complexities.
(you can track moving it to another repo in a FLUP, if merging it here gives you velocity, then I'm good with temporarily having it here)

.circleci/config.yml
Peeeekay: yepp; I do not have the cost data -- let me know if you know where to look for it.

Regarding the completion time - with `tilt-releaser` it took 58s to complete and with the older docker image it took 2s to complete. Overall, `build_cli` step took `30s` difference. Though baking `goreleaser into the image` may bring the execution time. I strongly think that docker api errors may be happening because of lack of caching on our end, as circle ci was pulling many layers in the image. @gbouv

starlark-lsp/LICENSE
Peeeekay: if `starlark-lsp` in another repo is that we will be using circleci with `tilt-releaser` in two repos ( and will executed twice) so there maybe additional cost. However, I am not able to determine the cost delta yet, should be slight increase though. Other than that, i do not think there is another major issue.

.circleci/config.yml
```diff
@@ -380,23 +380,24 @@ jobs:
...
       # Generate Kurtosis Version
       - run: "<<pipeline.parameters.generate-kurtosis-version-script-path>>"
 
       # Install goreleaser
       - run: |
-          echo 'deb [trusted=yes] https://repo.goreleaser.com/apt/ /' | sudo tee /etc/apt/sources.list.d/goreleaser.list
-          sudo apt update
-          sudo apt install goreleaser
+          echo 'deb [trusted=yes] https://repo.goreleaser.com/apt/ /' | tee /etc/apt/sources.list.d/goreleaser.list
+          curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
```
victorcolombo, line 399: Add a comment to explain why this curl is necessary

cli/cli/.goreleaser.yml
victorcolombo, line 20: Add comment to explain what it `CC_WITH_DEFAULT`. Is this a Go release environment variable? Or does it come with the container?

lsp/kurtosis_extension.go
```diff
@@ -0,0 +1,179 @@
+package lsp
```
victorcolombo, line 1: What is the rationale of having this as a separate go module?

cli/cli/commands/lsp/lsp.go
```diff
@@ -0,0 +1,15 @@
+package lsp
+
+import (
+	"github.com/kurtosis-tech/kurtosis/lsp"
+	starlark_lsp_cli "github.com/kurtosis-tech/vscode-kurtosis/starlark-lsp/pkg/cli"
+	"github.com/spf13/cobra"
+)
+
+func NewLspCommand() *cobra.Command {
```
victorcolombo, line 9: LSP naming is a very technical name, what reveals implementation details we don't care much for. I wonder if we can change this to something more seamless like `kurtosis dev` or `kurtosis code` (we could even open `code` on `.` for users)

Peeeekay: We can change the name later; this is internal command. If you notice, I have `Hidden: true` - this basically hides command. It does not show on `--help` for example. I don't feel strongly about the name, and therefore can be potentially changed later.

lsp/kurtosis_extension.go
Peeeekay: The main rationale here is that I am following `Single Responsibility Principle` - everything related to kurtosis apis is part of this package. This separation can easily allow us to improve the plugin suggestions without changing the code much, and importantly once we have an API ready to mostly need to change this package and everything else should largely remain the same. ( Open/Closed principle) @victorcolombo

h4ck3rk3y:
Excusing myself from reviews as @victorcolombo is on it

cli/cli/commands/lsp/lsp.go
victorcolombo: Cool, if this is hidden I'm fine with this naming

lsp/kurtosis_extension.go
victorcolombo: Make sense, but I don't know if I like this being on root folder. My rationale is that every folder in root is either a structural piece of kurtosis ("service"/"application") or a packages that is shared across services/app. I believe this could even live inside `cli/lsp` since it's only consumed by CLI.

Peeeekay: Is it cool if I make that change in the next PR? @victorcolombo .. This will just allow for folks to do quick testing without relying on any manual steps?

victorcolombo (approved):
LGTM

PR #224, opened 2023-03-15
fix: print the upgrade CLI warning at most hourly

### LINKED ISSUE #169: [FR]: Print the out-of-date Kurtosis version WARN message once per hour
### Background and motivation

This error message is really, really annoying. It actually does force me to upgrade though, so it's very good for what we want it - just would like this to be less annoying (and the Waku folks agree).

### Current behavior

The pester happens on every `kurtosis` CLI run

### Desired behavior

The pester happens maximum once per hour.

(Impl note: we can easily accomplish this by dropping a file with the last pester time in the "data" directory of XDG. Use the codepath for the CLI actually determining what the latest version is as an example. @leoporoli built this and it works very well).

h4ck3rk3y:
## Description:
We would check for versions on everyrun and warn users if they are on an old version. We only do that hourly now.

* [x] Yes
* [ ] No

https://github.com/kurtosis-tech/kurtosis/issues/169


Peeeekay (approved):

leeederek:
Hey @h4ck3rk3y , when you say:
> We would check for versions on everyrun

Do you mean that Kurtosis would check for the versions and print out a WARN message every time a command was passed to the APIC? And now it checks on a time-based cadence? 

Thanks!

PR #241, opened 2023-03-17
feat: Add `context` root command and simple `ls` subcommand

gbouv:

NO (at least we don't want to communicate it yet as it's doing nothing)


cli/cli/commands/kurtosis_context/kurtosis_context.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"github.com/kurtosis-tech/kurtosis/cli/cli/command_str_consts"
+	"github.com/kurtosis-tech/kurtosis/cli/cli/commands/kurtosis_context/ls"
+	"github.com/spf13/cobra"
+)
+
+// ContextCmd Suppressing exhaustruct requirement because this struct has ~40 properties
+// nolint: exhaustruct
+var ContextCmd = &cobra.Command{
```
leoporoli, line 11: why are you not using `lowlevel.LowlevelKurtosisCommand` ?

leoporoli: forget it. now I realized. I should read all the file before commenting.

cli/cli/commands/kurtosis_context/ls/ls.go
```diff
@@ -0,0 +1,98 @@
...
+	contextsConfigStore := store.GetContextConfigStore()
+	contextsConfig, err := contextsConfigStore.GetKurtosisContextsConfig()
+	if err != nil {
+		return stacktrace.Propagate(err, "Error retrieving currently configured contexts")
+	}
+	currentContextUuid := contextsConfig.GetCurrentContextUuid()
+
+	tablePrinter := output_printers.NewTablePrinter(contextCurrentColumnHeader, contextUuidColumnHeader, contextNameColumnHeader, contextRemoteColumnHeader)
+	for _, kurtosisContext := range contextsConfig.GetContexts() {
+		var isCurrent string
```
leoporoli: nit: I suggest to rename this to `isCurrentStr` because I was expecting a bool when a read the name

cli/cli/commands/kurtosis_context/ls/ls.go
```diff
@@ -0,0 +1,98 @@
...
+		if kurtosisContext.GetUuid().GetValue() == currentContextUuid.GetValue() {
+			isCurrent = isCurrentContextStrIndicator
+		}
+
+		var contextUuidToDisplay string
+		if showFullUuids {
```
leoporoli, line 69: nit: you could simplify this with:
```
contextUuidToDisplay := kurtosisContext.GetUuid().GetValue()
if !showFullUuids {
   contextUuidToDisplay = uuid_generator.ShortenedUUIDString(contextUuidToDisplay)
}
```

cli/cli/commands/kurtosis_context/ls/ls.go
```diff
@@ -0,0 +1,98 @@
...
+			contextUuidToDisplay = kurtosisContext.GetUuid().GetValue()
+		} else {
+			contextUuidToDisplay = uuid_generator.ShortenedUUIDString(kurtosisContext.GetUuid().GetValue())
+		}
+
+		var remoteStrToDisplay string
+		_, err = contexts_config_api.Visit[struct{}](kurtosisContext, contexts_config_api.KurtosisContextVisitor[struct{}]{
```
leoporoli: I suggest putting the second argument in a var here to improve legibility.

cli/cli/commands/kurtosis_context/ls/ls.go
```diff
@@ -0,0 +1,98 @@
...
+	contexts_config_generated_api "github.com/kurtosis-tech/kurtosis/contexts-config-store/api/golang/generated"
+	"github.com/kurtosis-tech/kurtosis/contexts-config-store/store"
+	"github.com/kurtosis-tech/stacktrace"
+)
+
+const (
+	contextCurrentColumnHeader = ""
+	contextUuidColumnHeader    = "UUID"
+	contextNameColumnHeader    = "Name"
+	contextRemoteColumnHeader  = "Remote"
```
leoporoli: nit: WDYT on renaming to `Remote host`? maybe??

leoporoli (approved):

cli/cli/commands/kurtosis_context/ls/ls.go
gbouv: I prefer using final variables, even at the cost of an extra line of code. Your code snippet makes the `contextUuidToDisplay` non final.
I personally find final variables easier to reason about. They are assigned once, then you don't have to think about them. 
FSA, I think it's a generic good practice to prefer final variables. Goland IDE is flagging them with an underline.
If you don't feel too strongly about it, I'll leave it as it is

cli/cli/commands/kurtosis_context/ls/ls.go
gbouv: Yeah, that makes sense to me. I'll rename! That's something we can easily change anyway :)

cli/cli/commands/kurtosis_context/ls/ls.go
leoporoli: good point, I'll take this practice, thanks, sure let it as it is

PR #242, opened 2023-03-17
docs: point to the right sdk within monorepo

h4ck3rk3y:
We used to point to the old sdk at https://github.com/kurtosis-tech/kurtosis-sdk but now we point to the one within the monorepo


victorcolombo (approved):

leeederek:
@h4ck3rk3y can we now archive https://github.com/kurtosis-tech/kurtosis-sdk then?

h4ck3rk3y:
@leeederek done

PR #246, opened 2023-03-17
build: Skip PR update for Dependabot

victorcolombo:

* [ ] Yes
* [X] No

### DIFF
--- a/.github/workflows/pr-description-update.yml
+++ b/.github/workflows/pr-description-update.yml
@@ -15,7 +15,8 @@ concurrency:
 jobs:
   update_pr:
     # we don't want this to run on breaking change PRs or release please PRs
-    if: contains(github.head_ref, 'release-please') != true && contains(github.head_ref, 'breaking-release') != true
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-description-update.yml
```diff
@@ -15,7 +15,8 @@ concurrency:
 jobs:
   update_pr:
     # we don't want this to run on breaking change PRs or release please PRs
-    if: contains(github.head_ref, 'release-please') != true && contains(github.head_ref, 'breaking-release') != true
+    if: contains(github.head_ref, 'release-please') != true && contains(github.head_ref, 'breaking-release') != true && ${{ github.actor != 'dependabot[bot]' }}
```
h4ck3rk3y, line 18: does the thing expand as expected? the other parts of the if condition has different syntax

h4ck3rk3y (approved):

.github/workflows/pr-description-update.yml
victorcolombo: This was from Dependabot docs, and it worked on this PR so I believe we are good

PR #256, opened 2023-03-20
fix: remove file paths from error message

Peeeekay:

* [ ] Yes
* [x] No

fix: remove file paths from error message


Peeeekay:
Before:

```
Error: An error occurred running command 'run'
 --- at /home/circleci/project/cli/cli/command_framework/lowlevel/lowlevel_kurtosis_command.go:295 (LowlevelKurtosisCommand.MustGetCobraCommand.func2) ---
Caused by: An error occurred calling the run function for command 'run'
 --- at /home/circleci/project/cli/cli/command_framework/highlevel/engine_consuming_kurtosis_command/engine_consuming_kurtosis_command.go:149 (EngineConsumingKurtosisCommand.getRunFunc.func1) ---
Caused by: Error executing Kurtosis code
 --- at /home/circleci/project/cli/cli/commands/run/run.go:249 (run) ---
Caused by: Kurtosis execution threw an error. See output above for more details
 --- at /home/circleci/project/cli/cli/commands/run/run.go:302 (readAndPrintResponseLinesUntilClosed) ---
 ```
 
 After:
 
 ```
 Error: An error occurred running command 'run'
 Caused by: An error occurred calling the run function for command 'run'
 Caused by: Error executing Kurtosis code
 Caused by: Kurtosis execution threw an error. See output above for more details
 ``

File: cli/cli/command_framework/lowlevel/lowlevel_kurtosis_command.go [hunk 1]
```diff
@@ -371,3 +374,24 @@ func renderArgUsageStr(arg *args.ArgConfig) string {
...
 }
+
+// this method removes the file path from the error
+//TODO: this is currently not integrated but will do once all the necessary PRs are merged.
+func removeFilePathFromErrorMessage(errorMessage string) error {
+	errorMessageConvertedInList := strings.Split(errorMessage, separator)
+
+	// only the even numbered elements needs to be picked.
+	var cleanErrorList []string
+	for i, arr := range errorMessageConvertedInList {
```
leoporoli: I suggest you rename `i and arr` for something more semantic

cli/cli/command_framework/lowlevel/lowlevel_kurtosis_command.go
```diff
@@ -371,3 +374,24 @@ func renderArgUsageStr(arg *args.ArgConfig) string {
...
+		if i%2 == 0 {
+			cleanErrorList = append(cleanErrorList, strings.Trim(arr, "\n"))
```
leoporoli: you could put "\n" in a constant

cli/cli/command_framework/lowlevel/lowlevel_kurtosis_command.go
```diff
@@ -371,3 +374,24 @@ func renderArgUsageStr(arg *args.ArgConfig) string {
...
+		}
+	}
+
+	if logrus.GetLevel() == logrus.DebugLevel {
+		logrus.Debug(errors.New(errorMessage))
+	} else {
+		out.FileLogger.Error(errors.New(errorMessage))
+	}
```
leoporoli: this part seems to be `out of the method scope` and brakes the single responsibility principle for functions. I suggest you move it to another method and probably call it before calling this one.

cli/cli/command_framework/lowlevel/lowlevel_kurtosis_command_test.go
```diff
@@ -0,0 +1,31 @@
+package lowlevel
+
+import (
+	"github.com/kurtosis-tech/stacktrace"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestRemoveFilePathFromErrorMessage(t *testing.T) {
```
leoporoli, line 9: Could you add a test containing a file path to check that this is removed?

cli/cli/out/file_logger.go
```diff
@@ -0,0 +1,68 @@
...
+// FileLogger this logger will only log to a file
+// TODO: In commands like inspect we use out.PrintOutLn - will need to add this fileLogger
+//  to print commands' output as well.
+var FileLogger = logrus.New()
+
+func InitFileLogger() {
+	// kurtosis-cli.log can be found in the same directory as kurtosis-config.yml
+	filePath, err := host_machine_directories.GetKurtosisCliLogsFilePath()
+	if err != nil {
+		panic(err)
```
leoporoli: why are you adding panic here instead of adding stacktrace.Propagate ?

cli/cli/out/file_logger.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	//TODO: store at least x number of commands in the file
+	//TODO: print the command, and then the output of the command
+	logFile, err := os.OpenFile(
+		filePath,
+		os.O_WRONLY|os.O_CREATE,
+		0666)
```
leoporoli: you could put this value in a constant

cli/cli/out/file_logger.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if err != nil {
+		panic(err)
```
leoporoli: same comment here about panic

cli/cli/out/file_logger.go
```diff
@@ -0,0 +1,68 @@
...
+		CallerPrettyfier:          nil,
+	}
+
+	FileLogger.SetOutput(logFile)
+	FileLogger.SetLevel(logrus.InfoLevel)
+	FileLogger.SetFormatter(textFormatter)
+
+	// added logrus hook which automatically make sure that all the logging is being done in the file as well
+	// currently only info, warn, debug and error level logs are added to the file but we can add more later.
+	logrus.AddHook(&Hook{
```
leoporoli: nice!

cli/cli/out/hook.go
```diff
@@ -0,0 +1,28 @@
...
+type Hook struct {
+	Writer    io.Writer
+	LogLevels []log.Level
+	Formatter log.Formatter
+}
+
+func (hook *Hook) Fire(entry *log.Entry) error {
+	line, err := hook.Formatter.Format(entry)
+	if err != nil {
+		return err
```
leoporoli: could we add stacktrace here ?

cli/cli/out/hook.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+	_, err = hook.Writer.Write(line)
```
leoporoli: I suggest you handle the error here, add the stacktrace and return the propagated error and the in the last line return nil, it's the standard way that we use to handle the errors

cli/cli/out/hook.go
```diff
@@ -0,0 +1,28 @@
...
+
+	log "github.com/sirupsen/logrus"
+)
+
+// Hook is a hook that writes logs of specified LogLevels to specified Writer
+type Hook struct {
+	Writer    io.Writer
+	LogLevels []log.Level
+	Formatter log.Formatter
+}
```
leoporoli, line 30: you could add a constructor for this object and then make the fields private maybe

cli/cli/command_framework/lowlevel/lowlevel_kurtosis_command.go
Peeeekay: good point; I can do this in the next PR when i am actually integrating the  method

cli/cli/command_framework/lowlevel/lowlevel_kurtosis_command_test.go
Peeeekay: This what it is intended to do. stacktrace.New() and stacktrace.Propagate adds the file path automatically and then i remove them. @leoporoli

cli/cli/out/file_logger.go
Peeeekay: good catch; will fix this.

cli/cli/out/hook.go
Peeeekay: @leoporoli thinking about this --> do you think it would be better to maybe silently fail here? The pros is that users won't be errored out because of logging issue. The downside would be that we would have no way of knowing if it's failing. One way is to use `fmt` to show that message on the cli or have metrics which can help us understand it. For now, I have added stack-traces here.

Peeeekay: or; i can create another logger without hooks - which i can use to print warnings -- and return error as nil.

cli/cli/command_framework/lowlevel/lowlevel_kurtosis_command_test.go
leoporoli: ok

cli/cli/out/hook.go
leoporoli: ok, yes probably will be better to silently fail here, and yes we could include both printing a warning on the CLI and sending a metric (but you can do it with iterations, it's not mandatory to include both now)

cli/cli/out/file_logger.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/kurtosis-tech/stacktrace"
+	"github.com/sirupsen/logrus"
+	"io/fs"
+	"os"
+)
+
+// FileLogger this logger will only log to a file
+// TODO: In commands like inspect we use out.PrintOutLn - will need to add this fileLogger
+//  to print commands' output as well.
+var FileLogger = logrus.New()
```
leoporoli: I suggest you do not use a global var for saving the logger, you could modify the `SetupFileLogger` and convert it to a `GetFileLogger` which implements a singleton pattern and store the values in a private var.
You can use the same pattern that we use for `GetMetricsUserIDStore() *MetricsUserIDStore` you will need to create a new `FileLogger` struct

leoporoli (approved):

cli/cli/out/file_logger.go
Peeeekay: good catch; will do that.

cli/cli/out/hook.go
Peeeekay: ok; let me silently fail here atm - add the extra capabilities in next prs. @leoporoli

PR #264, opened 2023-03-20
feat!: rename enclave and service identifier flags

### LINKED ISSUE #202: [FR]: Simplify `--enclave-identifier` flag in `kurtosis run` to `--enclave`
### Background and motivation

`--enclave-identifier` is very long; `--enclave` is much cleaner now that we have identifiers.

NOTE: There's a typo in the current helptext in that it's missing a space; this should get fixed too:
```
...script or package will be ran.An enclave with this...
```

### Current behavior

The flag is called `--enclave-identifier`

### Desired behavior

The flag should be called `--enclave`

Nice to have, this feature would make using Kurtosis more enjoyable.

h4ck3rk3y:
BREAKING-CHANGE: The --enclave-identifier, --enclave-identifiers  and --service-identifier  flags have been renamed to ,  --enclave, --enclaves and --service  respectively. Users will have to change any scripts or CI configurations that depend on those flags.

## Description:
renamed enclave-identifier -> enclave service-identifier> service and deprecated old "enclave-id"

Closes https://github.com/kurtosis-tech/kurtosis/issues/202


gbouv (approved):
Nice!

You're PR description seems messed up, not sure if it will appear in the changelog, but if it does make sure to fix it before merging. 
Also, I think those flags have been deprecated for some times now, but do we want to merge this to the breaking change branch instead of main?

PR #266, opened 2023-03-20
ci: No PR title checks for merge queue

gbouv:
## Description:
Disable PR title check for merge queue branches

NO


h4ck3rk3y (approved):
I am not sure if this will do what we want it to do

i think the event is a "pull_request_target" and the merge queue isn't a PR

Is it possible to skip "title_check" for merge queues but not for normal PRs?

I am approving this as this won't hurt us and if you have high confidence might as well try it out





so the thing never runs

.github/workflows/conventional-pr-title-checker.yml
```diff
@@ -15,6 +18,9 @@ concurrency:
 
 jobs:
   title_check:
+    # we don't want this job to be skipped on merge queue PRs
```
h4ck3rk3y: `# we want this job to be skipped`

I think don't is a typo

h4ck3rk3y: wait

PR #267, opened 2023-03-20
fix: Update PR title workflow for merge queue

gbouv:

NO

### DIFF
--- a/.github/workflows/conventional-pr-title-checker.yml
+++ b/.github/workflows/conventional-pr-title-checker.yml
@@ -1,15 +1,15 @@
 # Check PR title for conventional commits
 name: Check PR title
 on:
-  push:
-    branches:
-      - gh-readonly-queue/main/**
   pull_request_target:
     types:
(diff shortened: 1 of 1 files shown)


.github/workflows/conventional-pr-title-checker.yml
```diff
@@ -2,7 +2,7 @@
 name: Check PR title
 on:
   push:
-    branches:
+    branches-ignore:
```
h4ck3rk3y: This will ignore the branch completely but run on other branches I think. I don't think this is desired behaviour.

h4ck3rk3y (approved):
I've my doubt that this will work and have put it in comments, unblocking you with the approval if you have conviction

.github/workflows/conventional-pr-title-checker.yml
gbouv: Good catch sorry I mixed 2 updates in one. Revert this line

PR #271, opened 2023-03-20
ci: Fix concurrency groups

gbouv:
## Description:
`head_ref` is only available within a PR. Here we prefer using ref_name for a broader use. It points to the name of the current branch

NO

### DIFF
--- a/.github/workflows/breaking-change-description-title-update.yml
+++ b/.github/workflows/breaking-change-description-title-update.yml
@@ -9,7 +9,7 @@ on:
 
 # cancel redundant builds
 concurrency:
-  group: "update-breaking-change-${{ github.ref_name }}"
+  group: "update-breaking-change-${{ github.ref }}"
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/breaking-change-description-title-update.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 # cancel redundant builds
 concurrency:
-  group: "update-breaking-change-${{ github.head_ref }}"
+  group: "update-breaking-change-${{ github.ref_name }}"
```
h4ck3rk3y: instead of ref_name can this be just `ref`

I am probably being overly cautious here but if there are two prs with the same name but one from a fork and one from this repo, would concurrency work with ref_name or do we need to know origin/fork information

h4ck3rk3y (approved):

.github/workflows/breaking-change-description-title-update.yml
gbouv: Sure, can update

PR #275, opened 2023-03-20
feat: Add context `rm` command

gbouv:
## Description:
Adding `kurtosis context rm` CLI command to remove currently configured contexts.

NO - at least we don't want to communicate on this yet


cli/cli/commands/kurtosis_context/rm/rm.go
```diff
@@ -0,0 +1,53 @@
...
+
+	contextsConfigStore := store.GetContextsConfigStore()
+	contextUuids, err := context_id_arg.GetContextUuidForContextIdentifier(contextIdentifiers)
+	if err != nil {
+		return stacktrace.Propagate(err, "Error finding contexts matching the provided identifiers")
+	}
+
+	logrus.Info("Removing contexts...")
+	for _, contextUuid := range contextUuids {
+		if err := contextsConfigStore.RemoveContext(contextUuid); err != nil {
```
victorcolombo: Maybe we should print the contexts that were successfully deleted? Right now we can get a state were some of them are deleted, we don't know which of them are, but we never get `"Contexts successfully removed"` if one of them fail

cli/cli/command_framework/highlevel/context_id_arg/context_id_arg.go
```diff
@@ -0,0 +1,134 @@
...
+
+		_, err := GetContextUuidForContextIdentifier(contextIdentifiersToValidate)
+		if err != nil {
+			return stacktrace.Propagate(err, "Error finding context matching the provided identifiers")
+		}
+		return nil
+	}
+}
+
+func GetContextUuidForContextIdentifier(contextIdentifiers []string) (map[string]*generated.ContextUuid, error) {
```
victorcolombo: This could possibly benefit from a unit test with a mocked `ContextsConfigStore`

victorcolombo (approved):
LGTM, left 2 comments

cli/cli/commands/kurtosis_context/rm/rm.go
gbouv: Good suggestion, I'll update that slightly. I would like to remain consistent with the behaviour of `kurtosis enclave rm` so what I'll do is that if everything goes well we do print:
```
Removing contexts...
Contexts successfully removed
```

But if one or more error happens we do print the error logs

cli/cli/command_framework/highlevel/context_id_arg/context_id_arg.go
gbouv: That requires to slightly update how we pass in the contexts config store, but the more test the better I agree. I'll try to add

PR #277, opened 2023-03-20
fix: Lock default context in contexts config

gbouv:
## Description:
Disable possibility to add a new context with name `default` or removing the `default` context. This can lead to unexpected situation where someone ends up having a single remote context, and this remote context not working properly on the server. This might be something we want to allow in the future, but for now disabling it will save us some edge cases pain.

NO

### DIFF
--- a/contexts-config-store/store/contexts_config_store_impl.go
+++ b/contexts-config-store/store/contexts_config_store_impl.go
@@ -19,7 +19,7 @@ type contextConfigStoreImpl struct {
 	storage persistence.ConfigPersistence
 }
 
-func NewContextConfigStore(storage persistence.ConfigPersistence) ContextConfigStore {
+func NewContextConfigStore(storage persistence.ConfigPersistence) ContextsConfigStore {
 	return &contextConfigStoreImpl{
 		RWMutex: &sync.RWMutex{},
 		storage: storage,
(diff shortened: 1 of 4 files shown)


contexts-config-store/store/contexts_config_store_impl.go
```diff
@@ -93,6 +93,11 @@ func (store *contextConfigStoreImpl) AddNewContext(newContext *generated.Kurtosi
 	store.Lock()
 	defer store.Unlock()
 
+	if newContext.GetName() == persistence.DefaultContextName {
+		return stacktrace.NewError("Adding a new context names '%s' is not allowed as it is the reserved context "+
```
h4ck3rk3y: s/names/with name/g
 s/it is the/it is a/g

We say that contexts have equality based on uuid but here we are using name as proxy. I think its alright!

contexts-config-store/store/contexts_config_store_impl_test.go
```diff
@@ -159,9 +178,31 @@ func TestRemoveContext_FailureCurrentContext(t *testing.T) {
...
 	err := testContextConfigStore.RemoveContext(contextUuid)
 	require.Error(t, err)
-	expecteErr := fmt.Sprintf("Cannot remove context '%s' as it is currently the selected context. Switch to a "+
+	expectedErr := fmt.Sprintf("Cannot remove context '%s' as it is currently the selected context. Switch to a "+
 		"different context before removing it", contextUuid.GetValue())
-	require.Contains(t, err.Error(), expecteErr)
+	require.Contains(t, err.Error(), expectedErr)
+
+	// Need to check the method exist first because if the method name changes in the future this test would do nothing
+	persistMethod, found := reflect.TypeOf(storage).MethodByName(persistMethodName)
```
h4ck3rk3y, line 186: formatting?

h4ck3rk3y (approved):

PR #282, opened 2023-03-21
fix: be clear about the engine that is being started

### LINKED ISSUE #255: [bug]: An unknown image is used to start a Kurtosis engine (when performing a fresh engine start)

0.68.13

07d14724970e58810d93e5520436603ca4c113ad9bd608bec371f9122534cf4b

Papercut, this bug is frustrating, but I have a workaround.

When I start a fresh Kurtosis engine, the following gets printed:

image

The `from image ''...` looks like a bug/typo. I'm not entirely sure what it means nor does it make me feel like everything performing as intended. Can we remove it or make it more obvious what's happening? 

### Relevant log output

_No response_

### Desired behavior

When I start, upgrade, or re-install Kurtosis, its clear to me what images are being pulled (or abstract that away from the user entirely).

h4ck3rk3y:
## Description:
We would log something vague and we want to be more specific now

YES

Closes https://github.com/kurtosis-tech/kurtosis/issues/255


h4ck3rk3y:
image

cli/cli/commands/engine/start/start.go
```diff
@@ -71,8 +72,10 @@ func run(cmd *cobra.Command, args []string) error {
 	var engineClientCloseFunc func() error
 	var startEngineErr error
 	if engineVersion == defaultEngineVersion {
+		logrus.Infof("Starting Kurtosis engine from image '%v%v%v'...", kurtosisTechEngineImagePrefix, imageVersionDelimiter, kurtosis_version.KurtosisVersion)
 		_, engineClientCloseFunc, startEngineErr = engineManager.StartEngineIdempotentlyWithDefaultVersion(ctx, logLevel)
 	} else {
+		logrus.Infof("Starting Kurtosis engine from image '%v%v%v'...", kurtosisTechEngineImagePrefix, imageVersionDelimiter, engineVersion)
```
victorcolombo, line 78: Is it worth to warn if that is the case? Given that this is an advanced feature

victorcolombo (approved):

cli/cli/commands/engine/start/start.go
h4ck3rk3y: I think the warning would be off putting also user intends to do this so info is alright; at this point the user used the --version flag

PR #283, opened 2023-03-21
ci: make publish_cli depends on others

h4ck3rk3y:
## Description:
publish_cli would work parallel with publish_api_container_image and others, now it depends on that

NO

https://www.notion.so/kurtosistech/Broken-Kurtosis-0-68-9-due-to-publishing-issues-da3aff29f9484c2c86c82dbb3ec8870b


victorcolombo (approved):

h4ck3rk3y:
I will not publish this now. I'll publish it after 0.69.0 is published so that I can roll it back if it breaks things @victorcolombo

PR #285, opened 2023-03-21
feat: Kurtosis backend can now connect to a remote Docker backend

gbouv:
## Description:
This PR adds a specific type of backend consisting of both the local Docker backend and a remote Kurtosis backend (which is effectively Docker). CLI and Engine are using this specific backend to be able to offload some containers to the remote backend based on the current context config

NO -- since we don't have the possibility to switch backend right now, this is effectively not exposed


engine/server/engine/main.go
```diff
@@ -150,16 +153,21 @@ func getEnclaveManager(kurtosisBackend backend_interface.KurtosisBackend, kurtos
...
 	var err error
 	switch kurtosisBackendType {
 	case args.KurtosisBackendType_Docker:
-		kurtosisBackend, err = backend_creator.GetLocalDockerKurtosisBackend(apiContainerModeArgsForKurtosisBackend)
+		kurtosisBackend, err = remote_context_backend.GetContextAwareKurtosisBackend(remoteBackendConfigMaybe, apiContainerModeArgsForKurtosisBackend)
 		if err != nil {
 			return nil, stacktrace.Propagate(err, "An error occurred getting local Docker Kurtosis backend")
 		}
 	case args.KurtosisBackendType_Kubernetes:
+		if remoteBackendConfigMaybe != nil {
```
h4ck3rk3y, line 166: The wording here feels a bit off "Using Kuberenetes as a local cluster connected to a remote"

I was under the impression that with cluster config people could bring in remote Kuberentes Clusters as well. Can we word this like

"Using a Remote Kurtosis Backend isn't allowed with Kuberentes. Either switch to a local only context to use Kuberenets, or switch the cluster to Docker to connect to a remote Kurtosis backend"

engine/server/engine/main.go
```diff
@@ -150,16 +153,21 @@ func getEnclaveManager(kurtosisBackend backend_interface.KurtosisBackend, kurtos
...
+			return nil, stacktrace.NewError("Using Kubernetes as a local cluster connected to a remote " +
```
h4ck3rk3y: We eventually will get rid of cluster all together, then this messaging becomes clear. In that world

1. To run a remote docker Kurtosis enclave, use remote context
2. To run a local docker Kurtosis, use local context
3. To run remote Kubernetes use remote context
4. To run local Kubernetes run local context

Or are we dropping the local K8S support all together? How do people select between Docker/K8S?

contexts-config-store/store/serialization/kurtosis_context.go
```diff
@@ -0,0 +1,23 @@
+package serialization
+
+import (
+	"github.com/kurtosis-tech/kurtosis/contexts-config-store/api/golang/generated"
+	"github.com/kurtosis-tech/stacktrace"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+func SerializeKurtosisContext(kurtosisContext *generated.KurtosisContext) ([]byte, error) {
```
h4ck3rk3y: nit perhaps unit test this

contexts-config-store/store/serialization/contexts_config.go
```diff
@@ -0,0 +1,23 @@
+package serialization
+
+import (
+	"github.com/kurtosis-tech/kurtosis/contexts-config-store/api/golang/generated"
+	"github.com/kurtosis-tech/stacktrace"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+func SerializeKurtosisContextsConfig(kurtosisContextsConfig *generated.KurtosisContextsConfig) ([]byte, error) {
```
h4ck3rk3y: nit: perhaps unit test this

engine/launcher/engine_server_launcher/engine_server_launcher.go
```diff
@@ -68,12 +70,17 @@ func (launcher *EngineServerLauncher) LaunchWithCustomVersion(
...
 	didUserAcceptSendingMetrics bool,
 	backendConfigSupplier KurtosisBackendConfigSupplier,
+	kurtosisRemoteBackendConfigSupplier *KurtosisRemoteBackendConfigSupplier,
 ) (
 	resultPublicIpAddr net.IP,
 	resultPublicGrpcPortSpec *port_spec.PortSpec,
 	resultErr error,
 ) {
 	kurtosisBackendType, kurtosisBackendConfig := backendConfigSupplier.getKurtosisBackendConfig()
+	remoteBackendConfigMaybe, err := kurtosisRemoteBackendConfigSupplier.GetOptionalRemoteConfig()
```
h4ck3rk3y, line 80: just for my understanding - So we don't care about the local context config as its empty, we only check if remote is set otherwise our default path continues

engine/launcher/engine_server_launcher/kurtosis_context_supplier.go
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) 2022 - present Kurtosis Technologies Inc.
+ * All Rights Reserved.
+ */
+
+package engine_server_launcher
+
+import (
+	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_impls/remote_context_backend"
+	"github.com/kurtosis-tech/kurtosis/contexts-config-store/api/golang"
```
h4ck3rk3y: can we name this something different from golang downstream, golang.Visit seems confusing at first

engine/launcher/engine_server_launcher/kurtosis_context_supplier.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+func (supplier KurtosisRemoteBackendConfigSupplier) GetOptionalRemoteConfig() (*remote_context_backend.KurtosisRemoteBackendConfig, error) {
+	kurtosisContext, err := supplier.kurtosisContextSupplier()
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "Error retrieving Kurtosis context object")
+	}
+	remoteBackendConfigMaybe, err := golang.Visit[remote_context_backend.KurtosisRemoteBackendConfig](kurtosisContext, golang.KurtosisContextVisitor[remote_context_backend.KurtosisRemoteBackendConfig]{
+		VisitLocalOnlyContextV0: func(localContext *generated.LocalOnlyContextV0) (*remote_context_backend.KurtosisRemoteBackendConfig, error) {
```
h4ck3rk3y: can we take the instantiation of the visitor out and pass it to the "visit"; it would be less nested

engine/launcher/args/args.go
```diff
@@ -87,6 +92,7 @@ func NewEngineServerArgs(
 		DidUserAcceptSendingMetrics: didUserAcceptSendingMetrics,
 		KurtosisBackendType:         kurtosisBackendType,
 		KurtosisBackendConfig:       kurtosisBackendConfig,
```
h4ck3rk3y: Should we rename this to something more explicit I can't tell how KurtosisBackendConfig plays with KurtosisRemoteBackendConfig

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
```diff
@@ -0,0 +1,208 @@
...
+// RemoteContextKurtosisBackend is a dual context holding a reference to a local backend running on Docker (k8s is
+// currently not supported to work with a remote Kurtosis backend) and a reference to a remote Kurtosis backend (which
+// currently is a Docker backend, we do not support anything else).
+// Basically, requests are dispatched among those 2 backends depending on which area of Kurtosis they touch.
+// This might change, but right now the engine runs locally, therefore engine interacting endpoint goes to the local
+// backend, and all the rest goes to the remote backend, as all other containers (including API container) should be
+// running in the remote backend
+type RemoteContextKurtosisBackend struct {
+	localKurtosisBackend  backend_interface.KurtosisBackend
+	remoteKurtosisBackend backend_interface.KurtosisBackend
```
h4ck3rk3y, line 29: maybe we could be more explicit in the name that this is apic only

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
```diff
@@ -0,0 +1,208 @@
...
+	contextBackend backend_interface.KurtosisBackend,
+) *RemoteContextKurtosisBackend {
+	return &RemoteContextKurtosisBackend{
+		localKurtosisBackend:  localBackend,
+		remoteKurtosisBackend: contextBackend,
+	}
+}
+
+func (backend *RemoteContextKurtosisBackend) FetchImage(ctx context.Context, image string) error {
+	// without knowing which backend will need the image, it's tricky to filter which one should fetch it.
```
h4ck3rk3y: could it be simpler if we check if the image is "kurtosis/engine", then its engine otherwise its APIC.

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
h4ck3rk3y: s/localKurtosisBackend/localKurtosisBackendForEngine/g


s/remoteKurtosisBackend/remoteKurtosisBackendForAPIC/g

random quick suggestions

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
```diff
@@ -0,0 +1,208 @@
...
+
+func (backend *RemoteContextKurtosisBackend) GetLogsCollectorForEnclave(ctx context.Context, enclaveUuid enclave.EnclaveUUID) (*logs_collector.LogsCollector, error) {
+	return backend.remoteKurtosisBackend.GetLogsCollectorForEnclave(ctx, enclaveUuid)
+}
+
+func (backend *RemoteContextKurtosisBackend) DestroyLogsCollectorForEnclave(ctx context.Context, enclaveUuid enclave.EnclaveUUID) error {
+	return backend.remoteKurtosisBackend.DestroyLogsCollectorForEnclave(ctx, enclaveUuid)
+}
+
+func (backend *RemoteContextKurtosisBackend) DestroyDeprecatedCentralizedLogsResources(ctx context.Context) error {
```
h4ck3rk3y, line 209: tangent - perhaps we should delete this completely now, we have released a few versions since this

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
```diff
@@ -0,0 +1,208 @@
...
+)
+
+// RemoteContextKurtosisBackend is a dual context holding a reference to a local backend running on Docker (k8s is
+// currently not supported to work with a remote Kurtosis backend) and a reference to a remote Kurtosis backend (which
+// currently is a Docker backend, we do not support anything else).
+// Basically, requests are dispatched among those 2 backends depending on which area of Kurtosis they touch.
+// This might change, but right now the engine runs locally, therefore engine interacting endpoint goes to the local
+// backend, and all the rest goes to the remote backend, as all other containers (including API container) should be
+// running in the remote backend
+type RemoteContextKurtosisBackend struct {
```
h4ck3rk3y, line 27: How does this work with MetricsReportingKurtosisBackend? Tbh we are now only reporting metrics from the CLI and if this doesn't support it I am happy with a banner saying it doesn't support; I don't want to you to code the support just curious

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
```diff
@@ -20,9 +23,10 @@ var dockerBackendApiContainerModeArgs *backend_creator.APIContainerModeArgs = ni
 type kurtosisBackendSupplier func(ctx context.Context) (backend_interface.KurtosisBackend, error)
 
 type KurtosisClusterConfig struct {
```
h4ck3rk3y, line 24: can we slap a comment on top on what 

kurtosisBackendSupplier, kurtosisRemoteBackendConfigSupplier & engineBackendConfigSupplier are and used for

h4ck3rk3y: is kurtosisBackendSupplier a local supplier or does it work with kurtosisRemoteBackendConfigSupplier

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
```diff
@@ -77,21 +86,36 @@ func (clusterConfig *KurtosisClusterConfig) GetClusterType() KurtosisClusterType
 func getSuppliers(clusterId string, clusterType KurtosisClusterType, kubernetesConfig *v2.KubernetesClusterConfigV2) (
 	kurtosisBackendSupplier,
 	engine_server_launcher.KurtosisBackendConfigSupplier,
+	*engine_server_launcher.KurtosisRemoteBackendConfigSupplier,
 	error,
 ) {
+	kurtosisRemoteBackendConfigSupplier := engine_server_launcher.NewKurtosisRemoteBackendConfigSupplier(func() (*generated.KurtosisContext, error) {
```
h4ck3rk3y: can we name this anonymous function and pass it to the supplier?

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
h4ck3rk3y: perhaps kurtosisBackendSupplier is `contextAwareKurtosisBackendSupplier` now based on my read of the code below

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
```diff
@@ -133,6 +157,16 @@ func getSuppliers(clusterId string, clusterType KurtosisClusterType, kubernetesC
 		}
 
 		backendSupplier = func(ctx context.Context) (backend_interface.KurtosisBackend, error) {
+			kurtosisRemoteBackendConfigMaybe, err := kurtosisRemoteBackendConfigSupplier.GetOptionalRemoteConfig()
+			if err != nil {
+				return nil, stacktrace.Propagate(err, "Error building optional remote Kurtosis backend config")
+			}
+			if kurtosisRemoteBackendConfigMaybe != nil {
+				return nil, stacktrace.NewError("Using Kubernetes as a local cluster connected to a remote" +
```
h4ck3rk3y: similar comments on the same error above

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
```diff
@@ -4,7 +4,10 @@ import (
 	"context"
```
h4ck3rk3y, line 4: For now the cluster config is affected but eventually it will be swapped our for Kurtosis Context?

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
```diff
@@ -0,0 +1,95 @@
...
+	certFileName       = "cert.pem"
+	keyFileName        = "key.pem"
+	tlsFilesPerm       = 0644
+)
+
+func GetContextAwareKurtosisBackend(
+	remoteBackendConfig *KurtosisRemoteBackendConfig,
+	optionalApiContainerModeArgs *backend_creator.APIContainerModeArgs,
+) (backend_interface.KurtosisBackend, error) {
+	localDockerBackend, err := backend_creator.GetLocalDockerKurtosisBackend(optionalApiContainerModeArgs)
```
h4ck3rk3y, line 29: in a remote backend this is used for engine work 👍

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
h4ck3rk3y: Iff the engine starts running remotely than user doesn't need local docker at all perhaps

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil, stacktrace.Propagate(err, "Error building Kurtosis remote Docker backend")
+	}
+	return newRemoteContextKurtosisBackend(localDockerBackend, kurtosisRemoteBackend), nil
+}
+
+func buildRemoteDockerClient(remoteBackendConfig *KurtosisRemoteBackendConfig) (*client.Client, error) {
+	var clientOptions []client.Opt
+
+	// host and port option
+	url := fmt.Sprintf("%s://%s:%d", urlProtocol, remoteBackendConfig.Host, remoteBackendConfig.Port)
```
h4ck3rk3y: nit: perhaps some of these delimiters can be taken out into constants

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		tlsOpt := client.WithTLSClientConfig(
+			path.Join(tlsFilesDir, caFileName),
+			path.Join(tlsFilesDir, certFileName),
+			path.Join(tlsFilesDir, keyFileName))
+		clientOptions = append(clientOptions, tlsOpt)
+	}
+
+	// API version negotiation option
+	clientOptions = append(clientOptions, client.WithAPIVersionNegotiation())
```
h4ck3rk3y, line 72: what does this do?

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
```diff
@@ -0,0 +1,95 @@
...
+
+	remoteDockerClient, err := client.NewClientWithOpts(clientOptions...)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "Error building Docker remote client")
+	}
+	return remoteDockerClient, nil
+}
+
+func writeTlsConfigToTempDir(ca []byte, cert []byte, key []byte) (string, error) {
+	tempDirectory, err := os.MkdirTemp("", tempDirNamePattern)
```
h4ck3rk3y: perhaps take the "" out in. constant

h4ck3rk3y: who cleans this up? is there a risk that this gets cleaned up by the system while we are still using it?

is this something that should stay with the rest of the context config?

h4ck3rk3y (commented):
Have some questions on the design side / trying to understand the implementation. Otherwise looks good 👍

engine/server/engine/main.go
gbouv: I don't think we want to drop k8s forever. However, I do think running minikube locally might not be super relevant. So, I'm definitely seeing a world where we allow kubernetes as an option for a remote kurtosis backend (though we're not quite there yet). But in the short term, I believe we just want to hide any specificity on what a remote backend is running. 

So, basically, to your bullet points:
1. To run a remote Kurtosis enclave, you have to run docker locally, and have a RemoteContext configured and in use
2. To run local docker kurtosis, you have to run docker locally, and have a local context configured and in use (`default` context comes for free and is local)
3. To run Kurtosis on Kubernetes, you have to get the kubernetes plugin, get minikube setup, have a local context configured and in use, AND manually set your kurtosis cluster to `minikube`. This is the most painful option, but also the one that nobody uses since the kubernetes plugin is private
4. And basically there is no option 4, running Kurtosis on Remote Kubernetes is not supported right now (although you can hack things to get to it, I did it multiple times for Waku)

Does that make sense to you?

engine/launcher/engine_server_launcher/engine_server_launcher.go
gbouv: Yes pretty much. If there `remoteBackendConfigMaybe` is empty here, then the engine will just instantiate a regular docker backend as we currently do

engine/launcher/engine_server_launcher/kurtosis_context_supplier.go
gbouv: Ah, yeah, this is annoying, I tried to change it in a few places. I'll update it here

engine/launcher/args/args.go
gbouv: Can rename it to `KurtosisLocalBackendConfig` and stick a comment above explaining how the two relates to each other 👍

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
gbouv: Can do, but 1. I don't think that's even true. And 2. even if it was, on the long run it might become less and less true

1. It's not true because right now the CLI for example also uses this backend. 
2. We could change it to `localKurtosisBackendForEngineInteraction` and `remoteKurtosisBackendForEnclaveInteraction` because effectively it's what it is, but it is true only because we offload the entire enclave to the remote right now. Not even the APIC is local. However, I think we're soon going to have partial enclaves locally/remotely, and this will change. 

That's basically why I keep a somewhat generic name here. However you'll notice I explain that in the comment above. Let me know if that work, can also rename if you feel strongly about it

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
gbouv: This would be a huge hack, not sure we necessarily want to do it here, but yes we could. I'd bias toward keeping it generic here, at the cost of downloading the images on both client, but I could be convinced otherwise

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
gbouv: I don't think this is incompatible with MetricsReportingKurtosisBackend right now. It works the exact same as a local docker backend. Do you think it's wrong? Whatever the backend is, I was thinking we could track the metric in a similar way

gbouv: And yes, definitely if we want to move the engine remotely, then no need for local docker anymore. That being said, I think we're more tracking keeping some things locally and making that work with the others things running remotely. But that could change. IDK if requiring Docker is a big blocker to adoption

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
gbouv: The whole thing is super messy right now. I think it mostly comes from the fact that the CLI is backend aware. Most of this complexity comes from the fact that the CLI has its own backend, but it also provides the config for the engine to instantiate its backend... -.-
So, basically here 
- kurtosisBackendSupplier is the supplier for the CLI backend
- engineBackendConfigSupplier is the supplier for the config the engine will use to instantiate its own backend
And now we're adding:
- kurtosisRemoteBackendConfigSupplier is the supplier of the config for the remote backend the engine should connect to, if any

I think making the CLI non backend aware will be a big win here. This might be possible now that we have the portal, although it requires everyone to run the portal locally, which will not happen tomorrow. So, basically I'm good not spending too many cycles refactoring this to something better.

But I generally agree it's confusing, took me some time to fully understand those, so I'll stick a good comment here to make sure future us get ramped up on this quickly

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
gbouv: The cluster config itself is not affected at all by this change. It works the exact same.

This is part of the tech debt I mention above. This file is named `kurtosis_cluster_config` what it is used for globally is a `refreshable_kurtosis_backend_provider`, and nothing else.
So, this is why I'm plugging to it to instantiate my remote backend, but my changes have nothing to do with the cluster config. 

That being said, I do agree with you that I see the cluster config becoming outdated fairly soon in favour of different kind of contexts

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
gbouv: Right now, yes pretty much. Might become false in the near future

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
gbouv: It's a standard thing to instantiate a Docker Client. As I understand it you can force the client to talk to only a certain version of Docker, by forcing the version number. This is a more permissive flag where docker client and server enters a negotiation and if they consider they are both mutually compatible, they move forward, otherwise the client fails. Given that docker backward compatibilty is pretty good, this is pretty standard to use this.

FWIW this is what we currently use to build our docker clients

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
gbouv: I don't think we ca clean it up as it is used by the clients (but I can check). That being said, the Os will take care of cleaning it up. `MkdirTemp` as I understand it uses the OS temporary directory. So, it's pretty much equivalent to `mkdir /tmp/temp-dir-for-kurtosis`

engine/server/engine/main.go
h4ck3rk3y: This makes sense!

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
h4ck3rk3y: I meant `forEngineInteraction` as well but I guess yours is more explicit

I think the current thing is alright.

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
h4ck3rk3y: Alright this make sense, I looked at the code further `GetDockerKurtosisBackend` still returns a wrapped backend so things should continue working. Resolving.

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
h4ck3rk3y: Cleanup - Alright, I have seen parts of our code base where we explicitly defer clean temporary resources but this sounds fair.

Risk - Is there a risk that the OS cleans things up while clients still want to use it?

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
h4ck3rk3y: It just seems more simpler and less compute. In the future if we want to support multiple images we could handle that. Like if we need to fetch an image in the APIC but not in the engine we would now fail if it fails to fetch on Engine which it shouldn't IMO?

h4ck3rk3y (approved):

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
gbouv: It's just that we're introducing a behaviour that is not extensible, therefore making future change to this effectively a regression. And when you think about it, this backend is used only by the CLI and the Engine right now. So, everything it will download is the Engine and the APIC image. All user service images will be downloaded by the APIC, which is using a regular docker only backend. So, I'll expand the comment here but leave it this way for now.

contexts-config-store/store/serialization/kurtosis_context.go
gbouv: Yeah it's being tested at the persistence, but will re-shuffle the test to make it explicit it's being tested 👍

container-engine-lib/lib/backend_impls/remote_context_backend/remote_context_kurtosis_backend.go
gbouv: Will double check with Leandro and submit a separate PR to remove it 👍

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
gbouv: Good idea, I think I'll even burry it at the contexts-config-store, seems the right place to put it, and it will give us an easy access to the current context when we need it 👍

gbouv: Oh welp, no need actually you can just do `store.GetContextsConfigStore().GetCurrentContext` 👍

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
gbouv: Yeah, not a huge fan of taking everything out to constant, it gives a format string like this `%s%s%s%s%d` and I find it much much harder to reason about. I'll keep it like this. We do that in other places as well, I think it's fine

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
gbouv: Actually, dug into Docker client codes, and what it does is that it reads the files and then it doesn't need them anymore. I'll add a file cleanup function here to remove the entire folder by ourselves and make sure we keep things in a clean state

PR #298, opened 2023-03-22
docs: Improve runtime value typing documentation

victorcolombo:
## Description:

Right now our `gojq` output is typed and is parsed into its `Starlark` equivalent, but there is not mention to this behavior on the docs. Given that several of our function calls (like `assert`, `wait`) requires types to match, I'm updating the documentation with examples on how to parse types.

YES


docs/docs/reference/starlark-instructions.md
```diff
@@ -385,6 +402,26 @@ plan.assert(response["extract.second-element-from-list-head"], "==", "world")
...
+        "post-number": '.url | split ("/") | .[1]',
+        "post-number-as-int": '.url | split ("/") | .[1] | tonumber',
+    },
+)
+response = plan.request(
+    recipe = post_request_recipe,
+)
+# response["extract.post-number"] is "1" (starlark.String)
+# response["extract.post-number-as-int"] is 1 (starlark.Int)
+```
```
gbouv, line 423: Do we have a link somewhere to jq syntax already? If not, maybe it's a good place to add it. It's pretty advanced jq syntax already, if users stumble upon this and are curious to learn more about jq, having the link here might be handy. Deferring to you

gbouv (approved):

docs/docs/reference/starlark-instructions.md
victorcolombo: Done!

PR #309, opened 2023-03-23
ci: fix concurrency for on pr-title-checker

h4ck3rk3y:
## Description:
We had started using `github.ref` in our concurrency group which would be `refs/head/main` and not something unique to the PR as it would default to the target branch; so my PR might cancel Guillaumes' PR

NO

### DIFF
--- a/.github/workflows/breaking-change-description-title-update.yml
+++ b/.github/workflows/breaking-change-description-title-update.yml
@@ -9,7 +9,7 @@ on:
 
 # cancel redundant builds
 concurrency:
-  group: "update-breaking-change-${{ github.head_ref }}"
+  group: "update-breaking-change-${{ github.ref }}"
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/breaking-change-description-title-update.yml
+++ b/.github/workflows/breaking-change-description-title-update.yml
@@ -9,7 +9,7 @@ on:
 
 # cancel redundant builds
 concurrency:
-  group: "update-breaking-change-${{ github.ref }}"
+  group: "update-breaking-change-${{ github.head_ref }}"
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/conventional-pr-title-checker.yml
+++ b/.github/workflows/conventional-pr-title-checker.yml
@@ -1,20 +1,26 @@
 # Check PR title for conventional commits
 name: Check PR title
 on:
-  pull_request_target:
+  pull_request:
     types:
       - opened
       - reopened
       - edited
       - synchronize
+  push:
+    branches:
+      - gh-readonly-queue/main/**
 
 # cancel redundant builds
 concurrency:
-  group: "title-checker-${{ github.head_ref }}"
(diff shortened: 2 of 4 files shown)


.github/workflows/conventional-pr-title-checker.yml
```diff
@@ -7,20 +7,14 @@ on:
       - reopened
       - edited
       - synchronize
-  push:
-    branches:
-      - gh-readonly-queue/main/**
```
gbouv: Why are removing that? Can't we keep it in case we want to enable merge queues at some point? Took me a bunch of iterations to figure this one out, would be nice to not go trhough them again

.github/workflows/breaking-change-description-title-update.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 # cancel redundant builds
 concurrency:
-  group: "update-breaking-change-${{ github.ref }}"
+  group: "update-breaking-change-${{ github.head_ref }}"
```
gbouv: Not sure that's what you want here, from https://docs.github.com/en/actions/learn-github-actions/contexts ref should be the branch name

.github/workflows/conventional-pr-title-checker.yml
h4ck3rk3y: On the contrary one could look up the git history like I just did?

.github/workflows/breaking-change-description-title-update.yml
h4ck3rk3y: This was what we had earlier; it seems like for `pull_request_target` ref defaults to `main` while for `pull_request` it is unique to the PR. So this job doesn't need head_ref but the pull_request_target job (title check does). Reverting this.

.github/workflows/conventional-pr-title-checker.yml
gbouv: But they would need to know that some changes are required on this file first. Unless they are ok browsing the entire git history, which is very unlikely.
What's our stand with merge queue? Do we want to push it and try to make it work with release please? In which case I'd like to keep that around. If not, then yeah let's remove it

h4ck3rk3y: I have reverted the others and just have the title checker file changed as it uses `pull_request_target` instead of `pull_request` which makes `github.ref` not unique for it.

h4ck3rk3y: If we switch it to just `pull_request` then a forked PR will not have write permissions which feels alright to me. Let me try that out.

h4ck3rk3y: The check writes a status, it seems like the check with `pull_request` passes/fails without the status permissions and with `pull_request` this might just work.

h4ck3rk3y: After this gets merged I'll try to create a PR from a fork and see if that works to verify behavior for outsiders.

gbouv (approved):

PR #311, opened 2023-03-23
ci: make title checker use pull_request_target to allow forked PRs

h4ck3rk3y:
## Description:
Forked PRs break with title checker if we use pull_request instead of pull_request_target

NO

### DIFF
--- a/.github/workflows/conventional-pr-title-checker.yml
+++ b/.github/workflows/conventional-pr-title-checker.yml
@@ -7,9 +7,9 @@ on:
       - reopened
       - edited
       - synchronize
-#  push:
-#    branches:
-#      - gh-readonly-queue/main/**
+  push:
+    branches:
+      - gh-readonly-queue/main/**
 
 # cancel redundant builds
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/conventional-pr-title-checker.yml
```diff
@@ -13,7 +13,7 @@ on:
 
 # cancel redundant builds
 concurrency:
-  group: "title-checker-${{ github.ref }}"
+  group: "title-checker-${{ github.head_ref }}"
```
gbouv: So, if you do that, the merge queue will not work anymore. I think it's important for this file to be consistent. If we keep the push event above, then we need to use something here that is independent of pull request. And head ref is not
Or, we simply remove the push event above, if we can't find something that works for both

h4ck3rk3y: I am using fall back values now, head_ref if available otherwise it uses ref

https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value

gbouv (approved):
Perfect 👏

PR #317, opened 2023-03-23
feat: Add context `switch` CLI command

gbouv:
## Description:
Adding the `kurtosis context switch` command to switch from one context to the other

NO - I don't think we want to communicate on this just yet

### DIFF
--- a/api/golang/engine/lib/kurtosis_context/kurtosis_context.go
+++ b/api/golang/engine/lib/kurtosis_context/kurtosis_context.go
@@ -75,7 +75,7 @@ func NewKurtosisContextFromLocalEngine() (*KurtosisContext, error) {
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "Error retrieving current context")
 	}
-	portalClient, err := BuildPortalDaemonClient(currentContext, true)
+	portalClient, err := CreatePortalDaemonClient(currentContext, true)
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "Error building client for Kurtosis Portal daemon")
 	}
--- a/api/golang/engine/lib/kurtosis_context/kurtosis_context.go
+++ b/api/golang/engine/lib/kurtosis_context/kurtosis_context.go
@@ -11,8 +11,6 @@ import (
 	"github.com/kurtosis-tech/kurtosis/api/golang/core/lib/services"
 	"github.com/kurtosis-tech/kurtosis/api/golang/engine/kurtosis_engine_rpc_api_bindings"
 	"github.com/kurtosis-tech/kurtosis/api/golang/kurtosis_version"
-	contexts_store_api "github.com/kurtosis-tech/kurtosis/contexts-config-store/api/golang"
-	contexts_store_generated_api "github.com/kurtosis-tech/kurtosis/contexts-config-store/api/golang/generated"
 	"github.com/kurtosis-tech/kurtosis/contexts-config-store/store"
 	"github.com/kurtosis-tech/stacktrace"
 	"github.com/sirupsen/logrus"
@@ -30,8 +28,6 @@ const (
 	DefaultGrpcEngineServerPortNum = uint16(9710)
 
 	DefaultGrpcProxyEngineServerPortNum = uint16(9711)
-
-	DefaultGrpcPortalClientPortNum = uint16(9731)
 
 	// Blank tells the engine server to use the default
 	defaultApiContainerVersionTag = ""
@@ -79,34 +75,7 @@ func NewKurtosisContextFromLocalEngine() (*KurtosisContext, error) {
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "Error retrieving current context")
 	}
-	var portalClient portal_api.KurtosisPortalClientClient
-	instantiatePortalClientVisitor := contexts_store_api.KurtosisContextVisitor[struct{}]{
-		VisitLocalOnlyContextV0: func(localContext *contexts_store_generated_api.LocalOnlyContextV0) (*struct{}, error) {
-			portalClient = nil
-			return nil, nil
-		},
-		VisitRemoteContextV0: func(remoteContext *contexts_store_generated_api.RemoteContextV0) (*struct{}, error) {
-			// When the context is remote, we build a client to the locally running portal daemon
-			kurtosisPortalSocketStr := fmt.Sprintf("%v:%v", localHostIPAddressStr, DefaultGrpcPortalClientPortNum)
-			// TODO SECURITY: Use HTTPS to ensure we're connecting to the real Kurtosis API servers
-			portalConn, err := grpc.Dial(kurtosisPortalSocketStr, grpc.WithInsecure())
-			if err != nil {
-				return nil, stacktrace.Propagate(
-					err,
-					"An error occurred creating a connection to the Kurtosis Portal Client at '%v'",
-					kurtosisPortalSocketStr,
-				)
-			}
-			portalClient = portal_api.NewKurtosisPortalClientClient(portalConn)
-
-			if _, err = portalClient.Ping(context.Background(), portal_constructors.NewPortalPing()); err != nil {
-				return nil, stacktrace.Propagate(err, "Unable to connect to the locally running Kurtosis portal daemon."+
-					" To use remote context, a Kurtosis Portal daemon must be running on this machine")
-			}
-			return nil, nil
-		},
-	}
-	_, err = contexts_store_api.Visit[struct{}](currentContext, instantiatePortalClientVisitor)
+	portalClient, err := BuildPortalDaemonClient(currentContext, true)
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "Error building client for Kurtosis Portal daemon")
 	}
--- a/api/golang/engine/lib/kurtosis_context/portal_daemon_client.go
+++ b/api/golang/engine/lib/kurtosis_context/portal_daemon_client.go
@@ -1,0 +1,66 @@
+package kurtosis_context
+
+import (
+	"context"
+	"fmt"
+	portal_constructors "github.com/kurtosis-tech/kurtosis-portal/api/golang/constructors"
+	portal_api "github.com/kurtosis-tech/kurtosis-portal/api/golang/generated"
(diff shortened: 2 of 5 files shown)


api/golang/engine/lib/kurtosis_context/kurtosis_context.go
```diff
@@ -79,34 +75,7 @@ func NewKurtosisContextFromLocalEngine() (*KurtosisContext, error) {
...
-			portalClient = portal_api.NewKurtosisPortalClientClient(portalConn)
-
-			if _, err = portalClient.Ping(context.Background(), portal_constructors.NewPortalPing()); err != nil {
-				return nil, stacktrace.Propagate(err, "Unable to connect to the locally running Kurtosis portal daemon."+
-					" To use remote context, a Kurtosis Portal daemon must be running on this machine")
-			}
-			return nil, nil
-		},
-	}
-	_, err = contexts_store_api.Visit[struct{}](currentContext, instantiatePortalClientVisitor)
```
gbouv, line 109: I just refactored this into portal_daemon_client file

api/golang/engine/lib/kurtosis_context/portal_daemon_client.go
```diff
@@ -0,0 +1,66 @@
...
+
+// BuildPortalDaemonClient builds a portal daemon GRPC client based on the current context and a
+// forLocalContextReturnNilIfUnreachable flag
+// If the flag is set to true, it returns a nil client for local contexts when the portal is not reachable.
+// Local context are working seamlessly without the portal (unless the user wants to map port). Therefore we're not
+// requiring it for most workflows
+// TODO: when it's assumed that Kurtosis Portal should be running everywhere, we can remove this logic and just build a
+// regular client
+func BuildPortalDaemonClient(currentContext *contexts_store_generated_api.KurtosisContext, forLocalContextReturnNilIfUnreachable bool) (portal_api.KurtosisPortalClientClient, error) {
+	// When the context is remote, we build a client to the locally running portal daemon
```
leoporoli, line 27: I suggest checking `forLocalContextReturnNilIfUnreachable` in the first line for an early return

leoporoli: you can ommit the comment, now I understood that we need two conditions, an unreachable portal && the flag

api/golang/engine/lib/kurtosis_context/portal_daemon_client.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil, stacktrace.Propagate(
+			err,
+			"An error occurred creating a connection to the Kurtosis Portal Client at '%v'",
+			kurtosisPortalSocketStr,
+		)
+	}
+	portalClient := portal_api.NewKurtosisPortalClientClient(portalConn)
+	_, portalReachableError := portalClient.Ping(context.Background(), portal_constructors.NewPortalPing())
+	instantiatePortalClientVisitor := contexts_store_api.KurtosisContextVisitor[struct{}]{
+		VisitLocalOnlyContextV0: func(localContext *contexts_store_generated_api.LocalOnlyContextV0) (*struct{}, error) {
```
leoporoli: nit: I suggest putting the functions into vars to improve legibility

api/golang/engine/lib/kurtosis_context/portal_daemon_client.go
```diff
@@ -0,0 +1,66 @@
...
+			if portalReachableError != nil && forLocalContextReturnNilIfUnreachable {
+				logrus.Infof("Portal daemon unreachable on port '%d' but will not fail as the context in use is a local context: '%s'", DefaultGrpcPortalClientPortNum, currentContext.GetName())
+				// error is allowed here, overriding the portalClient to nil as we can't connect to it but this is
+				// more or less expected based on the forLocalContextReturnNilIfUnreachable flag
+				portalClient = nil
+				return nil, nil
+			}
+			if portalReachableError != nil {
+				return nil, portalReachableError
```
leoporoli: why you omitted the stack trace here?

api/golang/engine/lib/kurtosis_context/portal_daemon_client.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+			return nil, nil
+		},
+		VisitRemoteContextV0: func(remoteContext *contexts_store_generated_api.RemoteContextV0) (*struct{}, error) {
+			if portalReachableError != nil {
+				return nil, portalReachableError
```
leoporoli: same here about stack trace

api/golang/engine/lib/kurtosis_context/portal_daemon_client.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// BuildPortalDaemonClient builds a portal daemon GRPC client based on the current context and a
+// forLocalContextReturnNilIfUnreachable flag
+// If the flag is set to true, it returns a nil client for local contexts when the portal is not reachable.
+// Local context are working seamlessly without the portal (unless the user wants to map port). Therefore we're not
+// requiring it for most workflows
+// TODO: when it's assumed that Kurtosis Portal should be running everywhere, we can remove this logic and just build a
+// regular client
+func BuildPortalDaemonClient(currentContext *contexts_store_generated_api.KurtosisContext, forLocalContextReturnNilIfUnreachable bool) (portal_api.KurtosisPortalClientClient, error) {
```
leoporoli: just for curiosity and learning purposes, Why you decided to call it `Build` and not `New` or `Create`?

cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error {
+	contextIdentifier, err := args.GetNonGreedyArg(contextIdentifierArgKey)
+	if err != nil {
+		return stacktrace.Propagate(err, "Expected a value for context identifier arg '%v' but none was found; this is a bug with Kurtosis!", contextIdentifierArgKey)
+	}
+
+	isContextSwitchSuccessful := false
+	logrus.Info("Switching context...")
```
leoporoli, line 48: nice!

cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -0,0 +1,98 @@
...
+
+	contextsConfigStore := store.GetContextsConfigStore()
+	contextPriorToSwitch, err := contextsConfigStore.GetCurrentContext()
+	if err != nil {
+		return stacktrace.NewError("An error occurred retrieving current context prior to switching to the new one '%s'", contextIdentifier)
+	}
+
+	contextsMatchingIdentifier, err := context_id_arg.GetContextUuidForContextIdentifier(contextsConfigStore, []string{contextIdentifier})
```
leoporoli: nit: rename to `contextsMatchingIdentifiers`

cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -0,0 +1,98 @@
...
+				contextIdentifier, contextUuidToSwitchTo.GetValue(), contextPriorToSwitch.GetName(),
+				contextPriorToSwitch.GetUuid(), contextIdentifier, contextPriorToSwitch.GetName(), err.Error())
+		}
+	}()
+
+	currentContext, err := contextsConfigStore.GetCurrentContext()
+	if err != nil {
+		return stacktrace.Propagate(err, "Error retrieving context info for context '%s' after switching to it", contextIdentifier)
+	}
+	portalDaemonClient, err := kurtosis_context.BuildPortalDaemonClient(currentContext, true)
```
leoporoli: nit: you could put the second argument in a constant

cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return stacktrace.Propagate(err, "Error connecting to Kurtosis portal after switching to the context'%s'", contextIdentifier)
+	}
+
+	if portalDaemonClient != nil {
+		if _, err = portalDaemonClient.SwitchContext(ctx, constructors.NewSwitchContextArgs()); err != nil {
```
leoporoli: nit: same here you could put the second argument into a ver

leoporoli (approved):

api/golang/engine/lib/kurtosis_context/portal_daemon_client.go
gbouv: it's because that the error returned by the visitor, not the error returned by the function. If you take a look, the error returned by the visitor (i.e. this error) is wrapped into a stacktrace.Propagate line 63. Wrapping it here would basically correspond to wrapping it twice in the same place, which seems unnecessary
Will add a comment to clarify this

api/golang/engine/lib/kurtosis_context/portal_daemon_client.go
gbouv: Same response, that's the error returned by the visitor, wrapped right below

api/golang/engine/lib/kurtosis_context/portal_daemon_client.go
gbouv: Good call here, I guess I didn't think too much about the name of this function. Based on what we already discussed in previous reviews, I'll rename it `Create*`

gbouv: So, yes, I reserve `New` and `Create` for struct constructors (Where New is a dumb constructor returning the struct, and Create a more evolved one which can fail and return an error).
Here, I don't consider this method a constructor. First, it's not implemented next to the struct it returns. Second, it does a fair amount of logic, with this specific flag to not fail in certain condition. So, to me this is more a helper method to "build" a client to the portal.
I would be open to rename it to `Create` here (not New though), as it _could_ be seen as a distant constructor. But I personally feel better about Build here

gbouv: (resolving to unblock merging but happy to discuss more it you think it's worth it)

PR #322, opened 2023-03-24
fix: Fix starlark value reference bug

gbouv:
## Description:
Running:
```
def run(plan):
	my_dict = {}
	plan.print(my_dict)
	my_dict["hello"] = "world"
```

right now returns:
```
> print msg={"hello": "world"}
{"hello": "world"}
```

This is because of the 3 steps execution, combined with the fact that "complex" starlark.Value object are passed as pointers to our instructions. This PR fixes it by copying the arg value into each instruction at the interpretation phase.

NO


core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/builtin_argument/argument_value_copier_test.go
```diff
@@ -0,0 +1,134 @@
...
+		starlark.None,
+	})))
+
+	valueCopy, err := DeepCopyArgumentValue(value)
+	require.NoError(t, err)
+	require.Equal(t, valueCopy, value)
+	require.NotSame(t, valueCopy, value)
+}
+
+func TestCopyValueStruct(t *testing.T) {
```
victorcolombo, line 120: Can we have this for all named Kurtosis defined structs?

victorcolombo (approved):

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/builtin_argument/argument_value_copier_test.go
gbouv: 👍  Added it inside the framework test engine

PR #329, opened 2023-03-27
feat: Automatically restart engine on context switch

gbouv:
## Description:
Engine is automatically restarted when context is switched.

NO

### DIFF
--- a/cli/cli/commands/kurtosis_context/context_switch/switch.go
+++ b/cli/cli/commands/kurtosis_context/context_switch/switch.go
@@ -103,11 +103,14 @@ func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error
 	logrus.Infof("Context switched to '%s', Kurtosis engine will now be restarted", contextIdentifier)
 
 	if err = engineManager.StopEngineIdempotently(ctx); err != nil {
-		return stacktrace.Propagate(err, "Unable to stop the Kurtosis engine currently running. Context will be rolled back to '%s'", contextPriorToSwitch.GetName())
+		return stacktrace.Propagate(err, "Unable to stop the Kurtosis engine currently running. Context will be "+
+			"rolled back to '%s'", contextPriorToSwitch.GetName())
 	}
 	_, _, err = engineManager.StartEngineIdempotentlyWithDefaultVersion(ctx, logrus.InfoLevel)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -94,8 +100,17 @@ func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error
...
 		}
 	}
+	logrus.Infof("Context switched to '%s', Kurtosis engine will now be restarted", contextIdentifier)
+
+	if err = engineManager.StopEngineIdempotently(ctx); err != nil {
+		return stacktrace.Propagate(err, "Unable to stop the Kurtosis engine currently running. Context will be rolled back to '%s'", contextPriorToSwitch.GetName())
+	}
+	_, _, err = engineManager.StartEngineIdempotentlyWithDefaultVersion(ctx, logrus.InfoLevel)
+	if err != nil {
+		return stacktrace.Propagate(err, "Unable to start a new Kurtosis engine. Context will be rolled back to '%s' and the Kurtosis engine will remain stopped", contextPriorToSwitch.GetName())
```
h4ck3rk3y: Should we tell the user how to restart the engine?

If the user is already playing with contexts maybe they know how engine restart works so I'm fine as is

cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -94,8 +100,17 @@ func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error
 			return stacktrace.Propagate(err, "Error switching Kurtosis portal context")
 		}
 	}
+	logrus.Infof("Context switched to '%s', Kurtosis engine will now be restarted", contextIdentifier)
+
+	if err = engineManager.StopEngineIdempotently(ctx); err != nil {
```
h4ck3rk3y, line 105: Curious why we need to restart engines? Does that create a new backend?

h4ck3rk3y (approved):

cli/cli/commands/kurtosis_context/context_switch/switch.go
gbouv: Yes, right now both the CLI and the engine have the "dual-backend" engine. The CLI needs it to interact with the remote enclave, the engine needs it to start the APIC running remotely :)

cli/cli/commands/kurtosis_context/context_switch/switch.go
gbouv: I can add!

PR #333, opened 2023-03-27
feat: add a CLI command to dump all of Kurtosis

### LINKED ISSUE #305: [FR]: Build `kurtosis dump` to dump the entire state of the Kurtosis system

During [a user experience session today](https://www.notion.so/kurtosistech/Cam-Sinko-Kurtosis-649a527030764b96bde916b55453b69e?pvs=4) we were debugging some very thorny issues, and didn't have a good way to get insight into what was going wrong inside Kurtosis itself. We had to resort to using `docker` commands to try and get the logs, which proved to be a huge pain. Furthermore, the user didn't have a good way to get those logs to us (he eventually copy-pasted them into an email body, which lost all the formatting). It would have been much, much nicer if we could get a package of the state of the system, so that we could debug it in our side.

### Desired behavior

`kurtosis dump` will create a directory of the entire state of the Kurtosis system:
- Engine: inspect output + logs (same as `enclave dump` does for each service)
- `enclave dump` for each enclave in Kurtosis

This command could be what we upload to Github issues via `kurtosis feedback --bug`

In the future, when we capture CLI commands + logs, we can also include these.

Critical, Kurtosis is unusable for us without it.

h4ck3rk3y:
YES/NO


cli/cli/command_str_consts/command_str_consts.go
```diff
@@ -44,6 +44,7 @@ const (
 	FilesStoreWebCmdStr     = "storeweb"
 	FilesStoreServiceCmdStr = "storeservice"
 	FilesRenderTemplate     = "rendertemplate"
+	KurtosisDumpCmdStr      = "dump"
```
h4ck3rk3y, line 47: needs docs!!!!!!

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/dump_kurtosis.go
```diff
@@ -0,0 +1,57 @@
...
+	if err = backend.GetEngineLogs(ctx, engineOutputDir); err != nil {
+		return stacktrace.Propagate(err, "An error occurred while dumping engine logs to dir '%v'", err)
+	}
+
+	allEnclavesOutputSubdir := path.Join(outputDirpath, enclavesSubDirpathFragment)
+	if err = os.Mkdir(allEnclavesOutputSubdir, createdDirPerms); err != nil {
+		return stacktrace.Propagate(err, "An error occurred creating output directory for all enclaves at '%v'", outputDirpath)
+	}
+
+	// perhaps dont error immediately and dump what you can before sending errors
```
h4ck3rk3y: TODO convert this to best effort

h4ck3rk3y:
Uploading Screen Recording 2023-03-27 at 6.05.46 PM.mov…

cli/cli/commands/dump/dump.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+var KurtosisDump = &engine_consuming_kurtosis_command.EngineConsumingKurtosisCommand{
+	CommandStr:                command_str_consts.KurtosisDumpCmdStr,
+	ShortDescription:          "Dumps entire Kurtosis State",
+	LongDescription:           "Dumps entire Kurtosis State to the given directory",
+	KurtosisBackendContextKey: kurtosisBackendCtxKey,
+	EngineClientContextKey:    engineClientCtxKey,
+	Flags:                     nil,
+	Args: []*args.ArgConfig{
```
gbouv, line 34: Maybe we should add a `--enclave` filter here, such that we can eventually deprecate the enclave dump one. Wdyt? (as a FLUP, of course)

cli/cli/commands/dump/dump.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	outputDirpathArg = "output-dirpath"
+
+	kurtosisBackendCtxKey = "kurtosis-backend"
+	engineClientCtxKey    = "engine-client"
+
+	defaultKurtosisDumpDir = "kurtosis-dump"
```
gbouv, line 22: Are we worried this will conflict? Should we add a random string of something out of cautiousness?

gbouv: (or make outputDirpathArg a required arg, but I also prefer it as optional I think)

cli/cli/commands/root.go
```diff
@@ -109,6 +110,7 @@ func init() {
 	RootCmd.AddCommand(service.ServiceCmd)
 	RootCmd.AddCommand(twitter.TwitterCmd.MustGetCobraCommand())
 	RootCmd.AddCommand(version.VersionCmd)
+	RootCmd.AddCommand(kurtosisdump.KurtosisDump.MustGetCobraCommand())
```
gbouv: Can we alphabetically sort that?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -138,7 +138,11 @@ func (backend *DockerKurtosisBackend) DestroyEngines(
 }
 
 func (backend *DockerKurtosisBackend) GetEngineLogs(ctx context.Context, outputDirpath string) error {
-	return engine_functions.EngineLogs(ctx, outputDirpath, backend.dockerManager)
+	return engine_functions.GetEngineLogs(ctx, outputDirpath, backend.dockerManager)
+}
+
+func (backend *DockerKurtosisBackend) DumpKurtosis(ctx context.Context, outputDirpath string) error {
```
gbouv, line 144: I guess you'll need to update the k8s plugin here

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/dump_kurtosis.go
```diff
@@ -0,0 +1,69 @@
...
+
+func DumpKurtosis(ctx context.Context, outputDirpath string, backend backend_interface.KurtosisBackend) error {
+
+	allEnclaves, err := backend.GetEnclaves(ctx, allEnclavesFilter)
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred while getting a list of enclaves registered with the underlying engine")
+	}
+
+	// Create the main output directory
+	if _, err = os.Stat(outputDirpath); !os.IsNotExist(err) {
```
gbouv, line 33: Looking at the code, `os.IsNotExist` I think will panic when err is nil here. But also surprised you didn't catch that in your tests, so I might be reading this wrong. Anyways, can you maybe just do `err != nil && !os.IsNotExist(err)` to avoid any confusion ?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/dump_kurtosis.go
```diff
@@ -0,0 +1,69 @@
...
+		return stacktrace.NewError("Cannot create output directory at '%v'; directory already exists", outputDirpath)
+	}
+	if err = os.Mkdir(outputDirpath, createdDirPerms); err != nil {
```
gbouv, line 37: So, you call MkDir even when the dir already exist? Will that no-op in this case? And if it no-ops, will we collocate 2 dumps in the same folder, isn't this confusing?
Maybe add a comment clarifying the expected behaviour here, for future readers

gbouv (approved):
Mostly minor comments, but curious on what happens when you run dump twice to the same place:
```
kurtosis dump dump-folder
kurtosis dump dump-folder
```
Will that dump the entire things twice in the same folder? Is that expected? Maybe we should error if the directory isn't empty or something?

cli/cli/commands/dump/dump.go
h4ck3rk3y: added a TODO above the args as a comment but this does both engine & enclave dumps

cli/cli/commands/dump/dump.go
h4ck3rk3y: If people use the same dir then it fails! I could attach a random string but that seems more confusing. Thoughts?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/dump_kurtosis.go
h4ck3rk3y: I copied another code path in our code when I wrote this!

https://go.dev/play/p/yCZQczVivLX

Seems like IsNotExist doesn't panic; I can add err != nil for reducing confusion

PR #334, opened 2023-03-27
feat: Add portal `status`, `start` and `stop` command

gbouv:
## Description:
Add minimal set of lifecycle command to manage Kurtosis Portal. `start` makes sure to download the Portal binary if it's not present yet.

YES - but similar to context, might not want to communicate on this just yet


cli/cli/helpers/host_machine_directories/host_machine_directories.go
```diff
@@ -106,6 +112,42 @@ func GetKurtosisCliLogsFilePath() (string, error) {
 	return kurtosisCliLogFilePath, nil
 }
 
+func GetPortalBinaryFilePath() (string, error) {
+	xdgRelFilepath := getRelativeFilepathForPortalForXDG(portalBinaryFilename)
+	portalBinaryFilePath, err := xdg.DataFile(xdgRelFilepath)
+	if err != nil {
+		return "", stacktrace.Propagate(err, "An error occurred getting the last pestered user about old version file path using '%v'", xdgRelFilepath)
```
h4ck3rk3y: change the error message here

cli/cli/helpers/host_machine_directories/host_machine_directories.go
```diff
@@ -106,6 +112,42 @@ func GetKurtosisCliLogsFilePath() (string, error) {
...
+	}
+	return portalBinaryFilePath, nil
+}
+
+func GetPortalLogFilePath() (string, error) {
```
h4ck3rk3y, line 124: change the error message here

cli/cli/helpers/host_machine_directories/host_machine_directories.go
```diff
@@ -106,6 +112,42 @@ func GetKurtosisCliLogsFilePath() (string, error) {
...
+		return "", stacktrace.Propagate(err, "An error occurred getting the last pestered user about old version file path using '%v'", xdgRelFilepath)
+	}
+	return portalBinaryFilePath, nil
+}
+
+func GetPortalVersionFilePath() (string, error) {
+	xdgRelFilepath := getRelativeFilepathForPortalForXDG(portalVersionFilename)
+	portalVersionFilePath, err := xdg.DataFile(xdgRelFilepath)
+	if err != nil {
+		return "", stacktrace.Propagate(err, "An error occurred getting the last pestered user about old version file path using '%v'", xdgRelFilepath)
```
h4ck3rk3y: change the error message here

cli/cli/helpers/host_machine_directories/host_machine_directories.go
```diff
@@ -106,6 +112,42 @@ func GetKurtosisCliLogsFilePath() (string, error) {
...
+	}
+	return portalVersionFilePath, nil
+}
+
+func GetPortalPidFilePath() (string, error) {
+	xdgRelFilepath := getRelativeFilepathForPortalForXDG(portalPidFilename)
+	portalPidFilePath, err := xdg.StateFile(xdgRelFilepath)
+	if err != nil {
+		return "", stacktrace.Propagate(err, "An error occurred getting the last pestered user about old version file path using '%v'", xdgRelFilepath)
```
h4ck3rk3y: change the error message here

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	currentVersionFilePath, err := host_machine_directories.GetPortalVersionFilePath()
+	if err != nil {
+		return false, stacktrace.Propagate(err, "Unable to get file path to Kurtosis Portal version file")
+	}
+
+	ghClient := github.NewClient(http.DefaultClient)
+	latestRelease, err := getLatestRelease(ctx, ghClient)
+	if err != nil {
+		return false, stacktrace.Propagate(err, "Unable to retrieve latest release of Kurtosis Portal from Github repository")
```
h4ck3rk3y: should this always error? I mean can we run with old version and shoot a warning?

s/Github/GitHub

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	latestVersionStr, currentVersionStr, currentVersionMatchesLatest, err := compareLatestVersionWithCurrent(currentVersionFilePath, latestRelease)
+	if err != nil {
+		return false, stacktrace.Propagate(err, "Error checking if latest version matches current")
```
h4ck3rk3y: similar here; maybe someone is connected to their intranet but have troubles with internet/GitHub is down.

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+
+	if currentVersionStr == "" {
+		logrus.Infof("Installing Kurtosis Portal version '%s'", latestVersionStr)
+	} else {
+		logrus.Infof("Upgrading currently Kurtosis Portal version '%s' to '%s'", currentVersionStr, latestVersionStr)
+	}
+
+	githubAssetContent, err := downloadGithubAsset(ctx, ghClient, latestRelease)
+	if err != nil {
+		return false, stacktrace.Propagate(err, "Unable to browse Kurtosis Portal released assets from Github to download the latest")
```
h4ck3rk3y: s/Github/GitHub/g

same concerns about why we are erroring instead of warning/logging an error

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+		return false, stacktrace.Propagate(err, "Unable to get file path to Kurtosis Portal version file")
+	}
+
+	ghClient := github.NewClient(http.DefaultClient)
+	latestRelease, err := getLatestRelease(ctx, ghClient)
+	if err != nil {
+		return false, stacktrace.Propagate(err, "Unable to retrieve latest release of Kurtosis Portal from Github repository")
+	}
+
+	latestVersionStr, currentVersionStr, currentVersionMatchesLatest, err := compareLatestVersionWithCurrent(currentVersionFilePath, latestRelease)
```
h4ck3rk3y: should this be "maybeCurrentVersionStr"

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return false, stacktrace.Propagate(err, "Error checking if latest version matches current")
+	}
+	if currentVersionMatchesLatest {
+		logrus.Infof("Kurtosis Portal version '%s' is the latest and already installed", latestVersionStr)
+		return false, nil
+	}
+
+	if currentVersionStr == "" {
```
h4ck3rk3y: can you put a comment on whats going on here; 

ah i get it "" means no portal. maybe a constant is enough and no comment is needed.

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	} else {
+		logrus.Infof("Upgrading currently Kurtosis Portal version '%s' to '%s'", currentVersionStr, latestVersionStr)
+	}
+
+	githubAssetContent, err := downloadGithubAsset(ctx, ghClient, latestRelease)
+	if err != nil {
+		return false, stacktrace.Propagate(err, "Unable to browse Kurtosis Portal released assets from Github to download the latest")
+	}
+
+	if err = extractBinaryToFile(githubAssetContent, latestVersionStr, binaryFilePath, currentVersionFilePath); err != nil {
```
h4ck3rk3y: what does extracting a binary file mean? maybe 

extractGzippedBinaryToFileOnDisk maybe

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+		return false, stacktrace.Propagate(err, "Unable to extract Kurtosis Portal binary file content to file on disk")
```
h4ck3rk3y: perhaps talk about it being zipped

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return true, nil
+}
+
+func getLatestRelease(ctx context.Context, ghClient *github.Client) (*github.RepositoryRelease, error) {
+	// First, browse the list of releases for the Kurtosis Portal repo
+	latestRelease, _, err := ghClient.Repositories.GetLatestRelease(ctx, kurtosisTechGithubOrg, kurtosisPortalGithubRepoName)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "Unable to retrieve latest version of Kurtosis Portal form Github")
```
h4ck3rk3y: s/Github/GitHub/g

wont mention this anymore. Learned this from a previous PR by Leandro

cli/cli/helpers/portal_manager/file_helpers.go
```diff
@@ -0,0 +1,24 @@
+package portal_manager
+
+import (
+	"github.com/kurtosis-tech/stacktrace"
+	"os"
+)
+
+// CreateFileIfNecessary creates the file if it doesn't already exist. If it does, do nothing.
+// Return true if the file was created, false if it existed already, and error if something went wrong
+func CreateFileIfNecessary(filePath string) (bool, error) {
```
h4ck3rk3y: can this be private for now? I couldn't find an invocation for it outside of the package.

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+		return latestVersionStr, currentVersionStr, true, nil
+	}
+	return latestVersionStr, currentVersionStr, false, nil
+}
+
+func downloadGithubAsset(ctx context.Context, ghClient *github.Client, latestRelease *github.RepositoryRelease) (io.ReadCloser, error) {
+	// Get all assets associated with this release and identify the one matching the current machine architecture
+	opts := &github.ListOptions{
+		Page:    0,
+		PerPage: 50, // we will probably never have more than 10 assets per release, hopefully...
```
h4ck3rk3y: lets push these to constants

cli/cli/commands/portal/portal.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/kurtosis-tech/kurtosis/cli/cli/commands/portal/status"
+	"github.com/kurtosis-tech/kurtosis/cli/cli/commands/portal/stop"
+	"github.com/spf13/cobra"
+)
+
+// ContextCmd Suppressing exhaustruct requirement because this struct has ~40 properties
+// nolint: exhaustruct
+var PortalCmd = &cobra.Command{
+	Use:   command_str_consts.PortalCmdStr,
+	Short: "Manage lifecycle of Kurtosis Portal",
```
h4ck3rk3y: nit: add a Long and explain what portal is; for now or link to docs

cli/cli/commands/portal/start/start.go
```diff
@@ -0,0 +1,53 @@
...
+	Flags:                    []*flags.FlagConfig{},
+	Args:                     []*args.ArgConfig{},
+	PreValidationAndRunFunc:  nil,
+	RunFunc:                  run,
+	PostValidationAndRunFunc: nil,
+}
+
+func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error {
+	logrus.Infof("Starting Kurtosis Portal")
+	portalManager := portal_manager.NewPortalManager()
```
h4ck3rk3y, line 27: Maybe later we can add a portal consuming kurtosis command

cli/cli/commands/portal/start/start.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/kurtosis-tech/kurtosis/cli/cli/command_str_consts"
+	"github.com/kurtosis-tech/kurtosis/cli/cli/helpers/portal_manager"
+	"github.com/kurtosis-tech/stacktrace"
+	"github.com/sirupsen/logrus"
+)
+
+var PortalStartCmd = &lowlevel.LowlevelKurtosisCommand{
+	CommandStr:               command_str_consts.PortalStartCmdStr,
+	ShortDescription:         "Starts Kurtosis Portal",
+	LongDescription:          "Starts Kurtosis Portal in the background. The portal can then be stopped or restarted using the corresponding commands",
```
h4ck3rk3y: s/portal/Portal

cli/cli/commands/portal/start/start.go
```diff
@@ -0,0 +1,53 @@
...
+	PostValidationAndRunFunc: nil,
+}
+
+func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error {
+	logrus.Infof("Starting Kurtosis Portal")
+	portalManager := portal_manager.NewPortalManager()
+
+	// Checking if new version is available and downloading it
+	if _, err := portal_manager.DownloadLatestKurtosisPortalBinary(ctx); err != nil {
+		return stacktrace.Propagate(err, "Unable to download Kurtosis Portal binary")
```
h4ck3rk3y: do we always want to run portal with an active internet connection ; the server can be intranet

GitHub can be down. Can keep this for now if it makes life easier with versioning and such. 

Curious, will new portal work with any version of Kurtosis that has ever supported portal?

cli/cli/commands/portal/start/start.go
```diff
@@ -0,0 +1,53 @@
...
+	_, process, isPortalReachable, err := portalManager.CurrentStatus(ctx)
+	if err != nil {
+		return stacktrace.Propagate(err, "Unable to determine current state of Kurtosis Portal process")
+	}
+	if isPortalReachable {
+		logrus.Infof("Portal is currently running and healthy.")
+		return nil
+	}
+	if process != nil {
+		logrus.Infof("A non-healthy Portal process is currently running. Stop it first before starting a new one")
```
h4ck3rk3y: Warn perhaps

Possible FLUP a restart command

cli/cli/commands/portal/start/start.go
```diff
@@ -0,0 +1,53 @@
...
+	if _, err := portal_manager.DownloadLatestKurtosisPortalBinary(ctx); err != nil {
+		return stacktrace.Propagate(err, "Unable to download Kurtosis Portal binary")
+	}
+
+	_, process, isPortalReachable, err := portalManager.CurrentStatus(ctx)
+	if err != nil {
+		return stacktrace.Propagate(err, "Unable to determine current state of Kurtosis Portal process")
+	}
+	if isPortalReachable {
+		logrus.Infof("Portal is currently running and healthy.")
```
h4ck3rk3y: Infof or Info?

cli/cli/commands/portal/start/start.go
```diff
@@ -0,0 +1,53 @@
...
+	if process != nil {
+		logrus.Infof("A non-healthy Portal process is currently running. Stop it first before starting a new one")
+		return nil
+	}
+
+	pid, err := portalManager.StartNew(ctx)
+	if err != nil {
+		return stacktrace.Propagate(err, "Error starting portal")
+	}
+	logrus.Infof("Kurtosis portal started successfully on PID %d", pid)
```
h4ck3rk3y: s/portal/Portal

cli/cli/commands/portal/status/status.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/kurtosis-tech/kurtosis/cli/cli/command_framework/lowlevel/flags"
+	"github.com/kurtosis-tech/kurtosis/cli/cli/command_str_consts"
+	"github.com/kurtosis-tech/kurtosis/cli/cli/helpers/portal_manager"
+	"github.com/kurtosis-tech/stacktrace"
+	"github.com/sirupsen/logrus"
+)
+
+var PortalStatusCmd = &lowlevel.LowlevelKurtosisCommand{
+	CommandStr:               command_str_consts.PortalStatusCmdStr,
+	ShortDescription:         "Displays the status of Kurtosis Portal",
```
h4ck3rk3y: s/of Kurtosis/ of the Kurtosis

cli/cli/commands/portal/status/status.go
```diff
@@ -0,0 +1,45 @@
...
+	LongDescription:          "Determines and displays the status of the Kurtosis Portal process running locally",
```
h4ck3rk3y: Do we have to mention "locally" or can we mention if this is Daemon/Client

cli/cli/commands/portal/status/status.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	if isPortalReachable {
+		logrus.Infof("Kurtosis Portal is running and healthy on PID %d", pid)
+		return nil
+	} else if process != nil {
+		logrus.Infof("Kurtosis Portal process is on PID %d but it is unreachable", pid)
+		return nil
+	} else if pid != 0 {
+		logrus.Infof("Kurtosis Portal PID file exits but process is dead")
```
h4ck3rk3y: Infof or Info

cli/cli/commands/portal/status/status.go
```diff
@@ -0,0 +1,45 @@
...
+		return nil
+	}
+	logrus.Infof("Kurtosis Portal is not running")
```
h4ck3rk3y: This differs a bit from engine status which prints on terminal instead of using logging.

I'm fine with either but maybe we should be consistent. Which one do you prefer?

cli/cli/commands/portal/stop/stop.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/kurtosis-tech/kurtosis/cli/cli/command_str_consts"
+	"github.com/kurtosis-tech/kurtosis/cli/cli/helpers/portal_manager"
+	"github.com/kurtosis-tech/stacktrace"
+	"github.com/sirupsen/logrus"
+)
+
+var PortalStopCmd = &lowlevel.LowlevelKurtosisCommand{
+	CommandStr:               command_str_consts.PortalStopCmdStr,
+	ShortDescription:         "Stops Kurtosis Portal",
+	LongDescription:          "Stops Kurtosis Portal if it is currently running",
```
h4ck3rk3y: Can we be more explicit if this is the portal client / server / daemon

cli/cli/commands/portal/stop/stop.go
```diff
@@ -0,0 +1,33 @@
...
+	Flags:                    []*flags.FlagConfig{},
+	Args:                     []*args.ArgConfig{},
+	PreValidationAndRunFunc:  nil,
+	RunFunc:                  run,
+	PostValidationAndRunFunc: nil,
+}
+
+func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error {
+	logrus.Infof("Stopping Kurtosis Portal")
```
h4ck3rk3y: "the Kurtosis Portal"

cli/cli/commands/portal/stop/stop.go
```diff
@@ -0,0 +1,33 @@
...
+	portalManager := portal_manager.NewPortalManager()
+	if err := portalManager.StopExisting(ctx); err != nil {
+		return stacktrace.Propagate(err, "Error stopping portal")
```
h4ck3rk3y: Will this put "paths" on the terminal. 

I don't think we should block on PKs work. I have asked him what the expectation from Devs is and if he can make it transparent.

cli/cli/commands/root.go
```diff
@@ -111,6 +112,7 @@ func init() {
 	RootCmd.AddCommand(gateway.GatewayCmd)
 	RootCmd.AddCommand(run.StarlarkRunCmd.MustGetCobraCommand())
 	RootCmd.AddCommand(service.ServiceCmd)
+	RootCmd.AddCommand(portal.PortalCmd)
```
h4ck3rk3y: Alphabetical order

cli/cli/commands/portal/status/status.go
h4ck3rk3y: I understand not mentioning if we want to be transparent about it.

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	allReleaseAssets, _, err := ghClient.Repositories.ListReleaseAssets(ctx, kurtosisTechGithubOrg, kurtosisPortalGithubRepoName, latestRelease.GetID(), opts)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "Unable to get list of assets from latest version")
+	}
+	detectedArchitecture := fmt.Sprintf("%s_%s", runtime.GOOS, runtime.GOARCH)
```
h4ck3rk3y: neat :-) 

nit: perhaps move the "_" to a constant

h4ck3rk3y: s/detectedOSArch/g

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	assetFileExpectedSuffix := fmt.Sprintf("%s%s", detectedArchitecture, githubAssetExtension)
+	var releaseAssetToUse *github.ReleaseAsset
+	for _, releaseAsset := range allReleaseAssets {
+		if strings.HasSuffix(releaseAsset.GetName(), assetFileExpectedSuffix) {
+			releaseAssetToUse = releaseAsset
+		}
+	}
+	if releaseAssetToUse == nil {
+		return nil, stacktrace.NewError("Unable to find Kurtosis Portal binary matching the current architecture. Detected architecture was: '%s'", detectedArchitecture)
```
h4ck3rk3y: and operating system

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	logrus.Debugf("Kurtosis Portal binary found: '%s' with ID '%d'", releaseAssetToUse.GetName(), releaseAssetToUse.GetID())
+
+	// Download the content of the asset. It is expected to be a .tar.gz file containing the Kurtosis Portal binary
+	artifactContent, _, err := ghClient.Repositories.DownloadReleaseAsset(ctx, kurtosisTechGithubOrg, kurtosisPortalGithubRepoName, releaseAssetToUse.GetID(), http.DefaultClient)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "Unable to download Kurtosis portal binary. Name was: '%s', ID was: '%d'", releaseAssetToUse.GetName(), releaseAssetToUse.GetID())
```
h4ck3rk3y: s/portal/Portal/g

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return artifactContent, nil
+}
+
+func extractBinaryToFile(assetContent io.ReadCloser, assetVersion string, destFilePath string, destVersionFilePath string) error {
+	gzipReader, err := gzip.NewReader(assetContent)
```
h4ck3rk3y, line 150: side - i wonder if we can push some of this to a shared lib `kurtosis_compress` maybe; we do some compression/extraction fairly often

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	defer gzipReader.Close()
+
+	assetTarReader := tar.NewReader(gzipReader)
+	onlyFileHeader, err := assetTarReader.Next()
+	if err != nil || onlyFileHeader == nil {
+		return stacktrace.Propagate(err, "Asset archive seems to be empty, this is unexpected")
+	}
+	if onlyFileHeader.Typeflag != tar.TypeReg {
+		return stacktrace.Propagate(err, "Archive seems to be a directory, but expecting a single file")
```
h4ck3rk3y: if the new release is borked; do we error? Someone who has version

12 and version 13 is broken but released

will their workflow break?

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	if _, err = io.Copy(portalBinaryFile, assetTarReader); err != nil {
+		return stacktrace.Propagate(err, "Unable to copy content of Kurtosis Portal binary to executable file")
+	}
+
+	if err = os.WriteFile(destVersionFilePath, []byte(assetVersion), portalVersionFileMode); err != nil {
+		return stacktrace.Propagate(err, "Portal binary file successfully stored but an error occurred persisting its corresponding version.")
+	}
+	logrus.Debugf("Kurtosis Portal binary downloaded to '%s'", destFilePath)
```
h4ck3rk3y, line 182: good stuff!

1. Have we verified that this works with different operating systems signing requirements?
2. Do we know if default upgrade is okay UX wise? I'm fine with it being what we do now and iterating & I don't think this should be blocking. Just getting your thoughts. Like if I open any app it offers me an upgrade and doesn't fail if it doesn't.

cli/cli/helpers/portal_manager/file_helpers.go
```diff
@@ -0,0 +1,24 @@
...
+	if _, err := os.Stat(filePath); err != nil {
```
h4ck3rk3y: ha don't have to check if err is not nil!

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -6,8 +6,21 @@ import (
...
+	"strconv"
+	"syscall"
+)
+
+const (
+	portalPidFileMode       = 0600
+	portalProcessPingSignal = 0
+
+	pidNumberBase    = 10
+	pidNumberBitSize = 64
```
h4ck3rk3y: is this a constant we can pull from somewhere? is this a constant across OS's?

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -33,15 +46,139 @@ func (portalManager *PortalManager) GetClient() portal_generated_api.KurtosisPor
 	return portalManager.portalClientMaybe
 }
 
+// CurrentStatus returns the status of any current Kurtosis Portal process running locally
```
h4ck3rk3y: can we be more explicit in comments that this is the Daemon! Maybe I am wrong; but I got confused which of the two portals run locally. I guess its cause its new to me.

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -33,15 +46,139 @@ func (portalManager *PortalManager) GetClient() portal_generated_api.KurtosisPor
...
+// It returns:
+// - an int corresponding to the PID of the Portal process as stored in its PID file, 0 otherwise
+// - a pointer to the os.Process object corresponding to the currently running Portal process, nil if none is running
+// - a boolean flag corresponding to whether the Portal process is reachable on its ports
+// - an error if something unexpected happened
+func (portalManager *PortalManager) CurrentStatus(ctx context.Context) (int, *os.Process, bool, error) {
+	pid, err := getPidFromPidFile()
```
h4ck3rk3y: s/Pid/PID/

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -33,15 +46,139 @@ func (portalManager *PortalManager) GetClient() portal_generated_api.KurtosisPor
...
+	if err != nil {
+		return 0, nil, false, stacktrace.Propagate(err, "Unexpected error reading PID from PID file")
+	}
+	if pid == 0 {
```
h4ck3rk3y: can we put this in a constant / what does 0 mean

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -33,15 +46,139 @@ func (portalManager *PortalManager) GetClient() portal_generated_api.KurtosisPor
...
+	}
+	return pid, process, true, nil
+}
+
+// StartNew starts a new Portal process. The caller needs to make sure a process is not already running, as ports
+// might conflict
+// It writes the PID of the new process to the Portal PID file
+// It returns the PID of the new Portal process, or an error if something went wrong
+func (portalManager *PortalManager) StartNew(ctx context.Context) (int, error) {
+	portalPidFilePath, err := host_machine_directories.GetPortalPidFilePath()
```
h4ck3rk3y, line 89: s/Pid/PID

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -33,15 +46,139 @@ func (portalManager *PortalManager) GetClient() portal_generated_api.KurtosisPor
...
+	}
+	portalBinaryFile, err := host_machine_directories.GetPortalBinaryFilePath()
+	if err != nil {
+		return 0, stacktrace.Propagate(err, "Unable to get file path to PID file")
+	}
+
+	// Start portal daemon
+	portalLogFilePath, err := host_machine_directories.GetPortalLogFilePath()
+	if err != nil {
+		return 0, stacktrace.Propagate(err, "Unable to get file path to portal log file")
```
h4ck3rk3y: s/porta/Portal/

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -33,15 +46,139 @@ func (portalManager *PortalManager) GetClient() portal_generated_api.KurtosisPor
...
+	kurtosisPortalCmd := exec.Command(portalBinaryFile)
+	kurtosisPortalCmd.Stdout = portalLogFile
+	kurtosisPortalCmd.Stderr = portalLogFile
+	if err = kurtosisPortalCmd.Start(); err != nil {
+		// TODO: maybe print the portal logs here
+		return 0, stacktrace.Propagate(err, "Error trying to start Kurtosis Portal")
+	}
+
+	// Persist new PID to PID file
+	newPid := kurtosisPortalCmd.Process.Pid
```
h4ck3rk3y: s/Pid/PID/

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -33,15 +46,139 @@ func (portalManager *PortalManager) GetClient() portal_generated_api.KurtosisPor
...
+	if err = os.WriteFile(portalPidFilePath, []byte(strconv.Itoa(newPid)), portalPidFileMode); err != nil {
+		return 0, stacktrace.Propagate(err, "Portal was successfully started, but Kurtosis failed at persisting its "+
```
h4ck3rk3y: does this have to be a failure - can we log a warning

is the checking of PID fragile?

Note if you think this comment is valuable, don't let this block you. Address as FLUPs if you think its worth addressing & if it takes too long.

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -33,15 +46,139 @@ func (portalManager *PortalManager) GetClient() portal_generated_api.KurtosisPor
...
+
+// StopExisting stops the existing Portal process, if any, and removes the PID file
+// It returns an error if something went wrong
+func (portalManager *PortalManager) StopExisting(_ context.Context) error {
+	portalPidFilePath, err := host_machine_directories.GetPortalPidFilePath()
+	if err != nil {
+		return stacktrace.Propagate(err, "Unable to get file path to PID file")
+	}
+
+	pid, err := getPidFromPidFile()
```
h4ck3rk3y: s/Pid/PID/g

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -33,15 +46,139 @@ func (portalManager *PortalManager) GetClient() portal_generated_api.KurtosisPor
...
+		return nil
+	}
+
+	process, err := getRunningProcessFromPid(pid)
+	if err != nil {
+		return stacktrace.Propagate(err, "Error getting Portal process from its PID %d", pid)
+	}
+	if process == nil {
+		if portalManager.IsReachable() {
+			logrus.Warnf("Kurtosis Portal process not registered in Kurtosis, but it seems a Portal can be " +
```
h4ck3rk3y: This seems like a duplicate; lets put this in a constant.

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -33,15 +46,139 @@ func (portalManager *PortalManager) GetClient() portal_generated_api.KurtosisPor
...
+		}
+	}
+
+	if err = process.Signal(syscall.SIGINT); err != nil {
+		logrus.Warnf("Error stopping currently running portal on PID: '%d'. It might already be stopped. "+
+			"PID file will be removed. Error was: %s", pid, err.Error())
+	}
+
+	if err = os.Remove(portalPidFilePath); err != nil {
+		return stacktrace.Propagate(err, "Portal was successfully stopped but Kurtosis couldn't remove the "+
```
h4ck3rk3y: perhaps log an error instead of failing? but that doesn't return a non 0 exit code

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -82,3 +219,45 @@ func (portalManager *PortalManager) instantiateClientIfUnset() error {
...
+}
+
+// getRunningProcessFromPid returns the os.Process object corresponding to the Portal process, or nil if the process is not running
+func getRunningProcessFromPid(pid int) (*os.Process, error) {
+	process, err := os.FindProcess(pid)
+	if err != nil {
+		// this should never happen on Unix system, see FindProcess docs
+		return nil, stacktrace.Propagate(err, "Unexpected error getting process attached to PID '%d'", pid)
+	}
+	if err = process.Signal(syscall.Signal(portalProcessPingSignal)); err != nil {
```
h4ck3rk3y, line 258: fine! we catch this in the "IsReachable" thing

h4ck3rk3y (approved):

h4ck3rk3y (approved):

cli/cli/helpers/host_machine_directories/host_machine_directories.go
gbouv: sorry :(

cli/cli/helpers/host_machine_directories/host_machine_directories.go
gbouv: sorry :(

cli/cli/helpers/host_machine_directories/host_machine_directories.go
gbouv: sorry :(

cli/cli/helpers/host_machine_directories/host_machine_directories.go
gbouv: sorry :(

cli/cli/helpers/portal_manager/binary_artifact_getter.go
gbouv: That's very relevant, I'll rework that to not error if an version is present and getting latest version info fails. Good suggestion 👍

cli/cli/helpers/portal_manager/binary_artifact_getter.go
gbouv: Yup, I changed that, we're not failing hard anymore if a version is currently installed 👍

cli/cli/helpers/portal_manager/binary_artifact_getter.go
gbouv: Changed it to `extractAssetTgzToBinaryFileOnDisk`

cli/cli/helpers/portal_manager/binary_artifact_getter.go
gbouv: Replaced all (I think)

cli/cli/helpers/portal_manager/file_helpers.go
gbouv: Oups, yes this was intended to be private, good catch

cli/cli/commands/portal/portal.go
gbouv: Added!

cli/cli/commands/portal/start/start.go
gbouv: Yeah, we should think about that. Technically right now and engine consuming command is a portal consuming, with portal being optional on local context. I agree there's something to dig into here, and we should think about it in the context of Engine potentially not existing at all in the mid term future. Food for thoughts :D

cli/cli/commands/portal/start/start.go
gbouv: Yes I updated DownloadLatestKurtosisPortalBinary to not fail hard if a aversion is currently installed 👍 
And to your second point, I think yes, the portal API should be highly backward compatible, as mentioned in the docs. That might not be true at the beginning, but hopefully we can quickly trend toward something fairly stable. 
The responsibility of the portal is fairly narrow and well defined, right now, so hoping the API doesn't grow crazy

We will see, worst case we can add a flag to request an upgrade. Or we can even encode somewhere in our code the compatibility between Kurtosis and the Portal.

cli/cli/commands/portal/start/start.go
gbouv: Yes to a restart command, already in the pipe :)

cli/cli/commands/portal/status/status.go
gbouv: Makes sense, mentioned "daemon" instead, it's more explicit 👍

cli/cli/commands/portal/status/status.go
gbouv: Interesting insight here. I agree we should be consistent. I'll switch to a normal print here as engine status predates this one and I don't have a strong opinion

gbouv: That being said, I find the mix of log and regular print a bit confusing in the CLI output. Dunno what you think, it seems we don't have clear guidelines on what to use when. This is far outside the scope of this PR, but maybe something we can improve at some point

cli/cli/commands/portal/stop/stop.go
gbouv: For example here, not clear whether it should be a log or a regular print 🤷

cli/cli/commands/portal/stop/stop.go
gbouv: Seems like we should not think too much about it and just continue to use stacktrace.Propagate, which is quite nice

cli/cli/helpers/portal_manager/binary_artifact_getter.go
gbouv: kurtosis STD library maybe? :D

cli/cli/helpers/portal_manager/binary_artifact_getter.go
gbouv: 1. Works on Mac, at least. Will try on linux on the CI. But since in both case we use GO to generate the asset and here retrieve the variable, I'd hope internally it uses GOOS and GOARCH. Anyway, will try out of cautiousness. I am planning on adding integration tests for all this afterwards anyway, which will run on Linux in CI
2. Yeah, I was thinking about this, and same as you I wasn't able to land on what the best option is. So, I chose to default to upgrade automatically because it's less things users have to think about, and also I don't see any major problem with automating the upgrade. If we introduce incompatibilities, we can always add a version compatibiility table somewhere and improve the automatic download logic

cli/cli/helpers/portal_manager/binary_artifact_getter.go
gbouv: Nope it won't break, they will remain on version 12 until version 14 is out and can be installed successfully :)

cli/cli/helpers/portal_manager/file_helpers.go
gbouv: Well, here we have 3 different cases if I'm not mistaken:
- err != nil && IsNotExist == true -> create the file
- err != nil -> return an unexpected error
- err == nil -> file exists already, don't do anything
So, we do have to distinguish between the three anyway. I can refactor this to a `if ... else if ... else`, but can't really simplify the overall logic. I'll do the refactoring as it will have less indentation

cli/cli/helpers/portal_manager/portal_manager.go
gbouv: Hmm, not sure, will take a look

cli/cli/helpers/portal_manager/portal_manager.go
gbouv: That's good input, will add more colour to comments 👍

cli/cli/helpers/portal_manager/portal_manager.go
gbouv: I guess zero here is just GO default values for integer :D But seems right to introduce our own default valuer here, will add

cli/cli/helpers/portal_manager/portal_manager.go
gbouv: So, yeah, it's kind of a big deal actually because it means it won't be able to stop it, users will basically need to kill it themselves :/ I'd like to keep the fail loudly here, it shouldn't happen too often anyway, writing to a file shouldn't be too too risky those days. If it becomes too noisy we can revisit. I was contemplating have the PID be served by the daemon itself, as well as its version number. If/When we have this we will be able to write less things on disk, and this will become more robust 👍

cli/cli/helpers/portal_manager/portal_manager.go
gbouv: Yup that seems right to not fail loudly here 👍

cli/cli/helpers/portal_manager/portal_manager.go
gbouv: found `strconv.IntSize` for the latter, but couldn't find anything on the shelf for `10`

PR #344, opened 2023-03-28
feat: Publish multi-arch image for `files-artifacts-expander`

gbouv:
## Description:
files-artifacts-expander will now be published for Mac M1 as well as AMD64

Technically YES, but not sure they really care about this


gbouv:
Starting with only file expander for now, will add others once we have one successful release with this one

scripts/docker-image-builder.sh
```diff
@@ -0,0 +1,102 @@
...
+echo "Docker file located at '${dockerfile_filepath}' in directory '${dockerfile_dirpath}'"
+
+image_tags_concatenated=""
+for image_tag in ${image_tags}; do
+  echo "Image will be tagged with: '${image_tag}'"
+  image_tags_concatenated=" -t ${image_tag}${image_tags_concatenated} "
+done
+
+# Build Docker image
+kurtosis_docker_builder="kurtosis-docker-builder"
```
h4ck3rk3y, line 61: maybe some formatting can be fixed here and have some new lines between the different blocks!

scripts/docker-image-builder.sh
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env bash
```
h4ck3rk3y, line 1: for this mostly trusting you; i am a "linter" and I just checked that args are quoted!

.circleci/config.yml
```diff
@@ -817,21 +817,19 @@ jobs:
           version: "<< pipeline.parameters.docker-engine-version>>"
       - run: echo "${DOCKER_PASSWORD}" | docker login -u ${DOCKER_USERNAME} --password-stdin
 
-      - run: core/files_artifacts_expander/scripts/build.sh
+      - run: core/files_artifacts_expander/scripts/build.sh true
```
h4ck3rk3y, line 820: cool we are doing this so that we aren't building the image multiple times

h4ck3rk3y (approved):

.circleci/config.yml
gbouv: Exactly 👍

scripts/docker-image-builder.sh
gbouv: ACK, not a bash pro neither, but tested everything I could test I think, will see how it runs on the CI

scripts/docker-image-builder.sh
gbouv: ACK, will add blank lines and some comments

PR #345, opened 2023-03-28
build(deps): Fix replace dependency in files-artifacts-expander

gbouv:
## Description:
Fix failing dependency to contexts-config-store in files-artifacts-expander, missing a replace

NO


victorcolombo (approved):
I believe we should eventually move to https://go.dev/doc/tutorial/workspaces so we don't have to rely on all these replaces on different places

gbouv:
Yes, I agree @victorcolombo, this is quick fix to get back on track but definitely go workspace will help here. I'll see if I have time to set that up

PR #346, opened 2023-03-28
feat: Print the running engine version while running Kurtosis Version

### LINKED ISSUE #232: [FR]:  Kurtosis version prints both the CLI and the engine versions

Engineers go to `kurtosis version` to understand whether they have version mismatches between installations, and in order to truly check this, they'd have to go to both `kurtosis version` and `kurtosis engine status` - which requires a more advanced/in-depth understanding of kurtosis architecture and version management

### Desired behavior

Engineers only have to go to `kurtosis version` to fully understand the kurtosis versions they're running

h4ck3rk3y:
## Description:
We would only print the CLI now we print the running engine version if there is a container running

YES

Closes #232

fix: remove dependence on interactive terminal for configuration


cli/cli/commands/version/version.go
```diff
@@ -20,6 +30,31 @@ func init() {
...
+
+	status, _, maybeEngineVersion, err := engineManager.GetEngineStatus(ctx)
+	if err != nil {
+		// if the engine status can't be fetched we just print the CLI version
+		logrus.Errorf(errorDeterminingEngineVersionLogStr)
+		keyValuePrinter.Print()
+		return nil
+	}
+	if status == engine_manager.EngineStatus_Running {
+		keyValuePrinter.AddPair(runningEngineVersionKey, maybeEngineVersion)
```
gbouv, line 54: Do we want to print something if engine is not running? No strong opinion

cli/cli/kurtosis_config/kurtosis_config_initializer.go
```diff
@@ -12,17 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func initInteractiveConfig() (*resolved_config.KurtosisConfig, error) {
```
gbouv, line 15: This isn't something we're using I guess?

gbouv (approved):
Can we release it after https://github.com/kurtosis-tech/kurtosis/pull/347?

cli/cli/kurtosis_config/kurtosis_config_initializer.go
gbouv: Caught up async. All good to remove

cli/cli/commands/version/version.go
gbouv: Can be tracked as a FLUP if relevant, otherwise also fine to ignore

PR #350, opened 2023-03-28
build: Publish APIC and Engine multi-arch images

gbouv:
## Description:
API container and Engine Docker images are now published for both Mac M1 and Linux AMD64 architecture

YES

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -849,10 +849,11 @@ jobs:
       - run: echo "${DOCKER_PASSWORD}" | docker login -u ${DOCKER_USERNAME} --password-stdin
 
       - run: |
-          core/server/scripts/build.sh
+          core/server/scripts/build.sh true
       - run: |
           set -euo pipefail
           source core/server/scripts/_constants.env
+          dockerfile_filepath='core/server/Dockerfile'
           version_build="$(./scripts/get-docker-tag.sh)"
           version_to_publish="$(cat version.txt)"
           echo "Version that was built: ${version_build}"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -849,21 +849,19 @@ jobs:
       - run: echo "${DOCKER_PASSWORD}" | docker login -u ${DOCKER_USERNAME} --password-stdin
 
       - run: |
-          core/server/scripts/build.sh
```
h4ck3rk3y, line 852: nit: can we put tiny comment on top of what "true" is

core/server/scripts/build.sh
```diff
@@ -12,6 +12,8 @@ source "${script_dirpath}/_constants.env"
 
 BUILD_DIRNAME="build"
 
+DEFAULT_SKIP_DOCKER_IMAGE_BUILDING=false
```
h4ck3rk3y, line 15: super nit: how do you compare this to DEFAULT_DOCKER_BUILD_IMAGE = false; and then you pass in true above

h4ck3rk3y (approved):

.circleci/config.yml
gbouv: Sure, I'll store the `true` in a variable with an explicit name 👍

core/server/scripts/build.sh
gbouv: So, those are different, on docker_image_builder we have `push_to_registry_container`, and on build.sh this is `skip_docker_image_building`. So, slightly different. I could switch `skip_docker_image_building` to `build_docker_image` and turn the default to true, no strong opinion here 🤷

PR #359, opened 2023-03-28
ci: Switch to GH Actions for CloudFlare Preview

victorcolombo:
## Description:

Building docs at each PR is time consuming. This PR migrates away from CloudFlare owned automation to a Kurtosis-managed GitHub Action to build and publish docs.

With that, we can skip build and push if no docs were changed.

NO


.github/workflows/push-docs.yml
```diff
@@ -16,20 +12,28 @@ jobs:
...
       - name: Append changelog
         run: cat CHANGELOG.md >> docs/docs/changelog.md
+        if: github.ref == 'refs/heads/main' || steps.docs-diff.outputs.DOCS_CHANGED != 0
       - name: Yarn Install
         run: yarn --cwd ./docs install
+        if: github.ref == 'refs/heads/main' || steps.docs-diff.outputs.DOCS_CHANGED != 0
       - name: Yarn Build
         run: yarn --cwd ./docs build
-
+        if: github.ref == 'refs/heads/main' || steps.docs-diff.outputs.DOCS_CHANGED != 0
```
gbouv, line 29: Wow, you're the GH action wizard in my head, from now on

gbouv (approved):
That's black magic to me, but overall I think it makes sense. And we can iterate on it if it doesn't work as expected!

PR #361, opened 2023-03-28
fix: added a fix for release failure

Peeeekay:
YES/NO


.circleci/config.yml
```diff
@@ -907,14 +906,14 @@ jobs:
...
     docker:
-      - image: "cimg/go:<< pipeline.parameters.cli-go-version >>"
+      - image: "docker/tilt-releaser"
     steps:
       # Install goreleaser
       - run: |
-          echo 'deb [trusted=yes] https://repo.goreleaser.com/apt/ /' | sudo tee /etc/apt/sources.list.d/goreleaser.list
-          sudo apt update
-          sudo apt install goreleaser
-      - checkout
```
h4ck3rk3y, line 917: are we sure we want to remove the checkout step???

.circleci/config.yml
```diff
@@ -911,9 +911,10 @@ jobs:
     steps:
```
h4ck3rk3y, line 911: the image here is off

h4ck3rk3y (approved):

PR #363, opened 2023-03-28
feat: Automatically map all service ports to local ports post Starlark run and service add

gbouv:
## Description:
When running on a remote context, all services added by `kurtosis run` and `kurtosis service add` will have their ephemeral ports automatically mapped to the user laptop local port

YES - but it's related to remote contexts


cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/kurtosis-tech/stacktrace"
+	"github.com/sirupsen/logrus"
+	"strings"
+)
+
+const (
+	allowNilPortalClientForLocalContext = true
+)
+
+type PortalManager struct {
```
gbouv, line 18: This object can seem unecessary and wrongly named to just map ports here, but I will also use that it to start / stop the portal. That's why it's a "manager"

cli/cli/commands/run/run.go
```diff
@@ -270,35 +272,53 @@ func run(
...
+	if servicesInEnclaveForMetricsError != nil {
+		logrus.Warn("Tried getting number of services in the enclave to log metrics but failed")
+	} else {
+		if err = metricsClient.TrackKurtosisRunFinishedEvent(starlarkScriptOrPackagePath, len(servicesInEnclavePostRun), runStatusForMetrics); err != nil {
+			logrus.Warn("An error occurred tracking kurtosis run finished event")
 		}
+	}
 
+	if errRunningKurtosis != nil {
 		// do not print the go trace in case of APIC failures
```
gbouv: not entirely sure of this one, asked PK here: https://kurtosistech.slack.com/archives/C032ZB39AHH/p1680027261713139

cli/cli/commands/run/run.go
```diff
@@ -270,35 +272,55 @@ func run(
 		return stacktrace.Propagate(errRunningKurtosis, "An error starting the Kurtosis code execution '%v'", starlarkScriptOrPackagePath)
 	}
 
-	if err := metricsClient.TrackKurtosisRun(starlarkScriptOrPackagePath, isRemotePackage, dryRun, isStandAloneScript); err != nil {
+	if err = metricsClient.TrackKurtosisRun(starlarkScriptOrPackagePath, isRemotePackage, dryRun, isStandAloneScript); err != nil {
```
h4ck3rk3y, line 294: nice catch

cli/cli/commands/run/run.go
```diff
@@ -270,35 +272,55 @@ func run(
...
+		runStatusForMetrics = runFailed
+	} else {
+		runStatusForMetrics = runSucceeded
 	}
 
-	servicesInEnclaveForMetrics, servicesInEnclaveForMetricsError := enclaveCtx.GetServices()
+	servicesInEnclavePostRun, servicesInEnclaveForMetricsError := enclaveCtx.GetServices()
 	if servicesInEnclaveForMetricsError != nil {
-		logrus.Error("Tried getting number of services in the enclave to log metrics but failed")
+		logrus.Warn("Tried getting number of services in the enclave to log metrics but failed")
```
h4ck3rk3y, line 309: Agree! I wonder if we could go to an even lower level here

cli/cli/commands/run/run.go
```diff
@@ -270,35 +272,55 @@ func run(
...
 
+	if errRunningKurtosis != nil {
+		// This error thrown by the APIC is not informative right now as it just tells the user to look at errors
+		// in the above log. For this reason we're ignoring it and returning nil. This is exceptional to not clutter
+		// the CLI output. We should still use stacktrace.Propagate for other errors.
+		return nil
+	}
+
+	if servicesInEnclaveForMetricsError != nil {
+		logrus.Warnf("Unable to retrieve the services running inside the enclave so their ports will not be" +
```
h4ck3rk3y, line 324: is this the default behavior we want or do we want to put this port forwarding behind a flag `--forward-ports`; a potentially large enclave could lead to lots of port forwards; maybe this is me future proofing but if we have remote enclaves running we have

1. ports for APIC x2
2. ports for every service

We might quickly run through available ports locally; and maybe this is extra work that the person doesn't care about.

cli/cli/commands/run/run.go
```diff
@@ -270,35 +272,55 @@ func run(
...
+		serviceCtx, err := enclaveCtx.GetServiceContext(string(servicesInEnclaveUuid))
+		if err != nil {
+			return stacktrace.Propagate(err, "Error getting service object for service '%s' with UUID '%s'", serviceInEnclaveName, servicesInEnclaveUuid)
+		}
+		for _, portSpec := range serviceCtx.GetPublicPorts() {
+			portsMapping[portSpec.GetNumber()] = portSpec
+		}
+	}
+	if err = portalManager.MapPorts(ctx, portsMapping); err != nil {
+		// TODO: once we have a manual `kurtosis port map` command, suggest using it here to manually map the failed port
```
h4ck3rk3y, line 350: this would be nice!

cli/cli/commands/run/run.go
```diff
@@ -270,35 +272,55 @@ func run(
...
+		return stacktrace.Propagate(err, "Error mapping the remote enclave ports to local ports.")
+	}
 	return nil
```
h4ck3rk3y, line 358: Can we log a "all ports have been mapped locally"

cli/cli/commands/service/add/add.go
```diff
@@ -298,6 +300,17 @@ func run(
 	publicPorts := serviceCtx.GetPublicPorts()
 	publicIpAddr := serviceCtx.GetMaybePublicIPAddress()
 
+	// Map the service public ports to their local port
+	portalManager := portal_manager.NewPortalManager()
+	portsMapping := map[uint16]*services.PortSpec{}
+	for _, portSpec := range serviceCtx.GetPublicPorts() {
+		portsMapping[portSpec.GetNumber()] = portSpec
+	}
+	if err = portalManager.MapPorts(ctx, portsMapping); err != nil {
```
h4ck3rk3y: lets log the success path too!? I am not sure if port forwarding is expected behavior. Also same concerns about making it optional apply but this is just one service maybe we don't hit limits that quick but still does a user always want mapped ports?

cli/cli/helpers/portal_manager/portal_manager.go
h4ck3rk3y: Sure 👍 

Thoughts on whether it should live here or in the portal?

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -0,0 +1,81 @@
...
+		portalClientMaybe: nil,
+	}
+}
+
+func (portalManager *PortalManager) MapPorts(ctx context.Context, localPortToRemotePortMapping map[uint16]*services.PortSpec) error {
+	if err := portalManager.instantiateClientIfUnset(); err != nil {
+		return stacktrace.Propagate(err, "Unable to set ")
+	}
+	if portalManager.portalClientMaybe == nil {
+		// context is local and portal not present. Port mapping doesn't make sense in a local context anyway, return
```
h4ck3rk3y, line 42: maybe log something here!

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -0,0 +1,81 @@
...
+
+	failedMapping := map[uint16]uint16{}
+	for localPort, remotePort := range localPortToRemotePortMapping {
+		var transportProtocol portal_generated_api.TransportProtocol
+		if remotePort.GetTransportProtocol() != services.TransportProtocol_UDP {
+			transportProtocol = portal_generated_api.TransportProtocol_UDP
+		} else if remotePort.GetTransportProtocol() != services.TransportProtocol_TCP {
+			transportProtocol = portal_generated_api.TransportProtocol_TCP
+		} else {
+			logrus.Debugf("Mapping other than TCP or UDP port is not supported right now. Will skip port '%d' because protocal is '%v'", remotePort.GetNumber(), remotePort.GetTransportProtocol())
```
h4ck3rk3y: should we return `skippedPorts` to user? Maybe if a user intends to have everything mapped and if we move to the `--forward-ports` flags then maybe I'd log a warning here so that the user is aware.

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		forwardPortsArgs := portal_constructors.NewForwardPortArgs(uint32(localPort), uint32(remotePort.GetNumber()), &transportProtocol)
+		_, err := portalManager.portalClientMaybe.ForwardPort(ctx, forwardPortsArgs)
+		if err != nil {
```
h4ck3rk3y: maybe inline this with the ForwardPort line above

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -0,0 +1,81 @@
...
+			failedMapping[localPort] = remotePort.GetNumber()
+		}
+	}
+
+	if len(failedMapping) > 0 {
+		var stringifiedPortMapping []string
+		for localPort, remotePort := range failedMapping {
+			stringifiedPortMapping = append(stringifiedPortMapping, fmt.Sprintf("%d:%d", localPort, remotePort))
```
h4ck3rk3y: super nit: take out delimiter in a constant

h4ck3rk3y (approved):

cli/cli/commands/run/run.go
gbouv: Hopefully that should almost never happen, right? So, I'm good with WARN. Let's see overtime, if we feel it's still too noisy we can lower it again

cli/cli/commands/run/run.go
gbouv: Well, this is technically what we're doing right now (mapping all ports by default). So the "We might quickly run through available ports locally" applies today.
And I think what's messed up right now in SL is that we don't have the possibility to _not_ expose a port. I.e. let's say if I have a server + DB in my package, maybe I want to expose the server ports to the outside so that users can hit the server, but I want the DB port to remain internal to docker as I don't need them exposed. Right now we can't do that, and it's a bit unfortunate. Might be something to bring to product to see what they think about this.

That being said, I like the all or nothing flag, that's a good middle ground here given what we currently have. I'll make it default to expose, such that running a package locally or remotely ends up with the exact same state locally (i.e. ports are reachable). But that's just the starting point, we can flip the default if needed

cli/cli/commands/service/add/add.go
gbouv: Yup, added the same flag to service add and run 👍

cli/cli/helpers/portal_manager/portal_manager.go
gbouv: 👍  I'm now returning the set of successfully mapped ports and the failed ports, and I am more explicit in the logs at the CLI level 👍

cli/cli/helpers/portal_manager/portal_manager.go
gbouv: Yeah, we can move it around in the future if we want to use it in different places. For now this is very client oriented, and therefore I feel okay having it implemented on the client side

PR #369, opened 2023-03-28
fix: clean error paths for users

Peeeekay:
## Description:
We have shortened the error logs shown to the users on the CLI to make it easier for them to debug and troubleshoot issues. If users would like to see the entire stack-trace, users can look at the kurtosis-cli.log file located here: /Users/<username>/Library/Application Support/kurtosis/kurtosis-cli.log to get the detailed error messages.

Or, they would need to run the command again with - --cli-log-level on debug level. For example, kurtosis run .../kurtosis-package --cli-log-level debug - this will display the entire stack-trace on the cli.

YES/NO


cli/cli/commands/feedback/feedback.go
```diff
@@ -299,7 +299,7 @@ func validateUserMsgArg(_ context.Context, _ *flags.ParsedFlags, args *args.Pars
 	}
 
 	if userMsgArg == "" {
-		return stacktrace.Propagate(err, "Error validating the user message argument, it can be an empty string")
+		return stacktrace.NewError("Error validating the user message argument, it can be an empty string")
```
leoporoli: Thanks, man!!! good catch

cli/cli/commands/root.go
```diff
@@ -125,18 +128,24 @@ func globalSetup(cmd *cobra.Command, args []string) error {
...
 	}
 
+	// silently fail if there is an error accessing file logger
+	// downside is that we would not be storing the logs in the file, so could lose
+	// stack-traces and users may need to use --debug flag to get the entire stack-trace
+	fileLogger, err := out.GetFileLogger()
+	if err != nil {
+		logrus.Debugf("Error occurred while getting the file logger %+v", err)
+	} else {
+		// TODO: print the flags that user have entered too
```
leoporoli: I think you can remove the TODO here, right?

cli/cli/commands/root.go
```diff
@@ -125,18 +128,24 @@ func globalSetup(cmd *cobra.Command, args []string) error {
...
+		// log the command user is executing
+		fileLogger.Infof("===== Executing Command: %v %v =====", cmd.Name(), strings.Join(args, " "))
```
leoporoli: I think that you could also include the flags with `cmd.Flags()`

cli/cli/main.go
```diff
@@ -40,8 +46,51 @@ func main() {
 	})
 
 	if err := commands.RootCmd.Execute(); err != nil {
-		// We don't actually need to print the error because Cobra will do it for us
+		maybeCleanedError := getErrorMessageToBeDisplayedOnCli(err)
+		// cobra uses this method underneath as well to print errors
+		// so just used it directly.
+		commands.RootCmd.PrintErrln("Error: ", maybeCleanedError.Error())
```
leoporoli: you can put "Error: " in a constant

leoporoli: it seems that we miss this sentence `Run '___go_build_github_com_kurtosis_tech_kurtosis_cli_cli --help' for usage.` when we handle the error here.

If I execute `kurtosis wrong-command` in main branch I get this error message:
```
Error: unknown command "wrong-command" for "___go_build_github_com_kurtosis_tech_kurtosis_cli_cli"
Run '___go_build_github_com_kurtosis_tech_kurtosis_cli_cli --help' for usage.

```

but if I run this command on this branch, it miss the second line:
```
Error: unknown command "wrong-command" for "___go_build_github_com_kurtosis_tech_kurtosis_cli_cli"
```

cli/cli/main.go
```diff
@@ -40,8 +46,51 @@ func main() {
...
+	errorMessage := errorWithStacktrace.Error()
+	cleanError := removeFilePathFromErrorMessage(errorMessage)
+	return cleanError
+}
+
+// this method removes the file path from the error
+func removeFilePathFromErrorMessage(errorMessage string) error {
+	errorMessageConvertedInList := strings.Split(errorMessage, separator)
+	// safe to assume that the error was not generated using stacktrace package
+	if len(errorMessageConvertedInList) == 1 {
```
leoporoli: you can use a semantic constant for 1

cli/cli/main.go
```diff
@@ -40,8 +46,51 @@ func main() {
...
+		return errors.New(errorMessage)
+	}
+
+	// only the even numbered elements needs to be picked.
+	var cleanErrorList []string
+	for index, line := range errorMessageConvertedInList {
+		if index%2 == 0 {
+			cleanErrorList = append(cleanErrorList, strings.Trim(line, removeTrailingSpecialCharacter))
```
leoporoli:
```suggestion
                        cleanedLine := strings.Trim(line, removeTrailingSpecialCharacter)
			cleanErrorList = append(cleanErrorList, cleanedLine)
```

cli/cli/main.go
```diff
@@ -40,8 +46,51 @@ func main() {
...
+		}
+	}
+
+	return errors.New(strings.Join(cleanErrorList, ""))
```
leoporoli:
```suggestion
         cleanErrorMsg := strings.Join(cleanErrorList, "")
	return errors.New(cleanErrorMsg)
```

cli/cli/main.go
```diff
@@ -40,8 +46,51 @@ func main() {
...
+func getErrorMessageToBeDisplayedOnCli(errorWithStacktrace error) error {
+	// if we are running in the debug mode, just return the error with stack-traces back to the client
+	if logrus.GetLevel() == logrus.DebugLevel {
+		return errorWithStacktrace
+	}
+
+	// silently catch the file logger error and print it in the debug mode
+	// users should not worry about this error
+	// downside is that we may lose stack-traces during file logger failures
+	fileLogger, err := out.GetFileLogger()
```
leoporoli, line 76: now with a second review on this, what do you think about moving this logic inside the `filelogger` so you can add and expose only the `out.Errorln` and `out.Infof` methods (both do not return anything) and the error handling will be centralized an encapsulated inside the `out` package. Because now you are instantiating the object but not injecting it in any place, it's not necessary to respect the interface so far. But this is just a suggestion, I'm ok if you prefer to leave it as it is.

leoporoli (approved):

cli/cli/main.go
Peeeekay: oh good catch; i can add that too here. @leoporoli

cli/cli/main.go
Peeeekay: I just like this better - because it's very verbose i.e `FileLogger` ( is logging only to the file). I can take a look at in next pr. What i agree with you is -- i can create separate `file_logger` package -- and have file_logger.InfoF etc.. This way it's both encapsulated and verbose at the same time? @leoporoli

cli/cli/commands/feedback/feedback.go
Peeeekay: just looked at it again, it should be `it can not be an empty string` right? @leoporoli

cli/cli/main.go
leoporoli: makes sense to have a `file_logger` package, I like it

cli/cli/commands/feedback/feedback.go
leoporoli: yes, that is right!

leoporoli: thanks again!

leeederek:
hey @Peeeekay , can you be sure to label this as "user-facing" next time? Makes it easier to track which changes will go into our release notes (or not). A brief description of this change would be great too next time!

I'm going to use something like:

> In 0.7X.X, we've shortened errors logs returned in the CLI to make it easier to debug and troubleshoot by default. If you'd like to see the full stack trace of an error for a command that you ran, simply pass in the `--debug` flag and run the command again or use the `kurtosis enclave dump` to export logs and view the full stack trace in `kurtosis-cli.log`

Peeeekay:
When this gets released -
We have shortened the error logs shown to the users on the CLI to make it easier for them to debug and troubleshoot issues.  If users would like to see the entire stack-trace, users can look at the `kurtosis-cli.log` file located here: `/Users/<username>/Library/Application Support/kurtosis/kurtosis-cli.log` to get the detailed error messages.


Or, they would need to run the command again with - `--cli-log-level` on `debug` level. For example, `kurtosis run .../kurtosis-package --cli-log-level debug`  - this will display the entire stack-trace on the cli.

cli/cli/commands/feedback/feedback.go
```diff
@@ -291,7 +291,7 @@ func validateUserMsgArg(_ context.Context, _ *flags.ParsedFlags, args *args.Pars
 	}
 
 	if userMsgArg == "" {
-		return stacktrace.Propagate(err, "Error validating the user message argument, it can be an empty string")
+		return stacktrace.NewError("Error validating the user message argument, it can not be an empty string")
```
gbouv, line 294: good catch hehe

cli/cli/commands/root.go
```diff
@@ -125,21 +130,44 @@ func globalSetup(cmd *cobra.Command, args []string) error {
...
+	// stack-traces and users may need to use --debug flag to get the entire stack-trace
+	fileLogger, err := out.GetFileLogger()
+	if err != nil {
+		logrus.Debugf("Error occurred while getting the file logger %+v", err)
+	} else {
+		flagsSetByUsers := getFlagsSetByUsers(cmd.Flags())
+		fileLogger.Infof("===== Executing Command: kurtosis %v %v %v =====",
+			cmd.Name(),
+			strings.Join(args, " "),
+			strings.Join(flagsSetByUsers, " "),
```
gbouv, line 155: nit: Interesting to print the flags after the args. Usually flags are before

cli/cli/main.go
```diff
@@ -40,8 +49,59 @@ func main() {
...
+	if err != nil {
+		logrus.Debugf("Error occurred while getting the file logger %+v", err)
```
gbouv: IMO silenting an error is always bad, I'd bias toward being loud here. At least WARN, if not ERROR. Hopefully that should never happen, so we don't run the risk of being too noisy. But when that happens, at least people know their logs might be missing stuffs.

cli/cli/main.go
```diff
@@ -40,8 +49,59 @@ func main() {
...
+	errorMessageConvertedInList := strings.Split(errorMessage, separator)
+	// safe to assume that the error was not generated using stacktrace package
+	if len(errorMessageConvertedInList) == errorNotCreatedFromStacktrace {
+		return errors.New(errorMessage)
+	}
+
+	// only the even numbered elements needs to be picked.
+	var cleanErrorList []string
+	for index, line := range errorMessageConvertedInList {
+		if index%2 == 0 {
```
gbouv, line 99: Oh wow, this is ... hacky (this, why not), but what worries me the most here is that it is also super fragile. Let's saying tomorrow I implement a command where somewhere a function returned to following error:
```
returns stacktrace.Propagate("I am breaking --- the log cleaning --- logic --- hehehehe")
```
Since I introduced and odd number of what you call separator here, that will mess up your entire logic, and post this line you'll end up printing only the path to files, not the "Caused by"

What I'm trying to say here is that in those situations where we feel the behaviour we're introducing is hacky (sometime hacks are the only way, there's no good alternative), we should try to bias toward what's the safest. What would be safer here would be to have a good regexp to match the lines you want to mute.
If the regexp becomes out of date at some point, then at worse we will start printing those undesired lines again, which is not great, but at least we biased toward not hiding relevant info to the users.

Does that make sense? I feel a bit uncomfortable with this TBH, but let me know if there's something I missed

gbouv (commented):
I fee a bit uncomfortable with the way we mute lines TBH 😬  I mentioned a simple edge case that I think would mess up the "cleaned error". Let me know if I'm reading this right. I propose an alternative that I find safer and maybe not the end of the world to implement, happy to chat more if helpful

cli/cli/main.go
Peeeekay: discussed offline - will do that in next PR ( will add a regex to make this more strict.

cli/cli/commands/root.go
Peeeekay: oh we support flags like before and after ( but will add them in front) good catch.

leeederek:
@Peeeekay can you make sure to update the error output in our quickstart when you merge this PR? In [this section](https://github.com/kurtosis-tech/kurtosis/blob/main/docs/docs/quickstart.md#add-an-api) specifically

PR #374, opened 2023-03-29
feat: best effort autocomplete for service logs after engine restart

### LINKED ISSUE #373: [bug]: The enclave argument of `kurtosis service logs` doesn't have tab-complete enabled

0.70.7

Critical, I am blocked and Kurtosis is unusable to us because of this bug.

Type `kurtosis service logs <TAB>` with an existing enclave. Note that no completion happens. Service completion seems to be fine.

### Relevant logs from `kurtosis enclave dump`

_No response_

### Desired behavior

Enclave completion happens as expected.

h4ck3rk3y:
## Description:
Earlier engine restart would make autocomplete for service logs not work. Now we do a best effort and fetch whatever we can from disk. In the longer term we want identifiers to be persisted on disk so that this doesn't happen.

YES

Closes #373


Peeeekay (approved):
Looks good

PR #376, opened 2023-03-29
refactor!: Change starlark args from struct to dict

### LINKED ISSUE #205: [FR]: Deserialize `args` object as a dictionary, not a struct
### Background and motivation

I'm building the quickstart, and the `args` object is a top-level struct, but nested objects are dictionaries. This is very confusing.

### Current behavior

`args` object is a struct

### Desired behavior

`args` object, and all its nested objects, are dictionaries

Nice to have, this feature would make using Kurtosis more enjoyable.

### LINKED ISSUE #253: [FR]: Make the `args` object default to an empty dictionary when no arguments are passe din

I'm trying to have an optional argument for my script. If the arg is present (a files artifact name) then I'll use it. If it's not present, then I won't mount any files artifacts into the container.

However, it's very annoying to check the existence of a property thanks to #205 , and in the current incarnation (if the user passes in no arguments) then the `args` object itself is `None` so I need to check that first.

### Desired behavior

The `args` object is by default an empty dictionary.

victorcolombo:
## Description:
Despite args being passed to a Starlark scripts as a serialized JSON string, we parsed it to a struct. From now on, instead, the serialized JSON string will be parsed into a dictionary. For example:

```python
def run(plan, args):
    args.value
```

is now

```python
def run(plan, args):
    args["value"]
```

YES

Closes https://github.com/kurtosis-tech/kurtosis/issues/205
Closes https://github.com/kurtosis-tech/kurtosis/issues/253


internal_testsuites/typescript/src/testsuite/startosis_package_test/startosis_package.test.ts
```diff
@@ -73,7 +73,7 @@ test("Test valid Starlark package with input - missing key in params", async ()
         }
 
         expect(runResult.value.interpretationError).not.toBeUndefined()
-        expect(runResult.value.interpretationError?.getErrorMessage()).toContain("Evaluation error: struct has no .greetings attribute")
+        expect(runResult.value.interpretationError?.getErrorMessage()).toContain("Evaluation error: key \"greetings\" not in dict")
```
gbouv, line 76: nit: Is there a way we can print the dict name here? We should have the line number, so should be fine, but in case it's easy

internal_testsuites/starlark/ci_tests/startosis_simple_script.star
```diff
@@ -5,6 +5,7 @@ DATASTORE_PORT_NUMBER = 1323
 DATASTORE_PORT_PROTOCOL = "TCP"
 
 def run(plan, args):
+    plan.assert(args, "==", {})
```
gbouv, line 8: nice

core/server/api_container/server/startosis_engine/package_io/package_io.go
```diff
@@ -105,14 +105,17 @@ func convertValueToStructIfPossible(genericValue starlark.Value) (starlark.Value
...
+			postProcessedValue, interpretationError := convertValueToDictIfPossible(genericDictValue)
+			if interpretationError != nil {
 				// do not wrap the interpretation error here as it's coming from a recursive call.
 				return nil, interpretationError
 			}
-			stringDict[stringKey.GoString()] = postProcessedValue
+			err = dict.SetKey(stringKey, postProcessedValue)
+			if err != nil {
+				return nil, startosis_errors.NewInterpretationError("Unexpected error setting dictionary key '%s' to value '%v'", stringKey, postProcessedValue)
+			}
```
gbouv: Do we need all this logic here?? Can't we just forward the dict that comes back from JSON deserialization straight? What does this code do here?

gbouv (approved):
💥 💥 💥 
Wondering why `convertValueToDictIfPossible` is still needed here. Can't we just kill it with fire?

victorcolombo (commented):
Removed `convertValueToDictIfPossible` logic, added some unit tests for encode/decode

internal_testsuites/typescript/src/testsuite/startosis_package_test/startosis_package.test.ts
victorcolombo: I believe this comes from Starlark, we don't control the error message

core/server/api_container/server/startosis_engine/package_io/package_io.go
```diff
@@ -22,7 +20,7 @@ var (
 // DeserializeArgs deserializes the Kurtosis package args, which should be serialized JSON, into a starlark.Value type.
 //
 // It tries to convert starlark.Dict into starlarkstruct.Struct to allow users to do things like `args.my_param`
-// in their code package in place of `args["my_param"]`. See convertValueToStructIfPossible below for more info.
+// in their code package in place of `args["my_param"]`. See convertValueToDictIfPossible below for more info.
```
gbouv: Update comment as `convertValueToDictIfPossible ` doesn't exist anymore

core/server/api_container/server/startosis_engine/package_io/package_io.go
```diff
@@ -77,49 +68,6 @@ func SerializeOutputObject(thread *starlark.Thread, outputObject starlark.Value)
 	return maybeIndentedOutputObjectStr.GoString(), nil
 }
 
-// convertValueToStructIfPossible tries to convert a starlark.Value type to a starlarkstruct.Struct.
-// It no-ops for most Starlark "simple" types, like string, integer, even iterables.
-// It is expected to successfully convert starlark.Dict to starlarkstruct.Struct. Note however that there are certain
-// cases where this is not possible, like when the dictionary contain non-string keys for example. In this case, this
-// function throws an error. This error should never be hit here because what is being passed comes from serialized
-// JSON, which cannot contain non-string keys.
-func convertValueToStructIfPossible(genericValue starlark.Value) (starlark.Value, *startosis_errors.InterpretationError) {
```
gbouv, line 86: yay

gbouv (approved):
🔥

PR #387, opened 2023-03-29
ci: Notify Slack engineering channel when a release fails

gbouv:
## Description:
This will hopefully send a failure message to our Slack engineering channel when a Circle CI fails one of our release jobs

NO


.circleci/config.yml
```diff
@@ -909,6 +932,12 @@ jobs:
     docker:
       - image: "docker/tilt-releaser"
     steps:
+      - slack/notify:
+          channel: engineering
+          event: fail
+          # You can preview this template and know more about templates here: https://github.com/CircleCI-Public/slack-orb/wiki#templates
+          template: basic_fail_1
```
gbouv, line 940: Wondering if we could just add this block here when CLI is failing,  but better safe than sorry, so adding it everywhere

victorcolombo: I believe that having it everywhere is useful. We never found this but in theory we could have a flake on one of the images when pushing to Docker Hub (could be the case that their servers were unstable). It is a very catastrophic to have APIC push fail. CLI and Engine would work, and users will get an unrecoverable failure when trying to create an enclave

victorcolombo (approved):

.circleci/config.yml
gbouv: I think one image failing will make the CLI fail, Gyani added those requirements: https://github.com/kurtosis-tech/kurtosis/blob/main/.circleci/config.yml#L1131-L1134 which should technically cascade any failure. But yeah, generally agree it's better to have them everywhere. Might be too noisy in case of a failure, but I prefer it being too noisy than failing silently

PR #388, opened 2023-03-30
refactor: added all the starlark apis

Peeeekay:
YES/NO


Peeeekay:
Added all the Starlark commands and made the lsp consistent to our current docs; 
pinned the tilt-releaser version to avoid having issues when something changes upstream;
moved lsp logic to cli/commands/lsp instead of having it as a separate module.

cli/cli/go.mod
```diff
@@ -44,14 +43,13 @@ require (
 	github.com/briandowns/spinner v1.20.0
 	github.com/fatih/color v1.13.0
 	github.com/kurtosis-tech/kurtosis-portal/api/golang v0.0.0-20230328194643-b4dea3081e25
-	github.com/kurtosis-tech/kurtosis/lsp v0.0.0-00010101000000-000000000000
+	github.com/kurtosis-tech/kurtosis/lsp v0.0.0-20230329203818-3d66efd4fce8
```
victorcolombo: We can drop this dependency, right? The package doesn't even exist anymore

victorcolombo (approved):

cli/cli/go.mod
Peeeekay: yeah; suprised that `go mod tidy` didn;t clean this. Let me re-run it.

PR #390, opened 2023-03-30
feat: Noop when switching to current context

gbouv:
## Description:
Not do anything when running `context switch` to the context we're already on

NO


leoporoli (approved):
nice!

PR #393, opened 2023-03-30
feat: Restart engine post cluster set

gbouv:
## Description:
Closes https://github.com/kurtosis-tech/kurtosis-private/issues/1175

NO - as no user should be setting cluster right now


cli/cli/commands/cluster/set/set.go
```diff
@@ -38,15 +44,68 @@ func run(ctx context.Context, flags *flags.ParsedFlags, args *args.ParsedArgs) e
...
+	}
+
+	if err = clusterSettingStore.SetClusterSetting(clusterName); err != nil {
 		return stacktrace.Propagate(err, "Failed to set cluster name to '%v'.", clusterName)
 	}
+	defer func() {
+		if clusterUpdateSuccessful {
+			return
+		}
+		if err = clusterSettingStore.SetClusterSetting(clusterName); err != nil {
```
victorcolombo, line 76: This is nice!

cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -105,16 +107,30 @@ func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error
...
+	// We try to do our best to restart an engine on the same version the current on is on
+	_, _, currentEngineVersion, err := engineManager.GetEngineStatus(ctx)
 	if err != nil {
-		return stacktrace.Propagate(err, "Unable to start a new Kurtosis engine. Context will be rolled back "+
-			"to '%s' and the Kurtosis engine will remain stopped. It can be restarted with 'kurtosis %s %s'",
-			contextPriorToSwitch.GetName(), command_str_consts.EngineCmdStr, command_str_consts.EngineRestartCmdStr)
+		logrus.Warnf("Error getting current engine informationg before restarting it. A default engine will be started")
+	}
+	var engineClientCloseFunc func() error
+	var restartEngineErr error
```
victorcolombo: This is a little bit of repeated logic from the other file. I wonder if we can push this to engine manager or some other abstraction where we centralize the "best effort" versioning of engine we do

victorcolombo (approved):

cli/cli/commands/kurtosis_context/context_switch/switch.go
gbouv: That's not a bad idea, lemme try it!

PR #395, opened 2023-03-30
fix: fix help text and have flags instead of args for files download

### LINKED ISSUE #370: [FR]: Clean up `kurtosis files download` helptext

In [a user feedback session today](https://www.notion.so/kurtosistech/Aron-Kurtosis-8803338359824b56b94510fe5a9891f9?pvs=4) we saw a user struggle mightily getting `kurtosis files download` to work. He kept entering:
```
kurtosis files download enclave ....bunch of different attempts here...
```
and was very confused what a resource identifier is. We eventually determined:

1. Having the way to enter enclave be `--enclave` in `kurtosis run` but be a positional arg here was confusing to him and
2. The `enclave` string in the helptext was confusing to him
3. He wasn't sure what an identifier was (though it is in the helptext, to be fair - think the pressure of a quickstart experience session might have contributed)

### Desired behavior

Long:
```
Download the given files artifact from the given enclave to your machine. The files artifact and enclave are specified by identifier (name, UUID, or shortened UUID). Read more about identifiers here: https://docs.kurtosis.com/reference/resource-identifier
```

also the `enclave` argkey is replaced with `enclave-identifier`

(addressing the consistency of enclave arguments is something else percolating in my brain, for later)

h4ck3rk3y:
## Description:
Makes files download better

YES

Closes #370


cli/cli/commands/files/download/download.go
```diff
@@ -99,9 +99,9 @@ func run(
 	flags *flags.ParsedFlags,
 	args *args.ParsedArgs,
 ) error {
-	enclaveIdentifier, err := args.GetNonGreedyArg(enclaveIdentifierArgKey)
+	enclaveIdentifier, err := flags.GetString(enclaveIdentifierFlagKey)
```
gbouv, line 101: I believe `enclaveIdentifier ` can now be an empty string. What happens in that case?

cli/cli/commands/files/download/download.go
```diff
@@ -53,10 +52,17 @@ const (
...
+	LongDescription:           "Download the given files artifact from the given enclave to your machine. The files artifact and enclave are specified by identifier (name, UUID, or shortened UUID). Read more about identifiers here: https://docs.kurtosis.com/reference/resource-identifier",
 	KurtosisBackendContextKey: kurtosisBackendCtxKey,
 	EngineClientContextKey:    engineClientCtxKey,
 	Flags: []*flags.FlagConfig{
+		{
+			Key: enclaveIdentifierFlagKey,
+			Usage: "The enclave identifier of the enclave in which the script or package will be ran. " +
+				"An enclave with this name will be created if it doesn't exist.",
+			Type:    flags.FlagType_String,
+			Default: autogenerateEnclaveIdentifierKeyword,
```
gbouv: I'm not sure I fully understand the behaviour here... Like, what happens if I don't specify any enclave at all?
Usually, we use the --enclave flag when the enclave name is optional. Like for `kurtosis run`, you can specify an enclave name, if you don't we autogenerate one. But here it feels the enclave is part of the set of identifier that is required to uniquely identify the file to be downloaded. Am I missing something?

Seems like what we need is the concept of mandatory CLI flag, or something like this, dunno

gbouv (commented):
Not tagged as reviewer here but I went through Kevin's ticket this morning and I was wondering what the implementation would look like. Have some interrogations just reading through the change. Maybe I'm missing something

cli/cli/commands/files/download/download.go
h4ck3rk3y: Oh yeah makes sense; I blindly copied from Kurtosis run! Will correct. Enclave identifier should be mandatory. Will correct 👍

gbouv: ACK, that makes more sense. Feel free to re-tag me as reviewer when you push your change, I can approve

cli/cli/commands/files/download/download.go
h4ck3rk3y: This also makes sense; I just copied this from `run` and then tested the happy path. My bad! Fixing. Thanks @gbouv

cli/cli/commands/files/download/download.go
```diff
@@ -53,10 +52,16 @@ const (
...
-	LongDescription:           "Download a files artifact using an identifier(name, uuid, shortened uuid) from an enclave to the host machine",
+	LongDescription:           "Download the given files artifact from the given enclave to your machine. The files artifact and enclave are specified by identifier (name, UUID, or shortened UUID). Read more about identifiers here: https://docs.kurtosis.com/reference/resource-identifier",
 	KurtosisBackendContextKey: kurtosisBackendCtxKey,
 	EngineClientContextKey:    engineClientCtxKey,
 	Flags: []*flags.FlagConfig{
+		{
+			Key:     enclaveIdentifierFlagKey,
+			Usage:   "The enclave from which the file will be downloaded. This is a required flag.",
+			Type:    flags.FlagType_String,
+			Default: defaultEnclaveIdentifierKeyword,
```
gbouv, line 63: We could probably have introduced a "IsRequired" bool here and no default, but can be done later if/when we see more and more required CLI flag.
Wondering is required CLI flag are an anti-pattern here. I've always considered flags as optional as opposed to args which are usually mandatory. We will see

gbouv (approved):

cli/cli/commands/files/download/download.go
h4ck3rk3y: @gbouv agree and I thought of the same thing. But lets see how feedback goes.

PR #396, opened 2023-03-30
fix: Fix confusing warning about port mapping

gbouv:
## Description:
That was happening when the portal daemon was not running

NO

### DIFF
--- a/cli/cli/commands/run/run.go
+++ b/cli/cli/commands/run/run.go
@@ -20,6 +20,7 @@ import (
 	"github.com/kurtosis-tech/kurtosis/cli/cli/helpers/portal_manager"
 	"github.com/kurtosis-tech/kurtosis/cli/cli/user_support_constants"
 	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface"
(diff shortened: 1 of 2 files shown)


cli/cli/commands/run/run.go
```diff
@@ -330,6 +331,21 @@ func run(
 		logrus.Info("Not mapping service ports locally as requested")
 		return nil
 	}
+
+	currentContext, err := store.GetContextsConfigStore().GetCurrentContext()
+	if err != nil {
+		logrus.Warnf("Could not retrieve the current context. Kurtosis will assume context is local and not" +
```
h4ck3rk3y, line 337: nit - can we some how share this error across add/run.

h4ck3rk3y (approved):

cli/cli/commands/run/run.go
gbouv: It's a slightly different phrasing, as here we're talking about all enclaves services whereas in add we're just talking about the service that was added, so I feel okay not wanted to over-refactor here. Let me know if you feel strongly about it and we can brainstorm on potential options

PR #403, opened 2023-03-30
fix: Make GetCluster non fatal if it is unset

gbouv:
## Description:
Trying to be too clever here

No, not yet.


victorcolombo:
Can you add a test on test CLI with a set and unset of cluster?

victorcolombo (approved):

gbouv:
Will do in a FLUP @victorcolombo to get that in asap

leeederek:
Treating this as non-user-facing for now.

PR #413, opened 2023-03-30
feat: clean the error for starlark output as well

Peeeekay:
## Description:
This PR will keep the errors consistent for Starlark execution phase. It will not show the stack-trace path unless users either pass in `cli-log-level debug` or they can see the logs with stack-traces in `kurtosis-cli.log` file. This PR just does the same for starlark as https://github.com/kurtosis-tech/kurtosis/pull/369 did for cli. 

YES

### DIFF
--- a/cli/cli/command_str_consts/command_str_consts.go
+++ b/cli/cli/command_str_consts/command_str_consts.go
@@ -1,6 +1,7 @@
 package command_str_consts
 
 import (
+	"errors"
 	"os"
 	"path"
 )
@@ -61,5 +62,7 @@ const (
 	PathCmdStr              = "path"
 	VersionCmdStr           = "version"
 	GatewayCmdStr           = "gateway"
-	LspCmdStr               = "lsp"
 )
+
+// TODO: added constant error message here, can we move to another file later.
+var ErrorMessageDueToStarlarkFailure = errors.New("Kurtosis execution threw an error. See output above for more details")
--- a/cli/cli/main.go
+++ b/cli/cli/main.go
@@ -7,8 +7,10 @@ package main
 
 import (
 	"errors"
+	"github.com/kurtosis-tech/kurtosis/cli/cli/command_str_consts"
 	"github.com/kurtosis-tech/kurtosis/cli/cli/commands"
 	"github.com/kurtosis-tech/kurtosis/cli/cli/out"
+	"github.com/kurtosis-tech/stacktrace"
 	"github.com/sirupsen/logrus"
 	"os"
 	"strings"
@@ -21,11 +23,8 @@ const (
 	forceColors   = true
 	fullTimestamp = true
 
-	separator                      = "---"
-	removeTrailingSpecialCharacter = "\n"
-	errorNotCreatedFromStacktrace  = 1
-	errorPrefix                    = "Error: "
-	commandNotFound                = "unknown command"
+	errorPrefix     = "Error: "
+	commandNotFound = "unknown command"
 )
 
 func main() {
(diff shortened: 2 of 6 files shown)


cli/cli/command_str_consts/command_str_consts.go
```diff
@@ -57,5 +57,7 @@ const (
 	PathCmdStr              = "path"
 	VersionCmdStr           = "version"
 	GatewayCmdStr           = "gateway"
-	LspCmdStr               = "lsp"
 )
+
+// TODO: added constant error message here, can we move to another file later.
+const ErrorMessageDueToStarlarkFailure = "Kurtosis execution threw an error. See output above for more details"
```
leoporoli: you could use an error type here 
`var ErrorMessageDueToStarlarkFailure   = errors.New("Kurtosis execution threw an error. See output above for more details")`

cli/cli/main.go
```diff
@@ -49,11 +46,17 @@ func main() {
...
+		maybeCleanedError := out.GetErrorMessageToBeDisplayedOnCli(err)
 		errorMessageFromCli := maybeCleanedError.Error()
-		// cobra uses this method underneath as well to print errors
-		// so just used it directly.
-		commands.RootCmd.PrintErrln(errorPrefix, errorMessageFromCli)
+
+		// This error thrown by the APIC is not informative right now as it just tells the user to look at errors
+		// in the above log. For this reason we're ignoring it and returning nil. This is exceptional to not clutter
+		// the CLI output. We should still use stacktrace.Propagate for other errors.
+		if !strings.Contains(errorMessageFromCli, command_str_consts.ErrorMessageDueToStarlarkFailure) {
```
leoporoli: you could use `errors.Is(errorMessageFromCli, ErrorMessageDueToStarlarkFailure)` here instead, if `ErrorMessageDueToStarlarkFailure` is an error object.

cli/cli/out/post_process_error.go
```diff
@@ -0,0 +1,59 @@
...
+	"strings"
+)
+
+const (
+	lineWithStacktraceRegex       = "^--- at\\s*(.*?):([\\d]*)\\s*\\((.*?)\\)\\s*---$"
+	separator                     = "\n"
+	errorNotCreatedFromStacktrace = 1
+)
+
+var lineWithStacktrace = regexp.MustCompile(lineWithStacktraceRegex)
```
leoporoli, line 16: Nice that you compiled the regex first

leoporoli (approved):

cli/cli/main.go
Peeeekay: Let me try this - because we are using propagate wouldn't this be false - but let me try

leoporoli: oh! yes that is right, you could use `stacktrace.RootCause()` to get the root cause error, before you will have to cast the error, but yes probably is too much, let it as it is now.

Peeeekay: hehe already did it; no problem. @leoporoli

cli/cli/out/post_process_error_test.go
```diff
@@ -9,12 +9,12 @@ import (
 func TestRemoveFilePathFromErrorMessage(t *testing.T) {
 	stacktraceErr := createDummyStackTraceWithNonEmptyMsg()
 	errorClean := removeFilePathFromErrorMessage(stacktraceErr.Error())
-	expectedValue := "this is propagated error\n Caused by: Error: this is base error\n "
+	expectedValue := "this is propagated error\nCaused by: Error: this is base error"
```
gbouv, line 12: I think having a `\n` at the end was nice. If we ever print something after the error, it will be on the same line as the last stacktrace line, and will be very difficult to parse for a human. Ca we re-add this?

gbouv: AND, I think the blank space before `Caused by ` was nice as well as it provided some kind of indentation. I understand why it got away, would still be nice if we could get it back

gbouv (commented):
two nits, but otherwise I feel much better about this change than the one before, thanks for updating it :)

PR #414, opened 2023-03-30
fix: colourized the cli output and display starlark messages to the log

Peeeekay:
## Description:

This basically adds the colours the error message that we display on the cli. Like starlark, error message are red in colour

YES


cli/cli/out/cli_writers.go
```diff
@@ -40,13 +40,25 @@ func GetErr() io.Writer {
...
 	if _, printErr := fmt.Fprintln(std.err, msg); printErr != nil {
 		logrus.Errorf("Error printing message to StdErr. Message was:\n%s\nError was:\n%v", msg, printErr.Error())
 	}
 }
+
+func printLogsToFile(msg string) {
+	fileLogger, err := GetFileLogger()
+	if err != nil {
+		logrus.StandardLogger().Warnf("Error using the file logger as it failed with following error: %v. "+
+			"This is a bug where logs are not being collected properly, please file an issue using kurtosis feedback.", err)
```
h4ck3rk3y: s/kurtoiss/Kurtosis/g

cli/cli/out/cli_writers.go
```diff
@@ -40,13 +40,25 @@ func GetErr() io.Writer {
...
+	} else {
+		fileLogger.Println(msg)
+	}
```
gbouv, line 69: nit: Can we take that out of the if else?

cli/cli/out/cli_writers.go
```diff
@@ -40,13 +40,25 @@ func GetErr() io.Writer {
...
+	printLogsToFile(msg)
 	if _, printErr := fmt.Fprintln(std.err, msg); printErr != nil {
 		logrus.Errorf("Error printing message to StdErr. Message was:\n%s\nError was:\n%v", msg, printErr.Error())
 	}
 }
+
+func printLogsToFile(msg string) {
+	fileLogger, err := GetFileLogger()
+	if err != nil {
+		logrus.StandardLogger().Warnf("Error using the file logger as it failed with following error: %v. "+
```
gbouv, line 65: Do we need `.StandardLogger()` here? We usually do `logrus.Warnf`, does this not default to StandardLogger? If yes, can we stick to usual practice here?

cli/cli/out/cli_writers.go
```diff
@@ -40,13 +40,25 @@ func GetErr() io.Writer {
...
 
 func PrintOutLn(msg string) {
+	printLogsToFile(msg)
 	if _, printErr := fmt.Fprintln(std.out, msg); printErr != nil {
 		logrus.Errorf("Error printing message to StdOut. Message was:\n%s\nError was:\n%v", msg, printErr.Error())
 	}
 }
 
 func PrintErrLn(msg string) {
+	printLogsToFile(msg)
```
gbouv: Out of cautiousness, can we swap the 2 calls here and write to file AFTER we print to terminal? Shouldn't make a difference technically, but 🤷

cli/cli/out/cli_writers.go
gbouv: ... or quote `kurtosis feedback` to highlights we're talking about the command here

cli/cli/out/cli_writers.go
Peeeekay: Yes we need this logger; because i have added hook to logrus ( which trigger filelogger in the background to add logs to the file). It will potentially create a recursive loop. added a comment here.

cli/cli/out/cli_writers.go
Peeeekay: dont; think so i can; because it might throw null pointer otherwise.

cli/cli/out/cli_writers.go
gbouv: Not sure I fully understand what this hook thingy is, but does it mean that everywhere we do logrus.Infof in the CLI, we have the add this `.StandardLogger()`?

cli/cli/out/cli_writers.go
gbouv: Ah yes sorry I missed that you're not returning in the above clause, sorry

gbouv (approved):

cli/cli/out/cli_writers.go
Peeeekay: we have to use standardLogger only when we are logging fileLogger failure to be more precise. The main rationale was to have no change in current development. I do have a FLUP to hide this failure behind High Order Function so that developers don't need to think about it. We can discuss this over a call, if need mores clarification @gbouv

PR #416, opened 2023-03-31
docs: fix quickstart with some breaking changes

h4ck3rk3y:
## Description:
Had missed service_name from 0.71.0 in one place
Args serialization changed in 0.72.0

YES

### DIFF
--- a/docs/docs/quickstart.md
+++ b/docs/docs/quickstart.md
@@ -820,8 +820,8 @@ def run(plan, args):
     )
 
     # Insert data
-    if "data" in args:
-        insert_data(plan, args["data"])
+    if "actors" in args:
+        insert_data(plan, args["actors"])
 
 def insert_data(plan, data):
     plan.request(
--- a/docs/docs/quickstart.md
+++ b/docs/docs/quickstart.md
@@ -821,7 +821,7 @@ def run(plan, args):
 
     # Insert data
     if "data" in args:
-        insert_data(plan, args)
+        insert_data(plan, args["data"])
 
 def insert_data(plan, data):
     plan.request(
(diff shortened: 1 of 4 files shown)


docs/docs/quickstart.md
```diff
@@ -820,7 +820,7 @@ def run(plan, args):
     )
 
     # Insert data
-    if args != None:
+    if "data" in args:
```
gbouv: Technically we could also check if the dictionary is empty with a `len()` such that we're more future proof here, but I think that's find. If the structure of the json changes, this code will probably break somewhere else anyway

docs/versioned_docs/version-0.72.0/quickstart.md
```diff
@@ -840,7 +840,7 @@ def insert_data(plan, data):
 Now clean and run, only this time with extra args to `kurtosis run`:
 
 ```bash
-kurtosis clean -a && kurtosis run --enclave quickstart . '[{"first_name":"Kevin", "last_name": "Bacon"}, {"first_name":"Steve", "last_name":"Buscemi"}]'
+kurtosis clean -a && kurtosis run --enclave quickstart . '{"data": [{"first_name":"Kevin", "last_name": "Bacon"}, {"first_name":"Steve", "last_name":"Buscemi"}]}'
```
gbouv: Any chance we can change `data` to something more on topic here? With `data`, it feels a bit like a hack, whereas if we had something like `{persons: [kevin, steve, etc.]}` mayyybe that'd be nicer? 
Deferring to you here

gbouv (approved):
One suggestion, but don't block on it if you disagree

docs/versioned_docs/version-0.72.0/quickstart.md
h4ck3rk3y: point i can just call it `actors` I suppose 👍

PR #417, opened 2023-03-31
fix: Indent Caused by in stacktraces

gbouv:
NO

### DIFF
--- a/cli/cli/out/post_process_error.go
+++ b/cli/cli/out/post_process_error.go
@@ -24,11 +24,11 @@ func GetErrorMessageToBeDisplayedOnCli(errorWithStacktrace error) error {
 	// silently catch the file logger error and print it in the debug mode
 	// users should not worry about this error
 	// downside is that we may lose stack-traces during file logger failures
-	fileLogger, err := GetFileLogger()
+	loggerToFile, err := GetFileLogger()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cli/cli/out/post_process_error.go
```diff
@@ -48,9 +48,9 @@ func removeFilePathFromErrorMessage(errorMessage string) error {
 	var cleanErrorList []string
 	for _, line := range errorMessageConvertedInList {
 		// this only cleans spaces for the lines that contains the stack-trace information
-		cleanLine := strings.TrimSpace(line)
-		if !lineWithStacktrace.MatchString(cleanLine) {
-			cleanErrorList = append(cleanErrorList, cleanLine)
+		trimmedLine := strings.TrimSpace(line)
+		if !lineWithStacktrace.MatchString(trimmedLine) {
+			cleanErrorList = append(cleanErrorList, line) //
```
h4ck3rk3y: trusting you on this!

h4ck3rk3y (approved):
I thought we would have to add a `\t` somewhere :P

cli/cli/out/post_process_error.go
gbouv: You should not! You're right I thought Caused By were indented by default but it seems they are not. Added it, good catch!

PR #427, opened 2023-03-31
fix: Fix links that should be absolute

gbouv:
NO

### DIFF
--- a/docs/docs/explanations/architecture.md
+++ b/docs/docs/explanations/architecture.md
@@ -4,7 +4,7 @@ sidebar_label: Architecture
 sidebar_position: 2
 ---
 
-![Kurtosis Architecture](../../static/img/explanations/kurtosis-architecture.png)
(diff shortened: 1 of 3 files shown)


docs/docs/explanations/architecture.md
```diff
@@ -4,7 +4,7 @@ sidebar_label: Architecture
 sidebar_position: 2
 ---
 
-![Kurtosis Architecture](../../static/img/explanations/kurtosis-architecture.png)
+![Kurtosis Architecture](/static/img/explanations/kurtosis-architecture.png)
```
h4ck3rk3y: this doesn't work

use the @site/static syntax

h4ck3rk3y (approved):

docs/docs/explanations/architecture.md
gbouv: Yeah, using `/img/...` seem to work with both docusaurus versioned docs, and. the linter. So I'll try this

Issue #192, opened 2023-03-14
[FR]: all TCP ports for a service should automatically get a wait that checks if they're open


mieubrisse:
### Background and motivation

When building the quickstart, I found it to be quite cumbersome to build the various recipes (particularly HTTP). I realized that I'm already telling Kurtosis the ports, and if Kurtosis can do a simple "don't proceed until all ports are open" check then we'd solve a LOT of user pain.

### Current behavior

User has to manually wire up the waits in a bunch of different ways, and can ONLY do HTTP requests (which require a ton of information to configure).

### Desired behavior

By default, all TCP ports that get defined on a service come with a simple TCP-level "can Kurtosis connect to the port?" wait.

The user can also add extra configuration to define:
- Turning off the wait entirely
- Timeouts
- Initial delay

### How important is this feature or improvement to you?

Painful, the lack of this feature makes using Kurtosis frictionful.

leoporoli:
The first PR got merged, but the feature is not active yet.

PR #459, opened 2023-04-03
chore: added the main structure for the port wait feature

leoporoli:
## Description:
Added the main structure for the port wait feature, it's disabled because the purpose of the PR is just to review the code and the we will create another PR which will enable the feature, I wanted to split the PRs in order to reduce the size.

NO

https://www.notion.so/kurtosistech/Automatic-TCP-Port-Waiting-17f27247ddc14cca930f5cff5627319b
https://github.com/kurtosis-tech/kurtosis/issues/192


leoporoli:
@gbouv

container-engine-lib/lib/backend_interface/objects/port_spec/port_spec.go
```diff
@@ -8,12 +8,18 @@ type PortSpec struct {
 	number              uint16
 	transportProtocol   TransportProtocol
 	applicationProtocol *string
+	wait                *wait
 }
 
 /*
-	This method accepts port number, transportProtocol and application protocol ( which is optional)
+	This method accepts port number, transportProtocol, application protocol ( which is optional), and port wait
```
gbouv: nit: blank space before "which"

container-engine-lib/lib/backend_interface/objects/port_spec/wait.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+//TODO we probably will rename it, it's in the design stage
+type wait struct {
+	enable       bool
+	timeout      time.Duration
+	initialDelay time.Duration
+}
+
+func NewWait(enable bool, timeout time.Duration, initialDelay time.Duration) *wait {
```
gbouv: Do we want to be able to configure the interval here as well? I guess maybe the design work you're doing in parallel will answer this

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1287,6 +1295,25 @@ func (network *DefaultServiceNetwork) startRegisteredService(
 		}
 	}()
 
+	//TODO it's disable now because we can't use the desired default value because the eth2 needs 5 minutes timeout
+	//TODO we will enable it in the next PR where we will introduce wait's configs for users
```
gbouv, line 1299: Those comments are helpful!

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1686,6 +1713,64 @@ func (network *DefaultServiceNetwork) getServiceNameForIdentifierUnlocked(servic
...
+	}
+	defer func() {
+		for _, userServiceLogsReadCloser := range successfulUserServiceLogs {
+			if err := userServiceLogsReadCloser.Close(); err != nil {
+				logrus.Warnf("We tried to close the user service logs read-closer-objects after we're done using it, but doing so threw an error:\n%v", err)
+			}
+		}
+	}()
+
+	if len(erroredUserServiceUuids) > 0 {
```
gbouv: Can we remove one indentation level here and just do `err, found := erroredUserServiceUuids[serviceUUID]`? It's unlikely we will get another entry in the map anyway, and even if we do, what does it mean for this service specifically?

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -56,6 +58,12 @@ const (
 	exactlyOneShortenedUuidMatch = 1
 
 	singleServiceStartupBatch = 1
+
+	waitForPortsOpenRetriesDelayMilliseconds = uint16(100)
```
gbouv: 100 millis?? Is that a sane default? It's like 10 pings per seconds, seems a lot to me, but why not

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1791,3 +1878,102 @@ func renderTemplateToFile(templateAsAString string, templateData interface{}, de
...
+	var (
+		ticker                  = time.NewTicker(timeout)
+		startTime               = time.Now()
+		finishTime              = startTime.Add(timeout)
+		shouldContinueInTheLoop = true
+		retries                 = 0
+		err                     error
+	)
+
+	time.Sleep(portSpec.GetWait().GetInitialDelay())
```
gbouv: Interesting... So you're computing the finishTime BEFORE you do the initial delay. Can it lead to weird situation where, if I do `initialDelay=60*time.Second, timeout=30*time.Second` it actually throws without trying at least once?
I find this way of thinking about timeout VS initialDelay a bit confusing. To me, it's more natural to start accounting for the timeout once the initial delay has expired. 
1. Sleep for the initial delay
2. Compute finishTime from now
3. Start the retries until timeout has been reached or port is open, whichever comes first

this also has the advantage of completely decoupling the initial delay and timeout, having a less risky logic

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1791,3 +1878,102 @@ func renderTemplateToFile(templateAsAString string, templateData interface{}, de
...
+
+	for shouldContinueInTheLoop {
+		select {
+		case <-ticker.C:
```
gbouv: Oh wow, that's an interesting way to use a Ticker here! What about:
```
ticker = time.NewTicker(waitForPortsOpenRetriesDelayMilliseconds * time.Millisecond)
for {
    if time.Now().After(finishTime) {
        // throw a TIMEOUT error
    }
    //run your check and break the loop if it's successful, or if a hard failure is hit
    <-ticker.C // block until the next tick
}
```
I personally find it more in sync with the idea of a Ticker (which ticks at a fixed rate). Here you're using a ticker only to grab the first tick, it's a bit sad.
And, I believe you have one less indentation level here, since you don't have the `select`

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1791,3 +1878,102 @@ func renderTemplateToFile(templateAsAString string, templateData interface{}, de
...
+	portSpec *port_spec.PortSpec,
+) error {
+	// reject early if it's disable
+	if !portSpec.GetWait().GetEnable() {
+		return nil
+	}
+	timeout := portSpec.GetWait().GetTimeout()
+
+	var (
+		ticker                  = time.NewTicker(timeout)
```
gbouv: Don't forget to defer a Stop for the ticker. From the Ticker doc:
```
Stop the ticker to release associated resources.
```

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1791,3 +1878,102 @@ func renderTemplateToFile(templateAsAString string, templateData interface{}, de
...
+		case <-ticker.C:
+			return stacktrace.NewError("Scanning ports has reached the '%v' time out", timeout.String())
+		default:
+			now := time.Now()
+			scanPortTimeout := finishTime.Sub(now)
+			if err = scanPort(ipAddr, portSpec, scanPortTimeout); err == nil {
+				shouldContinueInTheLoop = false
+				break
+			}
+			time.Sleep(time.Duration(waitForPortsOpenRetriesDelayMilliseconds) * time.Millisecond)
```
gbouv: nit: you don't need the `time.Duration` here, something like `time.Sleep(waitForPortsOpenRetriesDelayMilliseconds * time.Millisecond)` should work, I think

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1791,3 +1878,102 @@ func renderTemplateToFile(templateAsAString string, templateData interface{}, de
...
+		time.Since(startTime),
+	)
+	return nil
+}
+
+func scanPort(ipAddr net.IP, portSpec *port_spec.PortSpec, timeout time.Duration) error {
+	portNumberStr := fmt.Sprintf("%v", portSpec.GetNumber())
+	networkAddress := net.JoinHostPort(ipAddr.String(), portNumberStr)
+	networkStr := strings.ToLower(portSpec.GetTransportProtocol().String())
+	conn, err := net.DialTimeout(networkStr, networkAddress, timeout)
```
gbouv, line 1958: Could we by any chance re-use the same connection object for all retries? Or a failed dial is not recoverable? Especially with an interval at 100 millis, it's kind of sad to have to instanciate so many connections :(

gbouv (approved):
Lots of TODOs in there!
Left a few minor comments, feel free to reach out if you have concerns. Overall the structure looks good!

container-engine-lib/lib/backend_interface/objects/port_spec/wait.go
leoporoli: yes, we don't want, we will just set the timeout here.

core/server/api_container/server/service_network/default_service_network.go
leoporoli: we won't get another entry on the map, only errors when something fails getting the logs

core/server/api_container/server/service_network/default_service_network.go
leoporoli: you are probably right, I will increase it to 500 milliseconds, just 2 pings by second

core/server/api_container/server/service_network/default_service_network.go
leoporoli: I removed the initial delay, it's not something that will provide now, v0 will include only timeouts, more here: https://www.notion.so/kurtosistech/Service-Ports-Wait-Design-Doc-25cc5ea9ab6f4d48a2f2917e4f25c5f5?pvs=4#ce1d2fdf6e5a4ca4a56c3a5995341523

core/server/api_container/server/service_network/default_service_network.go
leoporoli: good suggestion! I'll do the change

core/server/api_container/server/service_network/default_service_network.go
leoporoli: good catch!! thanks

core/server/api_container/server/service_network/default_service_network.go
leoporoli: I've removed this part, I following your suggested pattern instead

core/server/api_container/server/service_network/default_service_network.go
leoporoli: we are getting the connection only once, we retry when we receive an error here and when we receive an error `conn == nil`

PR #470, opened 2023-04-05
refactor: save last x commands

Peeeekay:
YES/NO

### DIFF
--- a/cli/cli/main.go
+++ b/cli/cli/main.go
@@ -49,25 +49,35 @@ func main() {
 		CallerPrettyfier:          nil,
 	})
 
-	if err := commands.RootCmd.Execute(); err != nil {
-		if !displayErrorMessageToCli(err) {
-			os.Exit(errorExitCode)
-		}
+	err := commands.RootCmd.Execute()
+	exitCode := executeCommand(err)
+	os.Exit(exitCode)
+}
 
-		maybeCleanedError := out.GetErrorMessageToBeDisplayedOnCli(err)
-		errorMessageFromCli := maybeCleanedError.Error()
+func executeCommand(err error) int {
+	defer out.RemoveLogFiles()
 
-		fullErrorMessage := fmt.Sprintf("%v %v", errorPrefix, errorMessageFromCli)
-		commands.RootCmd.PrintErrln(output_printers.FormatError(fullErrorMessage))
-
-		// if unknown command is entered - display help command
-		if strings.Contains(errorMessageFromCli, commandNotFound) {
-			helpUsageText := fmt.Sprintf("Run '%v --help' for usage.\n", commands.RootCmd.CommandPath())
-			commands.RootCmd.PrintErrf(output_printers.FormatError(helpUsageText))
-		}
-		os.Exit(errorExitCode)
+	if err == nil {
+		return successExitCode
 	}
-	os.Exit(successExitCode)
+
+	if !displayErrorMessageToCli(err) {
+		return errorExitCode
+	}
+
+	maybeCleanedError := out.GetErrorMessageToBeDisplayedOnCli(err)
+	errorMessageFromCli := maybeCleanedError.Error()
+
+	fullErrorMessage := fmt.Sprintf("%v %v", errorPrefix, errorMessageFromCli)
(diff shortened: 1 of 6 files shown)


cli/cli/main.go
```diff
@@ -49,25 +49,35 @@ func main() {
...
-			os.Exit(errorExitCode)
-		}
+	err := commands.RootCmd.Execute()
+	exitCode := executeCommand(err)
+	os.Exit(exitCode)
+}
 
-		maybeCleanedError := out.GetErrorMessageToBeDisplayedOnCli(err)
-		errorMessageFromCli := maybeCleanedError.Error()
+func executeCommand(err error) int {
```
leoporoli: I think that this method should be renamed, right? because it's not executing commands, is more like `handling the execution error`

cli/cli/out/file_logger.go
```diff
@@ -1,53 +1,107 @@
...
+	files, err := os.ReadDir(kurtosisCliLogsRootFolder)
 	if err != nil {
-		return nil, stacktrace.Propagate(err, "Error occurred while getting the file logger")
+		logrus.Warnf("Error occurred while listing the files from the directory at - '%v' with error %+v", kurtosisCliLogsRootFolder, err)
+	}
+
+	numberOfFilesToBeDeleted := len(files) - numberOfLogFilesForCommands
+	logrus.Debugf("Removing old %v log file(s) from path %v", numberOfFilesToBeDeleted, kurtosisCliLogsRootFolder)
+	if len(files) > numberOfLogFilesForCommands {
+		for i := 0; i < numberOfFilesToBeDeleted; i++ {
```
leoporoli, line 54: are files ordered in ascending order?

cli/cli/out/file_logger.go
```diff
@@ -1,53 +1,107 @@
...
+			fileName := files[i].Name()
+			logFilePath := path.Join(kurtosisCliLogsRootFolder, fileName)
+			if strings.HasPrefix(fileName, fileNamePrefix) {
```
leoporoli, line 56: I think that this condition should be analyzed before this iteration in order to filter the files list, otherwise you could end up without deleting any file

cli/cli/out/file_logger.go
```diff
@@ -1,53 +1,107 @@
...
+				logrus.Debugf("Removing log file at %v", logFilePath)
+				err := os.Remove(logFilePath)
```
leoporoli: nit: here you can use the short condition instead `if err := os.Remove(logFilePath); err != nil`

cli/cli/out/file_logger.go
```diff
@@ -1,53 +1,107 @@
...
+		}
+	}
+
+	// should have had this file in a cli folder but did not do it
+	// added this logic to remove kurtosis-cli.log in kurtosis folder
+	// TODO: remove this next week
+	kurtosisRootFolder := path.Dir(kurtosisCliLogsRootFolder)
+	kurtosisCliLogPath := path.Join(kurtosisRootFolder, fmt.Sprintf("%v.log", fileNamePrefix))
+	if _, err := os.Stat(kurtosisCliLogPath); err == nil {
+		err = os.Remove(kurtosisCliLogPath)
```
leoporoli: same here

cli/cli/out/file_logger.go
```diff
@@ -69,6 +123,7 @@ func setupFileLogger() error {
 		CallerPrettyfier:          nil,
 	}
 
+	// creates a files logger which logs data onto the log file
 	fileLogger = logrus.New()
 	fileLogger.SetOutput(logFile)
 	fileLogger.SetLevel(logrus.InfoLevel)
```
leoporoli, line 139: is `logrus.InfoLevel` enough?

leoporoli (approved):

cli/cli/out/file_logger.go
Peeeekay: yes; by default ReadDir returns in ascending order on `name`

cli/cli/out/file_logger.go
Peeeekay: yeah; `SetLevel` accepts the minimum required level .. so this means any level above than info will trigger hook .. so error will be included

PR #480, opened 2023-04-06
fix: wait command times out even if recipe is still executing

### LINKED ISSUE #477: `wait` is waiting way longer than it should
### What's your CLI version?

0.74.0

I've been periodically hitting cases where `wait` takes wayy too long to timeout. We just hit this in a quickstart user experience session (timeout = 5s, but the wait took 31s).

The tgz of `kurtosis dump` is on the feedback chann as "Dump from kurtosis quickstart exercise"

The impact is that users (myself included) are very confused as to what's happening.

### Desired behavior

`wait` times out when it should

victorcolombo:
## Description:

`wait` is clunky right now:
- We are using an assortment of `Sleep`, keeping various times around, that makes the code hard to read and it's not using idiomatic Go (channels).
- the real `timeout` is `max(last_execution_end, original_timeout)`, which is undesirable if you recipe is hanging way past your original `timeout`
- current `interval` mechanism ignores duration of code, so it acts like "waits exactly `Xms` between call ends and the next one starts" instead of wait at least `Xms` between each call start

This PR fixes all of the above.

YES

Closes https://github.com/kurtosis-tech/kurtosis/issues/477


victorcolombo:
I plan to add an E2E test for this @gbouv , but code wise this should be good to review.

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,63 @@ func ExecuteServiceAssertionWithRecipe(
...
+		case <-instantFirstRunChan:
+			lastResult, recipeErr, assertErr = runRecipe(contextWithDeadline, serviceNetwork, runtimeValueStore, serviceName, recipe, valueField, assertion, target)
+			if recipeErr == nil && assertErr == nil {
+				return lastResult, tries, nil
+			}
+		case <-tickChan:
+			lastResult, recipeErr, assertErr = runRecipe(contextWithDeadline, serviceNetwork, runtimeValueStore, serviceName, recipe, valueField, assertion, target)
+			if recipeErr == nil && assertErr == nil {
+				return lastResult, tries, nil
+			}
```
gbouv, line 124: I would like to have a unit test for this bit. Basically extract all this into an `executeWithTimeout` function that takes as an argument the `runRecipe` function. This way you can unit test it passing it a dumb equivalent of `runRecipe` which just sleeps for X seconds. And adjust the parameters to test all possible end case

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,63 @@ func ExecuteServiceAssertionWithRecipe(
...
+	var recipeErr error
 	var assertErr error
 	tries := 0
-	timedOut := false
 	lastResult := map[string]starlark.Comparable{}
-	startTime := time.Now()
-
-	backoffObj := backoff.NewConstantBackOff(interval)
+	deadline := time.Now().Add(timeout)
+	contextWithDeadline, cancelContext := context.WithDeadline(ctx, deadline)
```
gbouv: I don't understand why we need a contextWithDeadline here. We're passing it to runRecipe then to `Execute`, without doing anything else with it, but I don't see any change to the execute function. What is it used for? My assumption was that somehow the Execute would watch the context and eject when it's being cancelled, but looking at the code downstream I don't think it's doing that... So, confused 🤔

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,63 @@ func ExecuteServiceAssertionWithRecipe(
...
+	defer cancelContext()
+	timeoutChan := time.After(timeout)
+	tickChan := time.Tick(interval)
+	instantFirstRunChan := make(chan bool, 2)
+	instantFirstRunChan <- true
```
gbouv: I would think that I'mfairly good at using channels, but what you're doing here is beyond magic... Why do you need a buffered 2 channels if it's to immediately push a bool to it. Also, it seems we're not pushing any other value to this channel, so... it hangs forever?
I think generally speaking this code is lacking good comments to understand what is being achieved

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,63 @@ func ExecuteServiceAssertionWithRecipe(
...
+				return lastResult, tries, stacktrace.NewError("Recipe execution timed-out waiting for the recipe execution to become valid on service '%v'. Tried '%v' times. Last assertion execution error was:\n '$%v'\n", serviceName, tries, assertErr)
+			}
+			return lastResult, tries, stacktrace.NewError("Recipe execution timed-out waiting for recipe execution on service '%v'. Tries '%v'", serviceName, tries)
+		case <-instantFirstRunChan:
+			lastResult, recipeErr, assertErr = runRecipe(contextWithDeadline, serviceNetwork, runtimeValueStore, serviceName, recipe, valueField, assertion, target)
+			if recipeErr == nil && assertErr == nil {
+				return lastResult, tries, nil
+			}
+		case <-tickChan:
+			lastResult, recipeErr, assertErr = runRecipe(contextWithDeadline, serviceNetwork, runtimeValueStore, serviceName, recipe, valueField, assertion, target)
```
gbouv: What's confusing me the most here is that `runRecipe` is a blocking call here... So let's take the extreme case of runRecipe taking 3 hours to run, how will that function exit after timeout is elapsed (say timeout = 10 secs)?

For me, you need the runRecipe to happen asynchronously, so that you can keep an eye on the timeout timer while it's running, and potentially cancel the execution if the timeout is reached while the recipe is running

Something like this: https://go.dev/play/p/fyPiyd2D6t-

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,63 @@ func ExecuteServiceAssertionWithRecipe(
...
 	tries := 0
-	timedOut := false
 	lastResult := map[string]starlark.Comparable{}
-	startTime := time.Now()
-
-	backoffObj := backoff.NewConstantBackOff(interval)
+	deadline := time.Now().Add(timeout)
+	contextWithDeadline, cancelContext := context.WithDeadline(ctx, deadline)
+	defer cancelContext()
+	timeoutChan := time.After(timeout)
```
gbouv: It seems the docs on After advises us to us NewTimer instead, can we switch?

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,63 @@ func ExecuteServiceAssertionWithRecipe(
...
+	tickChan := time.Tick(interval)
```
gbouv: in addition to the linter, can we make sure we defer a Stop those timers/tickers?

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
gbouv: But again having a good unit test coverage would really help here, maybe I'm mis understanding something

gbouv (changes requested):
I would like us not to merge this until we have good UNIT test coverage here. I don't understand anything on what's possibly going on at runtime. If I need to fix a bug in 15 days in this function, as it is it will take me like at least 5 hours to understand this part.

Add comments to guide future Kurtosians, and test all possible codepath with small, well designed unit tests. So that if we change something, we catch any undesired behaviour

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -795,21 +795,15 @@ func (network *DefaultServiceNetwork) HttpRequestService(ctx context.Context, se
...
-			return nil, stacktrace.Propagate(err, "An error occurred on GET HTTP request on '%v'", url)
-		}
-		return response, err
-	} else {
-		return nil, stacktrace.NewError("An error occurred because %v is unsupported for HTTP request", method)
+	req, err := http.NewRequestWithContext(ctx, method, url, strings.NewReader(body))
+	if err != nil {
+		return nil, stacktrace.NewError("An error occurred building HTTP request on service '%v', URL '%v'", service, url)
+	}
+	resp, err := http.DefaultClient.Do(req)
```
gbouv, line 802: nice

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,93 @@ func ExecuteServiceAssertionWithRecipe(
...
-		lastResult, requestErr = recipe.Execute(ctx, serviceNetwork, runtimeValueStore, serviceName)
-		if requestErr != nil {
-			time.Sleep(backoffDuration)
-			continue
+	}()
+	/*
+		By passing 'contextWithDeadline' to recipe execution, we can make sure that when timeout is reached,
+		the underlying request is aborted.
+		'timeoutChan' serves as an exit signal for the loop repeating the recipe execution
+	*/
```
gbouv: nit: we tend to use "normal" comments exerywhere with simple `//` maybe we can do that here just to keep being consistent

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,93 @@ func ExecuteServiceAssertionWithRecipe(
...
-			break
+	/*
+		We would like to kick an execution right away and after that retry every 'interval' seconds,
+		considering time that took the request to complete.
+		So we prepend an element to 'tickChan'
+	*/
+	tickChan := time.NewTicker(interval)
+	executionTickChan := make(chan time.Time, 2)
+	executionTickChan <- time.Now()
+	go func() {
```
gbouv, line 35: nit: Can we extract those inline functions into real functions with a name that explains what they are doing?

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,93 @@ func ExecuteServiceAssertionWithRecipe(
...
+	/*
+		By passing 'contextWithDeadline' to recipe execution, we can make sure that when timeout is reached,
+		the underlying request is aborted.
+		'timeoutChan' serves as an exit signal for the loop repeating the recipe execution
+	*/
+	contextWithDeadline, cancelContext := context.WithTimeout(ctx, timeout)
+	defer cancelContext()
+	timeoutChan := time.After(timeout)
+
+	execFunc := func() (map[string]starlark.Comparable, error) {
```
gbouv, line 46: and same here. This `ExecuteServiceAssertionWithRecipe` function seems to only be for definining those inline function. That seems weird, I'd just use regular functions with parameters and call them or pass them as params to `executeServiceAssertionWithRecipeWithTicker `

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,93 @@ func ExecuteServiceAssertionWithRecipe(
...
-	if timedOut {
-		return lastResult, tries, stacktrace.NewError("Recipe execution timed-out waiting for the assertion to become valid on service '%v'. Waited for '%v'. Last assertion error was: \n%v", serviceName, time.Since(startTime), assertErr)
+		return lastResult, nil
 	}
-	if requestErr != nil {
-		return lastResult, tries, stacktrace.Propagate(requestErr, "Error executing recipe on service '%v'", serviceName)
-	}
-	if assertErr != nil {
-		return lastResult, tries, stacktrace.Propagate(assertErr, "Error asserting recipe on service '%v'", serviceName)
+	assertFunc := func(lastResult map[string]starlark.Comparable) error {
```
gbouv: and same here

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,93 @@ func ExecuteServiceAssertionWithRecipe(
...
+}
+
+/*
+Executes 'execFunc':
+  - If it errors, retry after the next tick from 'executionTickChan'.
+  - If it succeeds, executes result with 'assertFunc':
+    -- If it succeeds, returns.
+    -- If it errors, retry after the next tick from 'executionTickChan'
+
+If a signal is sent to 'timeoutChan', loop will be broken, last value is returned,
```
gbouv: nit: maybe let's rename timeoutChan to interruptChan maybe?

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,93 @@ func ExecuteServiceAssertionWithRecipe(
...
+alongside if the last error (assert or exec)
+*/
+func executeServiceAssertionWithRecipeWithTicker(
+	serviceName service.ServiceName,
+	execFunc func() (map[string]starlark.Comparable, error),
+	assertFunc func(map[string]starlark.Comparable) error,
+	executionTickChan <-chan time.Time,
+	timeoutChan <-chan time.Time,
+) (map[string]starlark.Comparable, int, error) {
```
gbouv, line 99: Since you already have a good documentation for this function, maybe you could add to it what this `int` returned is, otherwise it's quite obscure

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,93 @@ func ExecuteServiceAssertionWithRecipe(
...
+			if recipeErr != nil {
+				return lastResult, tries, stacktrace.NewError("Recipe execution timed-out waiting for the recipe execution to become valid on service '%v'. Tried '%v' times. Last recipe execution error was:\n '$%v'\n", serviceName, tries, recipeErr)
+			}
+			if assertErr != nil {
+				return lastResult, tries, stacktrace.NewError("Recipe execution timed-out waiting for the recipe execution to become valid on service '%v'. Tried '%v' times. Last assertion execution error was:\n '$%v'\n", serviceName, tries, assertErr)
+			}
+			return lastResult, tries, stacktrace.NewError("Recipe execution timed-out but no errors of assert and recipe happened on service '%v'. This is a bug in Kurtosis.", serviceName)
+		case <-executionTickChan:
+			tries += 1
+			lastResult, recipeErr = execFunc()
```
gbouv, line 117: I'm still confused here by the fact we're executing the exec synchronously. I haven't seen the test below, but I think the case of `execFunc` taking 10 sec and the timeout being 5 seconds for example will break the contract, will see

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,93 @@ func ExecuteServiceAssertionWithRecipe(
...
+			if recipeErr != nil {
+				continue
+			}
+			assertErr = assertFunc(lastResult)
```
gbouv, line 121: and _technically_, same for the assert function. Although maybe it's fine to assume that this one will return quickly

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
gbouv: OMG, you rely on the contextWithTimeout to eject this function!! Okay, seems fine, maybe stick a comment here explaining how it's going to eject on timeout, it's not trivial IMO

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/shared_helpers_test.go
```diff
@@ -0,0 +1,84 @@
...
+	assertFunc := func(map[string]starlark.Comparable) error {
+		return nil
+	}
+	executionTickChan <- time.Now()
+	go func() {
+		time.Sleep(10 * time.Millisecond)
+		timeoutChan <- time.Now()
+	}()
+	_, tries, err := executeServiceAssertionWithRecipeWithTicker(emptyServiceName, execFunc, assertFunc, executionTickChan, timeoutChan)
+	assert.NotNil(t, err)
```
gbouv, line 34: Well, it's testing the result of the function, but it's not testing its behaviour here right? Can we check that the execFunc was called X times, same for assert, and that it effectively timedout after 10 milliseconds, i.e. not before
Easy way of doing that would be to increment a variable inside each of those functions and assert their value at the end. Another way would be too use mocks for those functions

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -25,47 +24,93 @@ func ExecuteServiceAssertionWithRecipe(
...
-		if backoffDuration == backoff.Stop || time.Since(startTime) > timeout {
-			timedOut = true
-			break
+	/*
+		We would like to kick an execution right away and after that retry every 'interval' seconds,
+		considering time that took the request to complete.
+		So we prepend an element to 'tickChan'
+	*/
+	tickChan := time.NewTicker(interval)
+	executionTickChan := make(chan time.Time, 2)
```
gbouv, line 33: Why do you need channel of `time.Time` ? You never seem to read those values anyway, would a channel of boolean be enough?

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/shared_helpers_test.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	assertFirstRun := true
+	assertFunc := func(map[string]starlark.Comparable) error {
+		if assertFirstRun {
+			assertFirstRun = false
+			return stacktrace.NewError("Assert Error")
+		}
+		return nil
+	}
+	for i := 0; i < 10; i++ {
```
gbouv: maybe do that in a go routine and push a value every `interval` millis so that it's closer to the actual behavior

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
gbouv: and I still don't understand why you're using buffered channels here. Bufferred channels are much more difficult to reason about, and I don't think they are necessary here. Are they? If yes, then explain in a comment why

gbouv (approved):
Looks much better! I'm still confused by a few things though, and not really a big fan of relying on both the cancel context and the timeout chan to effectively have a timeout, but I think that's fine for now!

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
victorcolombo: Done!

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
victorcolombo: Chatted offline, the buffered channel is needed to unblock sending the first element to the channel (otherwise it blocks)

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
victorcolombo: Unfortunately context is the standard way to do this for `net/http`, there in't really a channel friendly approach to it

victorcolombo: The assumption here is that everything is very quick, expect the request, which is reasonable

PR #498, opened 2023-04-12
chore: removed validator for ReadyCondition.target_value attribute

leoporoli:
## Description:
Removed validator for the `ReadyCondition.target_value` attribute, because it was validating only string, and we don't have validators for comparable yet

NO


h4ck3rk3y (approved):
LGTM. should we check this in the interepret step perhaps?

PR #499, opened 2023-04-12
docs: Add how to seed MySQL database guide

victorcolombo:
YES/NO


docs/docs/guides/how-to-seed-mysql.md
```diff
@@ -0,0 +1,209 @@
...
+Finally, create a `main.star` file with the following contents:
+
+```python
+mysql = import_module("github.com/kurtosis-tech/mysql-package/mysql.star")
+
+SELECT_SQL_EXAMPLE = "SELECT * FROM post"
+
+def run(plan, args):
+    setup_sql = plan.upload_files(
+        src = "github.com/kurtosis-tech/awesome-kurtosis/blog-mysql-seed/setup.sql",
```
leeederek, line 167: @victorcolombo is there a way we can upload_files and read_file from the locally created .sql files rather than from Github? Otherwise this feels like a vulnerability and also defeats the purpose of having a user write the .sql files locally to begin with.

docs/docs/guides/how-to-seed-mysql.md
```diff
@@ -0,0 +1,209 @@
...
+title: How to programmatically seed MySQL databases for integration tests
+sidebar_label: Programmatically seeding MySQL databases
+slug: /how-to-seed-mysql
+toc_max_heading_level: 2
+sidebar_position: 6
+---
+
+Introduction
+------------
+In this guide, you will programmatically seed a MySQL database for integration tests in an ExpressJS application. ExpressJS can be replaced with any containerized microservice framework you wish to use.
```
leeederek, line 11: Reading through the entire How To guide, i don't believe we use ExpressJS at all. Can we take this out?

PR #500, opened 2023-04-12
refactor: Add exec output to error when it returns a unacceptable code

victorcolombo:
## Description:
When exec returns a code that is not acceptable, we do not return the command output as part of the error. That makes it hard to debug root causes of the failing script.

YES

Fixes #1213


adschwartz (approved):

tedim52 (approved):
LGTM

PR #504, opened 2023-04-13
feat: GRPC file streaming library

gbouv:
## Description:
Helper library to be able to send files through GRPC streaming to not be limited by the 4MB limit

NO


grpc-file-transfer/golang/grpc_file_streaming/client_stream.go
```diff
@@ -0,0 +1,73 @@
+package grpc_file_streaming
+
+import (
+	"github.com/kurtosis-tech/stacktrace"
+	"google.golang.org/grpc"
+)
+
+// ClientStream is a wrapper around a GRPC ClientStream object to be able to send and receive payloads bypassing the
+// 4MB limit set by GRPC.
+type ClientStream[DataChunkMessageType interface{}, ServerResponseType interface{}] struct {
```
victorcolombo: I find `any` to be more idiomatic than `interface{}` here

grpc-file-transfer/golang/grpc_file_streaming/client_stream.go
```diff
@@ -0,0 +1,73 @@
...
+		contentNameForLogging,
+		contentToSend,
+		clientStream.grpcStream.SendMsg,
+		grpcMsgConstructor,
+	)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred sending '%s'", contentNameForLogging)
+	}
+
+	// The client needs to close the stream to tell the server nothing more is expected
```
victorcolombo: I'm confused by terminology here. On readme there is a distinction between client/server, sender/receiver. Would the correct phrase here be `The SENDER needs to close`

grpc-file-transfer/golang/grpc_file_streaming/grpc_stream_io.go
```diff
@@ -0,0 +1,120 @@
...
+	hasher := sha1.New()
+	chunkNumber := 0
+	totalChunksNumber := int(math.Ceil(float64(len(payload)) / chunkSize))
+
+	for blockIdx := 0; blockIdx < totalChunksNumber; blockIdx++ {
+		logrus.Debugf("Sending content for %s. Block number %d/%d", payloadNameForLogging, chunkNumber, totalChunksNumber-1)
+
+		var contentChunk []byte
+		payloadStartIdx := blockIdx * chunkSize
+		if blockIdx == totalChunksNumber-1 {
```
victorcolombo: extract `totalChunksNumber-1` to `lastChunkIndex` for legibility?

grpc-file-transfer/golang/grpc_file_streaming/grpc_stream_io.go
```diff
@@ -0,0 +1,120 @@
...
+			// last chunk - copy everything that's left in the payload array
+			contentChunk = make([]byte, len(payload)-payloadStartIdx)
```
victorcolombo: instead of manually inputing the size here, can't you just `len(payload[payloadStartIdx:])`. That way we don't have to change 2 places if we want to change something in the future, and we absolutely don't run the possibility of under/over allocating

grpc-file-transfer/golang/grpc_file_streaming/grpc_stream_io.go
```diff
@@ -0,0 +1,120 @@
...
+	grpcMsgExtractor func(dataChunk *DataChunkMessageType) ([]byte, string, error),
+) ([]byte, error) {
+	var assembledContent []byte
+	var blockIdx int
+	var computedPreviousBlockHash string
+	hasher := sha1.New()
+
+	dataChunk := new(DataChunkMessageType)
+	errorReceivingChunk := readMsgFromStream(dataChunk)
+	for errorReceivingChunk == nil {
```
victorcolombo, line 92: I wonder if this can be combined into the `for` loop: `for err := call(); err == nil; err = call()`

victorcolombo (approved):

grpc-file-transfer/golang/grpc_file_streaming/client_stream.go
gbouv: I think any is just the more modern version of interface{}. I used interface here because grpc uses interface{} everywhere, but yeah I guess it can't hurt to use any. Will cahnge

grpc-file-transfer/golang/grpc_file_streaming/grpc_stream_io.go
gbouv: Oh yes good call. I changed this bit last minute because the linter flagged it, I should have changed the above as well. Sorry for the confusion here 👍

grpc-file-transfer/golang/grpc_file_streaming/grpc_stream_io.go
gbouv: Haha, interesting. I tried it, and dude I find it super confusing. I usually like this pattern, but here it's a bit different because we have to manually allocate the chunk with `dataChunk := new(DataChunkMessageType)` next to the call to `readMsgFromStream`. So if you move this latter call inside the `for ...;...;...` you end up with `dataChunk := new(...)` floating around. even though I agree the current version sliiiightly more verbose, I'll keep it this way to keep our brains sane

grpc-file-transfer/golang/grpc_file_streaming/client_stream.go
gbouv: Heh, no here it's the client. There's 2 things here: the sender/ receiver, and the client/server. And the 2 are orthogonal.
i.e. for an Upload workflow, the client is the sender, the server is the receiver. While for a download workflow, the server is the sender and the client is the receiver.

Here we're in client_stream so we're talking about the client, and we're in SendData, so the client is the sender. In this specific situation, the client (which again is the sender) needs to close the stream to tell the server "Hey bro you got everything now, Tchao".
I can rephrase the comment here, and I'll take a look at the README again to make sure I explain sender/ receiver, and client/server being orthogonal.

PR #507, opened 2023-04-16
fix: Support ExecRecipe in ReadyCondition

laurentluce:
## Description:
Setting the attribute `recipe` to an `ExecRecipe` in `ReadyCondition` throws the following error:

```
> add_service name="etcd" config=ServiceConfig(image="softlang/etcd-alpine:v3.4.14", ports={"client": PortSpec(number=2379, transport_protocol="TCP")}, env_vars={"ALLOW_NONE_AUTHENTICATION": "yes", "ETCD_DATA_DIR": "/etcd_data"}, ready_conditions=ReadyCondition(recipe=ExecRecipe(command=["etcdctl", "get", "test"]), field="code", assertion="==", target_value=0))
There was an error executing Starlark code 
An error occurred executing instruction (number 1) at github.com/kurtosis-tech/etcd-package[30:28]:
  add_service(name="etcd", config=ServiceConfig(image="softlang/etcd-alpine:v3.4.14", ports={"client": PortSpec(number=2379, transport_protocol="TCP")}, env_vars={"ALLOW_NONE_AUTHENTICATION": "yes", "ETCD_DATA_DIR": "/etcd_data"}, ready_conditions=ReadyCondition(recipe=ExecRecipe(command=["etcdctl", "get", "test"]), field="code", assertion="==", target_value=0)))
  Caused by: An error occurred while checking if service 'etcd' is ready
  Caused by: An error occurred getting the recipe value from ready conditions 'ReadyCondition(recipe=ExecRecipe(command=["etcdctl", "get", "test"]), field="code", assertion="==", target_value=0)'
  Caused by: Required attribute 'recipe' could not be found on type 'ReadyCondition'

Error encountered running Starlark code.
```
The root cause is in `ReadyCondition.GetRecipe()` where checking only for `found` returned by `ExtractAttrValue()` is not enough.  `ExtractAttrValue()` returns `found` set to `false` and an error if the `recipe` attribute is found but of the wrong type.

NO

N/A


core/server/api_container/server/startosis_engine/kurtosis_types/service_config/ready_condition.go
```diff
@@ -109,17 +109,16 @@ func (readyCondition *ReadyCondition) GetRecipe() (recipe.Recipe, *startosis_err
...
-			RecipeAttr, ReadyConditionTypeName)
-	}
 	//TODO we should rework the recipe types to inherit a single common type, this will avoid the double parsing here.
 	if interpretationErr != nil {
 		execRecipe, _, interpretationErr := kurtosis_type_constructor.ExtractAttrValue[*recipe.ExecRecipe](readyCondition.KurtosisValueTypeDefault, RecipeAttr)
 		if interpretationErr != nil {
 			return nil, interpretationErr
 		}
 		genericRecipe = execRecipe
+	} else if !found {
```
victorcolombo: I wonder if we could refactor this code a bit so we reduce how nested this is

There are three cases:
1. Parse a HTTP recipe, if `httpInterpretationErr == nil`, we can safely return. (eject early)
2. Parse an Exec recipe, if `execInterpretationErr == nil`, we can safely return. (eject early)
3. If the code wasn't ejected, return an error at the end of the func

victorcolombo (changes requested):
Great find! Thank you for the contribution! 😄 🚀 

I went ahead and left a small nit, but overall looks good!!!

I've also updated your branch to pick up some CI changes that were failing in your fork.

core/server/api_container/server/startosis_engine/kurtosis_types/service_config/ready_condition.go
laurentluce: Thanks @victorcolombo for the review and fixing the CI check!

`ExtractAttrValue()` returns:
- `found = false` and `err = nil` if the attribute does not exist.
- `found = false` and `err` if the attribute exists but is of the wrong type.
- `found = true` and `err = nil` if the attribute exists and is of the right type.

This means that getting `err = nil` means one of two things:
- The attribute does not exist or the attribute exists and is of the right type.

I could do something like this to reduce the nesting:

1. Parse a HTTPRecipe, if `found == false` and `err == nil`: return an error saying that the required attribute does not exist. If `err == nil`: return the HTTPRecipe.
2. Parse an ExecRecipe, if `err == nil`: return the ExecRecipe.
3. Return err

We could also look into updating `ExtracAttrValue()` to return:
- `found = true` and `err` if the attribute exists but is of the wrong type.

and then do the following:

1. Parse a HTTPRecipe, if `found == false`: return an error saying that the required attribute does not exist. If `err == nil`: return the HTTPRecipe.
2. Parse an ExecRecipe, if `err == nil`: return the ExecRecipe.
3. Return err

Looking at the callers of `ExtractAttrValue()`, it seems like most of them would not need to be updated.  Maybe this can be done in a different PR.

laurentluce: I updated `ReadyCondition.GetRecipe()` to use the following logic to reduce nesting.  Let me know if it looks good.

1. Parse a HTTPRecipe, if `found == false` and `err == nil`: return an error saying that the required attribute does not exist. If `err == nil`: return the HTTPRecipe.
2. Parse an ExecRecipe, if `err == nil`: return the ExecRecipe.
3. Return `err`

gbouv: Chiming in here as I was debugging something similar for Derek today. I think the issue was actually in the `ExtractAttrValue` function, it was returning `found = false` even if the attribute was found, but the type was incorrect (https://github.com/kurtosis-tech/kurtosis/pull/514/files). Which I guess could be discussed philosophically, but overall I believe with my quick fix you shouldn't need this extra logic here.

laurentluce: Thanks @gbouv!  I simplified the first condition based on this change.

victorcolombo (approved):
LGTM

victorcolombo:
@laurentluce your change was shipped in Kurtosis `0.75.4` 🚢

Thanks again!

PR #508, opened 2023-04-17
feat: Add extractors to exec recipe

### LINKED ISSUE #503: Exec recipe should support extract
### Background & motivation

Right now, `GetHttpRequestRecipe` and `PostHttpRequestRecipe` have a `extract` parameter that allows running `jq` on request's output body.

There is no impediment that we can't have that in `ExecRecipe`, and we have signal that it is useful (@adschwartz has a bash script that queries and endpoint and outputs a JSON)

### Desired behaviour

`ExecRecipe` contains a `extract` field that runs agains the command's `output`

victorcolombo:
YES

Closes https://github.com/kurtosis-tech/kurtosis/issues/503


core/server/api_container/server/startosis_engine/recipe/exec_recipe.go
```diff
@@ -114,10 +133,16 @@ func (recipe *ExecRecipe) Execute(
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "Failed to execute command '%v' on service '%s'", command, serviceName)
 	}
-	return map[string]starlark.Comparable{
+	resultDict := map[string]starlark.Comparable{
 		execOutputKey:   starlark.String(commandOutput),
 		execExitCodeKey: starlark.MakeInt(int(exitCode)),
-	}, nil
+	}
+	extractDict, err := runExtractors([]byte(fmt.Sprintf("%q", commandOutput)), extractors)
```
gbouv, line 140: Interesting, why are you using the `%q` here? Do you expect the string sent to runExtractors to be wrapped around two `"` even when it's a JSON? Would it make sense to change `runExtractors` to expect a regular string instead?

core/server/api_container/server/startosis_engine/recipe/exec_recipe.go
```diff
@@ -114,10 +133,16 @@ func (recipe *ExecRecipe) Execute(
...
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred while running extractors from exec recipe")
+	}
+	maps.Copy(resultDict, extractDict)
```
gbouv, line 144: Do you need this? Not used it but I'm dubious it actually copy the content of the comparable. It might just copy the address, which makes it a bit pointless IMO. But maybe I'm missing something

core/server/api_container/server/startosis_engine/recipe/exec_recipe.go
```diff
@@ -114,10 +133,16 @@ func (recipe *ExecRecipe) Execute(
...
 	}
-	return map[string]starlark.Comparable{
+	resultDict := map[string]starlark.Comparable{
 		execOutputKey:   starlark.String(commandOutput),
 		execExitCodeKey: starlark.MakeInt(int(exitCode)),
-	}, nil
+	}
+	extractDict, err := runExtractors([]byte(fmt.Sprintf("%q", commandOutput)), extractors)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred while running extractors from exec recipe")
```
gbouv: Is it worth printing the output here? And maybe the extractors>?

gbouv (approved):
Small nits but overall looks good

core/server/api_container/server/startosis_engine/recipe/exec_recipe.go
victorcolombo: We expect a JSON string as a input. So for example if we send `commandOutput = "str"`, we assume that `str` is a valid string which it is not, instead we need to send `"str"` that is a valid JSON, `commandOutput = "\"str\""`.

core/server/api_container/server/startosis_engine/recipe/exec_recipe.go
victorcolombo: Copy effectively moves all key values from `extractDict` to `resultDict`, "merging" the maps.

core/server/api_container/server/startosis_engine/recipe/exec_recipe.go
victorcolombo: Good catch, done

PR #512, opened 2023-04-17
refactor: Upload file streaming endpoint

gbouv:
## Description:
Add an UploadFileV2 endpoint to APIC that uses GRPC streaming. For now the 4MB limits remains, it will be lifted once all endpoints are migrated to streaming.

NO


victorcolombo (approved):
I would like to see an E2E test checked here, using enclave context to upload a >4MB file

gbouv:
Discussed offline. Makes sense to update the one we currently have for upload_file. However, as noted in the description, this PR does not raise the max file size of 4MB so adding the test in this PR doesn't make sense. I'll merge and add the test as a FLUP in the final PR that raises the file size limit.

PR #513, opened 2023-04-17
refactor: Download file streaming endpoint

gbouv:
## Description:
Add a DownloadFileV2 endpoint to APIC that uses GRPC streaming. For now the 4MB limits remains, it will be lifted once all endpoints are migrated to streaming.

NO


core/server/api_container/server/api_container_service.go
```diff
@@ -442,6 +442,42 @@ func (apicService ApiContainerService) DownloadFilesArtifact(ctx context.Context
 	return resp, nil
 }
 
+func (apicService ApiContainerService) DownloadFilesArtifactV2(args *kurtosis_core_rpc_api_bindings.DownloadFilesArtifactArgs, server kurtosis_core_rpc_api_bindings.ApiContainerService_DownloadFilesArtifactV2Server) error {
+	artifactIdentifier := args.Identifier
```
victorcolombo: Any chance we can reutilize the code from `DownloadFilesArtifact` (at least while they coexist) on the logic of get file?

victorcolombo (approved):
Same here, I would like to see an E2E test

core/server/api_container/server/api_container_service.go
gbouv: Heh, yeah, I can even call `apicService.DownloadFilesArtifact` to get the bytes, and then send them to the stream. Will do that!

gbouv:
Same as for #512, will add it as a FLUP when the file size limit is effectively raised, right now the integration tests we already have covers this

PR #515, opened 2023-04-17
refactor: Upload Starlark package streaming

gbouv:
## Description:
RunStarlarkPackage now streams the content of the package to the APIC. For now the 4MB limits remains, it will be lifted once all endpoints are migrated to streaming.

NO


api/golang/core/lib/binding_constructors/binding_constructors.go
```diff
@@ -70,12 +70,14 @@ func NewRunStarlarkScriptArgs(serializedString string, serializedParams string,
...
 
-func NewRunStarlarkPackageArgs(packageId string, compressedPackage []byte, serializedParams string, dryRun bool, parallelism int32) *kurtosis_core_rpc_api_bindings.RunStarlarkPackageArgs {
+func NewRunStarlarkPackageArgs(packageId string, serializedParams string, dryRun bool, parallelism int32) *kurtosis_core_rpc_api_bindings.RunStarlarkPackageArgs {
 	parallelismCopy := new(int32)
 	*parallelismCopy = parallelism
+	clonePackage := false
 	return &kurtosis_core_rpc_api_bindings.RunStarlarkPackageArgs{
 		PackageId:              packageId,
-		StarlarkPackageContent: &kurtosis_core_rpc_api_bindings.RunStarlarkPackageArgs_Local{Local: compressedPackage},
+		ClonePackage:           &clonePackage,
```
victorcolombo, line 79: We are here taking variable address of something that lives on the stack. Is this fine? Or should we make a defensive copy like a did on `parallelismCopy`?

victorcolombo (approved):

api/golang/core/lib/binding_constructors/binding_constructors.go
gbouv: Hmm, it's not on the stack, I instantiate it right above with `clonePackage := true`, so i thiiiiink it fine?

gbouv: Will merge for now, happy to revisit if I'm missing something

PR #519, opened 2023-04-18
feat: validate port ids before execution

h4ck3rk3y:
YES/NO


core/server/api_container/server/startosis_engine/kurtosis_instruction/wait/wait.go
```diff
@@ -221,7 +221,23 @@ func (builtin *WaitCapabilities) Interpret(arguments *builtin_argument.ArgumentV
 }
 
 func (builtin *WaitCapabilities) Validate(_ *builtin_argument.ArgumentValuesSet, validatorEnvironment *startosis_validator.ValidatorEnvironment) *startosis_errors.ValidationError {
-	// TODO(vcolombo): Add validation step here
+	if serviceExists := validatorEnvironment.DoesServiceNameExist(builtin.serviceName); !serviceExists {
+		return startosis_errors.NewValidationError("Tried creating a wait for service '%s' which doesn't exist", builtin.serviceName)
+	}
+
+	httpRequestRecipe, ok := builtin.recipe.(recipe.HttpRequestRecipe)
```
h4ck3rk3y, line 228: @gbouv is there a nicer way to do lines 228 -> 237?

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -862,6 +862,23 @@ func (network *DefaultServiceNetwork) GetServiceNames() map[service.ServiceName]
 	return serviceNames
 }
 
+func (network *DefaultServiceNetwork) GetServiceNameToPrivatePortIdsMap() (map[service.ServiceName][]string, error) {
```
gbouv: Do we need this function? It seems very specific, AND you're calling it only once, AND I don't think that's something we're going to need very often (and if we do we can move it here)
Can we replace it with a call to `GetServiceNames` and then multiple calls to `GetService`, and implement this directly in the validator, close to where it's needed?

core/server/api_container/server/startosis_engine/kurtosis_instruction/remove_service/remove_service.go
```diff
@@ -73,6 +73,7 @@ func (builtin *RemoveServiceCapabilities) Validate(_ *builtin_argument.ArgumentV
 		return startosis_errors.NewValidationError("There was an error validating '%v' as service name '%v' doesn't exist", RemoveServiceBuiltinName, builtin.serviceName)
 	}
 	validatorEnvironment.RemoveServiceName(builtin.serviceName)
+	validatorEnvironment.RemoveServiceFromPrivatePortIDMapping(builtin.serviceName)
```
gbouv, line 76: It's funny how our validator logic works. It's almost a read/write stack. You can push things to it. You can remove things to it, and then we have read-only operation which effectively do "check that at this point in time, this thing exists". It's kind of cool.

core/server/api_container/server/startosis_engine/kurtosis_instruction/wait/wait.go
```diff
@@ -221,7 +221,23 @@ func (builtin *WaitCapabilities) Interpret(arguments *builtin_argument.ArgumentV
...
+	// if the passed recipe isn't http request recipe we can't do much
+	if !ok {
+		return nil
+	}
+	// check if the recipe contains the PortIdAttr to verify it is a http request recipe
```
gbouv: Don't you already know it's a HttpRequestRecipe since you casted it above? Not sure I understand why you need this here

gbouv: Ohhh, I think I see, the HttpRequestRecipe interface is also implemented by the ExecRecipe. This is messed up, we should fix this instead and after that you can remove this extra check

core/server/api_container/server/startosis_engine/startosis_validator/validator_environment.go
```diff
@@ -42,6 +44,32 @@ func (environment *ValidatorEnvironment) DoesServiceNameExist(serviceName servic
 	return ok
 }
 
+func (environment *ValidatorEnvironment) AddPrivatePortIDForService(portID string, serviceName service.ServiceName) {
```
gbouv, line 47: nit: you're using `port_ID_` and then `port_Id_s` here

core/server/api_container/server/startosis_engine/startosis_validator/validator_environment.go
```diff
@@ -42,6 +44,32 @@ func (environment *ValidatorEnvironment) DoesServiceNameExist(serviceName servic
...
+	portIds, found := environment.serviceNameToPrivatePortIds[serviceName]
+	if found {
+		environment.serviceNameToPrivatePortIds[serviceName] = append(portIds, portID)
+		return
+	}
+	environment.serviceNameToPrivatePortIds[serviceName] = []string{portID}
```
gbouv: You don't need to check whether the key already exists in the map here. Check this out: https://go.dev/play/p/sR7gNr6bLEH

internal_testsuites/golang/testsuite/startosis_request_wait_assert_test/starlark_validation_failure_invalid_port_id_request_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func TestStarlark_InvalidPortIdRequest(t *testing.T) {
+	ctx := context.Background()
+	runResult := test_helpers.SetupSimpleEnclaveAndRunScript(t, ctx, requestInvalidPortIDTest, requestInvalidPortIDFailScript)
+
+	require.Nil(t, runResult.InterpretationError, "Unexpected interpretation error")
+	require.NotEmpty(t, runResult.ValidationErrors, "Expected validation errors")
+	require.Len(t, runResult.ValidationErrors, 1)
+	require.Contains(t, runResult.ValidationErrors[0].ErrorMessage, "Request required port ID 'invalid-port-id' to exist on service 'web-server' but it doesn't")
```
gbouv, line 40: Those tests make me sad. We're really testing a tiny feature, it doesn't quite fit in the idea I have of integration tests. We should think of a way to test our validation logic without having to write integration tests. Def not in the PR, more as food for thoughts.

gbouv (commented):
I'd like to avoid updating the default_service_network for this thing, let me know if that makes sense to you too. The rest looks good, a few nits and philosophical thoughts :P

core/server/api_container/server/startosis_engine/kurtosis_instruction/remove_service/remove_service.go
h4ck3rk3y: I think of it as a stack too.

core/server/api_container/server/service_network/default_service_network.go
h4ck3rk3y: I agree, moving it to the validator makes much more sense. Don't think anyone outside of it will be using it any time soon. Moved! I did feel dirty adding it; should have thought some more :)

core/server/api_container/server/startosis_engine/kurtosis_instruction/wait/wait.go
h4ck3rk3y: i added a `RequestType` method to HttpRequestRecipe that Get & Post implement while exec doesn't. Apart from telling between interfaces that method is useless

core/server/api_container/server/startosis_engine/recipe/http_request_recipe.go
```diff
@@ -33,6 +34,8 @@ type HttpRequestRecipe interface {
 	builtin_argument.KurtosisValueType
 
 	Recipe
+
+	RequestType() string
```
gbouv, line 39: Should we add a comment? I definitely see myself going through this code in 2 months, being like, this is not used, let's remove it, and then breaking the validation flow. WDYT?

gbouv (approved):
Perfect! Maybe just stick a comment to `RequestType` otherwise we might get rid of it as "unused", the rest looks good!

core/server/api_container/server/startosis_engine/kurtosis_instruction/wait/wait.go
gbouv: smart

core/server/api_container/server/startosis_engine/recipe/http_request_recipe.go
h4ck3rk3y: haha! I thought of adding one but got lazy. Added it :-)

h4ck3rk3y:
@leeederek for when you write notes we also now validate service name before the execution runs in request/wait

leeederek:
Thank you! Are there docs for this? @h4ck3rk3y

h4ck3rk3y:
@leeederek no docs. no starlark changes here.

PR #521, opened 2023-04-18
ci: Add back docker-user requirement for build and test steps

victorcolombo:
Reverts kurtosis-tech/kurtosis#520

As @gbouv noted, we are currently facing download rate limit from DockerHub (https://docs.docker.com/docker-hub/download-rate-limit/#whats-the-download-rate-limit-on-docker-hub), that's why we need to authenticate to fetch images

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -248,6 +248,7 @@ jobs:
       # Generate Kurtosis Version
       - run: "<<pipeline.parameters.generate-kurtosis-version-script-path>>"
 
+
       - setup_remote_docker:
           version: "<< pipeline.parameters.docker-engine-version>>"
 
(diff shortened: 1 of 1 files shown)


gbouv:
Yup, and for reference that's the thread where we pulled the trigger on upgrading our Docker account to Pro: https://kurtosistech.slack.com/archives/C032ZB39AHH/p1679319239501219.

An alternative we've been discussing is to have our own container registry cache on AWS or somewhere, which would alleviate Docker rate limiting hopefully to a point where we don't need the pro account anymore, and we can remove this again and downgrade our Docker account to a normal one. That being said, the container registry cache stuff might not be trivial, so we should weigh the pros and cons before pulling the trigger

gbouv (approved):

PR #528, opened 2023-04-19
fix: prune non 0 patch versions for docs that aren't current minor version

### LINKED ISSUE #487: [FR] Changing the version of the docs obstructs my view
### Where did you see the issue in our documentation?

https://docs.kurtosis.com

### Issue description

When choosing the version of the docs I wish to see, I have to use the drop-down bar. This is fine originally but with over 10+ older versions, the drop down bar gets quite big and obstructs my view. 

Is there something that can be done to periodically prune old versions (say, only keep the last 5 versions) to keep the drop down bar small, or add a scroll functionality to the drop down bar?

image

h4ck3rk3y:
## Description:
Prunes old docs with minor version not the same as current if they are a non 0 version patch

YES

Closes #487


.github/prune_old_docs.py
```diff
@@ -0,0 +1,57 @@
...
+        current_version = str_to_version(available_versions_json[0])
+        for version_str in available_versions_json:
+            version = str_to_version(version_str)
+            # if the version being processed is the same as the current major and minor version do nothing
+            if version.major == current_version.major and version.minor == current_version.minor:
+                versions_to_keep.append(version_str)
+                continue
+
+            # if the version is being processed is "0" patch then just keep it
+            if version.patch == "0":
```
victorcolombo: Why do we keep patch `0` but not patch `max_patch`

h4ck3rk3y: 0 would be consistent across versions and easier on eyes. but I can do this if you feel strongly about it?

victorcolombo: I would imagine that we want to keep the latest version of each documentation for each `major` and `minor`

h4ck3rk3y: Works for me 👍 Will change the script a bit more :)

h4ck3rk3y: image

victorcolombo (approved):

PR #493, opened 2023-04-11
refactor: Revamp JQ extractor for HTTP recipe

victorcolombo:
## Description:

Right now, the jq extraction code is rough. This PR improves the following:

1. Add unit tests for the code path
2. Add support for more JSON > Starlark type conversion (`Dict`, `List`, etc...)
3. Decouples extraction logic from HTTP (facilitating https://github.com/kurtosis-tech/kurtosis/issues/503)

## Is this change user facing?
NO

PR #531, opened 2023-04-19
fix: Pass Content-Type header to request

victorcolombo:
## Description:

Before https://github.com/kurtosis-tech/kurtosis/pull/493, we had a branch for `GET` and `POST` request, using `http.Get` and `http.Post`, the latter required you explicitly passing a `contentType`.

Now we are using a `http.DefaultClient.Do`, that does not pre-fill any headers for us, and we simply logged it without passing it along. This is currently breaking https://github.com/kurtosis-tech/eth2-package with

`invalid content type, only application/json is supported`

NO

### DIFF
--- a/internal_testsuites/golang/testsuite/startosis_request_wait_assert_test/startosis_complex_request_wait_assert_test.go
+++ b/internal_testsuites/golang/testsuite/startosis_request_wait_assert_test/startosis_complex_request_wait_assert_test.go
@@ -45,7 +45,7 @@ def run(plan):
 		body=response["extract.exploded-slash"],
 		extract = {
 			"my-body": ".body",
-			"my-content-type": ".headers.content-type"
+			"my-content-type": ".headers.[\"content-type\"]"
 		}
 	)
 	plan.wait(recipe=post_recipe, field="code", assertion="==", target_value=200, service_name="web-server")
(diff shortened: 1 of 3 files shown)


internal_testsuites/golang/testsuite/startosis_request_wait_assert_test/startosis_complex_request_wait_assert_test.go
```diff
@@ -41,22 +41,27 @@ def run(plan):
...
 		extract = {
-			"my-body": ".body"
+			"my-body": ".body",
+			"my-content-type": ".headers.content-type"
 		}
 	)
 	plan.wait(recipe=post_recipe, field="code", assertion="==", target_value=200, service_name="web-server")
 	post_response = plan.request(recipe=post_recipe, service_name="web-server")
 	plan.assert(post_response["code"], "==", 200)
+	plan.assert(post_response["extract.my-content-type"], "==", "text/plain")
```
tedim52, line 54: is the content-type `text/plain` by default if you don't set it?

tedim52 (approved):
LGTM

internal_testsuites/golang/testsuite/startosis_request_wait_assert_test/startosis_complex_request_wait_assert_test.go
victorcolombo: No it's not, fixed it!

PR #541, opened 2023-04-20
feat: replace runtime values in output with real values

h4ck3rk3y:
YES


api/golang/core/lib/enclaves/starlark_run_blocking.go
```diff
@@ -54,6 +54,14 @@ func ReadStarlarkRunResponseLineBlocking(starlarkRunResponseLines <-chan *kurtos
 			} else if responseLine.GetError().GetExecutionError() != nil {
 				executionError = responseLine.GetError().GetExecutionError()
 			}
+		} else if responseLine.GetRunFinishedEvent() != nil {
+			runFinishedEvent := responseLine.GetRunFinishedEvent()
+			if runFinishedEvent.GetIsRunSuccessful() {
+				if runFinishedEvent.GetSerializedOutput() != "" {
```
h4ck3rk3y: i am curious how controversial this is for you @gbouv

gbouv: Yeah, not sure TBH. It feels off to me to just dump the output object to the script output which already contains a bunch of other stuffs. Maybe we should add an attribute to StarlarkRunResult and store this separately? At the same time we don't know how this will be used. Maybe this entire stringified scriptOutput is dumb in the first place.
🤷

api/golang/core/lib/enclaves/starlark_run_blocking.go
```diff
@@ -54,6 +54,14 @@ func ReadStarlarkRunResponseLineBlocking(starlarkRunResponseLines <-chan *kurtos
 			} else if responseLine.GetError().GetExecutionError() != nil {
 				executionError = responseLine.GetError().GetExecutionError()
 			}
+		} else if responseLine.GetRunFinishedEvent() != nil {
+			runFinishedEvent := responseLine.GetRunFinishedEvent()
+			if runFinishedEvent.GetIsRunSuccessful() {
```
gbouv: nit: Maybe merge the 2 `if` together here to remove one level of indent

gbouv (approved):

api/golang/core/lib/enclaves/starlark_run_blocking.go
h4ck3rk3y: At the same time differentiating between the RunOutput and the FinalOutput feels off too; to me at least. Like why do you draw the line. 

I suppose this is a decision we can make in the future if people complain about the final "output" appearing in the RunOutput?

gbouv: Yeah 100%. It's not used that much, we're missing concrete datapoints to make a good decision here so, I'm fine with any solution here.

PR #543, opened 2023-04-21
feat: stream warnings to the terminal

Peeeekay:
YES/NO

### DIFF
--- a/core/server/api_container/server/api_container_service.go
+++ b/core/server/api_container/server/api_container_service.go
@@ -673,20 +673,21 @@ func (apicService ApiContainerService) runStarlarkPackageSetup(packageId string,
 func (apicService ApiContainerService) runStarlark(parallelism int, dryRun bool, packageId string, serializedStarlark string, serializedParams string, stream grpc.ServerStream) {
 	responseLineStream := apicService.startosisRunner.Run(stream.Context(), dryRun, parallelism, packageId, serializedStarlark, serializedParams)
 	startosis_warning.Setup(stream)
+	defer func() {
+		startosis_warning.Close()
+	}()
 
 	for {
 		select {
 		case <-stream.Context().Done():
 			// TODO: maybe add the ability to kill the execution
 			logrus.Infof("Stream was closed by client. The script ouput won't be returned anymore but note that the execution won't be interrupted. There's currently no way to stop a Kurtosis script execution.")
-			startosis_warning.Close()
 			return
 		case responseLine, isChanOpen := <-responseLineStream:
 			if !isChanOpen {
 				// Channel closed means that this function returned, so we won't receive any message through the stream anymore
 				// We expect the stream to be closed soon and the above case to exit that function
 				logrus.Info("Startosis script execution returned, no more output to stream.")
-				startosis_warning.Close()
 				return
 			}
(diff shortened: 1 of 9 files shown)


core/server/api_container/server/api_container_service.go
```diff
@@ -671,6 +672,11 @@ func (apicService ApiContainerService) runStarlarkPackageSetup(packageId string,
 
 func (apicService ApiContainerService) runStarlark(parallelism int, dryRun bool, packageId string, serializedStarlark string, serializedParams string, stream grpc.ServerStream) {
 	responseLineStream := apicService.startosisRunner.Run(stream.Context(), dryRun, parallelism, packageId, serializedStarlark, serializedParams)
+	startosis_warning.Setup(stream)
```
Peeeekay, line 675: I tried testing this approach multiple times, including running starlark script multiple times in same enclave and didn't see any issues. Let me know if you know of any other tests I should run @gbouv 

StartosisRunner.Run(...) uses chan to co-ordinate between various go-routines. I tried to move this in that function but it seems complex and to be honest not sure what value we would be gaining. 

I think I would need to add a new warning chan and forward the message from that channel to `starlarkResponseLines` channel, and this creates unnecessary dependency on the latter as it is required for collecting the output from each go-routines, whereas warning streams to me is more of a global module which can be called anytime  on-demand by any go-routine.

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
```diff
@@ -0,0 +1,48 @@
...
+// Close this method is called once the starlark execution is completed
+// This method is called during clean up and ensures that next time
+// when set up is called - warn struct will be initialized with current available stream
+func Close() {
+	once = nil
+}
+
+//TODO: add an abstraction so that this method actually prints to cli only once!
+func PrintOncef(message string, args ...interface{}) {
+	lock.Lock()
```
Peeeekay, line 39: `SendMsg` is unsafe to use concurrently - so that's why added a lock. More on this: https://github.com/grpc/grpc-go/blob/master/Documentation/concurrency.md

Peeeekay:
This gives me an ability to stream warning from starlark to cli on demand. 

In next PRs, I can add formatting so that the warning is `yellow` in colour and also to make sure only one warning is shown irrespective how many times an instruction is called. At the moment; if I am printing a warning in say `add_service` and calling `add_services` 10 times. The warning will appear 10 times ( which is not good)

core/server/api_container/server/api_container_service.go
gbouv: I am not sure I fully understand what's blocking the `starlarkResponseLines` TBH, but I trust you, if it creates a lot of complexity and pain let's use the raw stream then

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
gbouv: Yeah so that's the kind of thing the `starlarkResponseLines` would provide out of the box. Like, if you could plug the `starlarkResponseLines` instead of the raw stream, you wouldn't need to care about concurrency since a Go stream is concurrent by nature. Again I'm not sure what's wrong with using `starlarkResponseLines`, but might be wise to dig a bit more into it. The concurrency story with the manual Lock here is not very reassuring

PR #545, opened 2023-04-21
fix: restrict random network allocation to 10.0.0.0/8 following RFC 4096

h4ck3rk3y:
## Description:
Some services seem to follow RFC 6980; which requires certain flags to be passed to those services to allow non private ip addresses (as defined by the RFC). While working with the weaviate example I learned that docker has no pain composing

```
version: '3.4'
services:
  weaviate:
    image: semitechnologies/weaviate:1.18.3
    ports:
    - 8080:8080
    restart: on-failure:0
    environment:
      QUERY_DEFAULTS_LIMIT: 25
      AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED: 'true'
      PERSISTENCE_DATA_PATH: '/var/lib/weaviate'
      DEFAULT_VECTORIZER_MODULE: 'none'
      CLUSTER_HOSTNAME: 'node1'
```
While with Kurtosis you can't run (even though the default inspect output says RUNNING if you run it again you can see the service STOPPED)

```
WEAVIATE_IMAGE="semitechnologies/weaviate:1.18.3"

WEAVIATE_PORT = 8080
WEAVIATE_PORT_ID = "http"

def run(plan, args):

    plan.add_service(
        name = "weaviate",
        config = ServiceConfig(
            image = WEAVIATE_IMAGE,
            ports = {
                WEAVIATE_PORT_ID: PortSpec(number = WEAVIATE_PORT, transport_protocol = "TCP")
            },
            env_vars = {
                "QUERY_DEFAULTS_LIMIT": str(25),
                "AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED": 'true',
                "PERSISTENCE_DATA_PATH": '/var/lib/weaviate',
                "DEFAULT_VECTORIZER_MODULE": 'none',
                "CLUSTER_HOSTNAME": 'node1'
            },
        )
    )
```

This is because WEAVIATE uses hashicorp/memberlist which uses serf underneath https://github.com/hashicorp/serf/issues/385#issuecomment-208755148 underneath that disallows any non private ip addresses. Similar in Grafana https://github.com/grafana/grafana/issues/62453

NO


container-engine-lib/lib/backend_impls/docker/docker_network_allocator/docker_network_allocator.go
```diff
@@ -28,88 +28,16 @@ const (
...
-// If we don't, Docker will throw an error of "failed to set gateway while updating gateway: route for the gateway X.X.X.X could not be found: network is unreachable"
-// The start & end are BOTH inclusive (else we'd have no way to block off the single 255.255.255.255/32 address)
-// See https://en.wikipedia.org/wiki/IPv4#Special-use_addresses
-var disallowedIpRanges = [][][]byte{
-	{
-		// Current network (only valid as source address)
-		{0, 0, 0, 0}, {0, 255, 255, 255},
-	},
-	{
-		// Used for local communications within a private network
```
h4ck3rk3y, line 48: We don't run into "failed to set gateway" even though we are now using this very network.

Wonder why this was disallowed

internal_testsuites/golang/testsuite/check_that_apic_network_is_private/check_that_apic_network_is_private_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+const (
+	testName              = "apic-is-private"
+	isPartitioningEnabled = false
+	defaultDryRun         = false
+	defaultParallelism    = 4
+	emptyArgs             = "{}"
+
+	// This Script fails if the APIC has a non private IP address range
+	starlarkScript = `
```
victorcolombo: Having this as an E2E test seems like an overhead. I would rather have us unit test this or, even better, having `panic` + comment on the original function to make sure this precondition doesn't change without something bad happening

victorcolombo (approved):

internal_testsuites/golang/testsuite/check_that_apic_network_is_private/check_that_apic_network_is_private_test.go
h4ck3rk3y: I have the unit test and the comment already. have removed the e2e test. :)

PR #552, opened 2023-04-24
ci: limit the number of versions of current minor version of docs kept & historical versions kept

h4ck3rk3y:
## Description:
Enforces limits using

HISTORICAL_VERSIONS_TO_KEEP = 5
MOST_TO_KEEP_OF_LATEST_VERSION = 5

If they're set to -1 then we keep all of the current version.

YES


h4ck3rk3y:
image

With it set to 5,5

.github/prune_old_docs.py
```diff
@@ -11,33 +11,47 @@
 
 Version = namedtuple("Version", "major minor patch")
 
+HISTORICAL_VERSIONS_TO_KEEP = 5
```
h4ck3rk3y, line 14: Starting a conversation around these numbers @victorcolombo

victorcolombo (approved):

PR #562, opened 2023-04-25
refactor: added support to display warnings to the cli

Peeeekay:
YES/NO

### DIFF
--- a/core/server/api_container/server/api_container_service.go
+++ b/core/server/api_container/server/api_container_service.go
@@ -789,7 +789,6 @@ func (apicService ApiContainerService) runStarlarkPackageSetup(packageId string,
 
 func (apicService ApiContainerService) runStarlark(parallelism int, dryRun bool, packageId string, serializedStarlark string, serializedParams string, stream grpc.ServerStream) {
 	responseLineStream := apicService.startosisRunner.Run(stream.Context(), dryRun, parallelism, packageId, serializedStarlark, serializedParams)
-
 	for {
 		select {
 		case <-stream.Context().Done():
--- a/core/server/api_container/server/api_container_service.go
+++ b/core/server/api_container/server/api_container_service.go
@@ -789,6 +789,7 @@ func (apicService ApiContainerService) runStarlarkPackageSetup(packageId string,
 
 func (apicService ApiContainerService) runStarlark(parallelism int, dryRun bool, packageId string, serializedStarlark string, serializedParams string, stream grpc.ServerStream) {
 	responseLineStream := apicService.startosisRunner.Run(stream.Context(), dryRun, parallelism, packageId, serializedStarlark, serializedParams)
+
 	for {
 		select {
 		case <-stream.Context().Done():
--- a/core/server/api_container/server/startosis_engine/starlark_warning/warning.go
+++ b/core/server/api_container/server/startosis_engine/starlark_warning/warning.go
@@ -1,4 +1,4 @@
-package startosis_warning
+package starlark_warning
 
 import (
 	"fmt"
@@ -27,13 +27,15 @@ var warningMessageSet *sync.Map
 
 var once = new(sync.Once)
 
+const warningMessageValue = true
+
 // PrintOnceAtTheEndOfExecutionf This method stores the warnings in the warning set.
 // The unique constraint is just the warning message, however,
 //TODO: to have more comprehensive unique constraint such as instruction name
 func PrintOnceAtTheEndOfExecutionf(message string, args ...interface{}) {
 	once.Do(initialize)
 	formattedMessage := fmt.Sprintf(message, args...)
-	_, stored := warningMessageSet.LoadOrStore(formattedMessage, true)
+	_, stored := warningMessageSet.LoadOrStore(formattedMessage, warningMessageValue)
 	if !stored {
 		logrus.Tracef("Error occurred while adding warning to the set with message: %v", formattedMessage)
 	}
--- a/core/server/api_container/server/startosis_engine/starlark_warning/warning.go
+++ b/core/server/api_container/server/startosis_engine/starlark_warning/warning.go
@@ -32,6 +32,7 @@ const warningMessageValue = true
 // PrintOnceAtTheEndOfExecutionf This method stores the warnings in the warning set.
 // The unique constraint is just the warning message, however,
 //TODO: to have more comprehensive unique constraint such as instruction name
+//no:lint
 func PrintOnceAtTheEndOfExecutionf(message string, args ...interface{}) {
 	once.Do(initialize)
 	formattedMessage := fmt.Sprintf(message, args...)
--- a/core/server/api_container/server/startosis_engine/starlark_warning/warning.go
+++ b/core/server/api_container/server/startosis_engine/starlark_warning/warning.go
@@ -19,7 +19,7 @@ IS RECEIVED - AND IF NOT THEN ADD IT OTHERWISE IGNORE.
 
 MOST IMPORTANTLY - DO NOT DIRECTLY USE THIS VARIABLE AND INSTEAD USE PrintOnceAtTheEndOfExecutionf METHOD OR OTHER AVAILABLE
 ABSTRACTIONS. THE LONG TERM PLAN IS EITHER TO EXTRACT THIS OUT INTO ITS OWN PACKAGE LIKE stacktrace OR PASS IT DOWN TO
-INSTRUCTIONS ETC. I AM LEANING TOWARDS THE FIRST ONE, BECAUSE WARNING MESSAGES IS AN SEPARATE ENTITY AND DO NOT SEE ANY VALUE
+INSTRUCTIONS ETC. I (PK) LEANING TOWARDS THE FIRST ONE, BECAUSE WARNING MESSAGES IS AN SEPARATE ENTITY AND DO NOT SEE ANY VALUE
 ON COUPLING IT WITH OUR CORE LOGIC - THE WARNING SET ONLY CONTROLS THE STATE FOR WARNING MESSAGE WHICH CAN BE DONE IN ITS
 OWN PACKAGE - HOWEVER IT IS NOT SUPER URGENT ATM
 */
@@ -32,7 +32,6 @@ const warningMessageValue = true
 // PrintOnceAtTheEndOfExecutionf This method stores the warnings in the warning set.
 // The unique constraint is just the warning message, however,
 //TODO: to have more comprehensive unique constraint such as instruction name
-//no:lint
 func PrintOnceAtTheEndOfExecutionf(message string, args ...interface{}) {
 	once.Do(initialize)
 	formattedMessage := fmt.Sprintf(message, args...)
@@ -47,11 +46,10 @@ func PrintOnceAtTheEndOfExecutionf(message string, args ...interface{}) {
 // start with empty set
 func GetContentFromWarningSet() []string {
 	once.Do(initialize)
-
 	var warnings []string
-	warningMessageSet.Range(func(key interface{}, value interface{}) bool {
+	warningMessageSet.Range(func(key interface{}, _ interface{}) bool {
 		warnings = append(warnings, fmt.Sprint(key))
-		// once the key is read, remove the map
+		// once the key is read, remove from the map
 		warningMessageSet.Delete(key)
 		return true
 	})
--- a/core/server/api_container/server/startosis_engine/startosis_runner.go
+++ b/core/server/api_container/server/startosis_engine/startosis_runner.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"github.com/kurtosis-tech/kurtosis/api/golang/core/kurtosis_core_rpc_api_bindings"
 	"github.com/kurtosis-tech/kurtosis/api/golang/core/lib/binding_constructors"
+	"github.com/kurtosis-tech/kurtosis/core/server/api_container/server/startosis_engine/startosis_warning"
 	"github.com/sirupsen/logrus"
+	"sync"
 )
 
 type StartosisRunner struct {
@@ -13,6 +15,8 @@ type StartosisRunner struct {
 	startosisValidator *StartosisValidator
 
 	startosisExecutor *StartosisExecutor
+
+	mutex *sync.Mutex
 }
 
 const (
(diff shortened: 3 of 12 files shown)


Peeeekay:
In this PR - I am adding a facility to pipe warnings from starlark to the cli. Currently, i store the warnings during the execution in a Set and return the warnings right after the execution is completed. 

The reason is to make sure - that we are only displaying one more warning at a time - i.e we can call say `add_service` multiple times but the warning must be shown once associated with it. 

Another important consideration was to show all the warnings at the end of the execution; showing it during the execution is quite complex and also has multitude of edge-cases which I believe is not important for v0

h4ck3rk3y:
@Peeeekay can you share a screen shot with some demos?

core/server/api_container/server/api_container_service.go
```diff
@@ -671,6 +671,7 @@ func (apicService ApiContainerService) runStarlarkPackageSetup(packageId string,
 
 func (apicService ApiContainerService) runStarlark(parallelism int, dryRun bool, packageId string, serializedStarlark string, serializedParams string, stream grpc.ServerStream) {
 	responseLineStream := apicService.startosisRunner.Run(stream.Context(), dryRun, parallelism, packageId, serializedStarlark, serializedParams)
+
```
h4ck3rk3y: intentional or GoLand formatted this?

core/server/api_container/server/startosis_engine/startosis_runner.go
```diff
@@ -28,15 +32,29 @@ func NewStartosisRunner(interpreter *StartosisInterpreter, validator *StartosisV
...
-
 	go func() {
-		defer close(starlarkRunResponseLines)
+		defer func() {
+			warnings := startosis_warning.GetContentFromWarningSet()
+
+			if len(warnings) > 0 {
+				for _, warning := range warnings {
+					// TODO: create a new binding_constructor for warning message
+					starlarkRunResponseLines <- binding_constructors.NewStarlarkRunResponseLineFromInstructionResult(warning)
```
h4ck3rk3y, line 55: can we make this a typed thing that we return so that on other end we can tell between a warning and another message so that we can color it yellow!

core/server/api_container/server/startosis_engine/startosis_runner.go
```diff
@@ -28,15 +32,29 @@ func NewStartosisRunner(interpreter *StartosisInterpreter, validator *StartosisV
 		startosisInterpreter: interpreter,
 		startosisValidator:   validator,
 		startosisExecutor:    executor,
+		mutex:                &sync.Mutex{},
 	}
 }
 
 func (runner *StartosisRunner) Run(ctx context.Context, dryRun bool, parallelism int, packageId string, serializedStartosis string, serializedParams string) <-chan *kurtosis_core_rpc_api_bindings.StarlarkRunResponseLine {
+	runner.mutex.Lock()
```
h4ck3rk3y, line 43: the defer to unlock this should be right here on line 41

core/server/api_container/server/startosis_engine/startosis_runner.go
```diff
@@ -28,15 +32,29 @@ func NewStartosisRunner(interpreter *StartosisInterpreter, validator *StartosisV
...
+			warnings := startosis_warning.GetContentFromWarningSet()
+
+			if len(warnings) > 0 {
+				for _, warning := range warnings {
+					// TODO: create a new binding_constructor for warning message
+					starlarkRunResponseLines <- binding_constructors.NewStarlarkRunResponseLineFromInstructionResult(warning)
+				}
+			}
+
+			runner.mutex.Unlock()
```
h4ck3rk3y: why is this guy here?

core/server/api_container/server/startosis_engine/startosis_runner.go
```diff
@@ -13,6 +15,8 @@ type StartosisRunner struct {
 	startosisValidator *StartosisValidator
 
 	startosisExecutor *StartosisExecutor
+
+	mutex *sync.Mutex
```
h4ck3rk3y, line 19: we're adding this because

1. We don't support running two scripts against the same Enclave anyway
2. The Warning set can't be shared

Any other reasons? Can we add this as a comment?

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/sirupsen/logrus"
+	"sync"
+)
+
+var warningMessageSet *sync.Map
+
+// Initialize This method is responsible for creating warningMessageSet; if the warning set already exists it will override
+// that with an empty warningMessageSet. This method is intended to be used during initialization process.
+func Initialize() {
+	warningMessageSet = new(sync.Map)
```
h4ck3rk3y: And we aren't doing a once so that we don't have dupe warnings when a script is run against the same Enclave?

Can you put the reason on top?

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func PrintOnceAtTheEndOfExecutionf(message string, args ...interface{}) {
+	if warningMessageSet != nil {
+		formattedMessage := fmt.Sprintf(message, args...)
+		_, stored := warningMessageSet.LoadOrStore(formattedMessage, true)
```
h4ck3rk3y: lets take out this "True" to a constant

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
```diff
@@ -0,0 +1,38 @@
...
+		if !stored {
+			logrus.Warnf("Error occurred while adding warning to the set with message: %v", formattedMessage)
```
h4ck3rk3y: This will be too noisy! This should be debug or trace and something more direct like "not storing message '%v' as its already in the set" or something

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
```diff
@@ -0,0 +1,38 @@
...
+		}
+	}
```
h4ck3rk3y: what if the warning message set is nil? do we fail silently? seems off

throwing it out there - can we initialize it the first time its called if it isn't initialized? the current way seems a bit concerning. our code path continues to work if I go and remove the `startosis_warning.Initialize()`

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func GetContentFromWarningSet() []string {
+	var warnings []string
+
+	if warningMessageSet != nil {
```
h4ck3rk3y: this feels off too!

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
```diff
@@ -0,0 +1,38 @@
...
+		warningMessageSet.Range(func(key interface{}, value interface{}) bool {
```
h4ck3rk3y: can this be typed

warningMessage string, _ bool

?

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
```diff
@@ -0,0 +1,38 @@
+package startosis_warning
```
h4ck3rk3y: uncomfortable with this PR as this isn't tested. I could have lived with that if I had seen a working demo!

h4ck3rk3y: also can this be starlark_warning now

core/server/api_container/server/startosis_engine/startosis_runner.go
Peeeekay: i will do that in the next PR. will add another method in binding_constructor for that @h4ck3rk3y

core/server/api_container/server/startosis_engine/startosis_runner.go
Peeeekay: nope; those are the reasons - cool will add these.

core/server/api_container/server/startosis_engine/startosis_runner.go
Peeeekay: Though it does not matter; because we only run go-routine per invocation; but yeah your way is more verbose. will do that.

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
Peeeekay: that's a good suggestion; let me make it trace.

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
Peeeekay: I was debating between making it silent failures or not - because it's warnings and don't want to discomfort users with it's failure; but yeah in the hindsight it's fine because it can be caught while development if it the error is verbose. Let me wrap it in `once` to actually avoid it.

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
Peeeekay: I demoed it today - that should be it? @h4ck3rk3y

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
Peeeekay: i don;t think so i can do that; i would have to just go with interface atm.

core/server/api_container/server/startosis_engine/starlark_warning/warning.go
```diff
@@ -0,0 +1,65 @@
...
+INSTRUCTION INVOCATION. WE KEEP TRACK OF ALL THE WARNINGS IN THIS SET, AND ONCE THE EXECUTION IS COMPLETED WE DISPLAY ALL
+THE WARNINGS TO THE CLI, AND ALSO CLEAR THE SET - SO THAT FOR NEXT RUN WE HAVE AN EMPTY SET.
+
+WE CURRENTLY DO IT BY HAVING A UNIQUE CONSTRAINT ON THE WARNING MESSAGE.THIS IS DONE JUST TO GET THE BALL ROLLING
+- HOWEVER IN FUTURE - USING VISITOR PATTERN WHERE WE KEEP TRACK OF WHETHER ALL THE WARNINGS FROM AN INSTRUCTION/TYPE
+IS RECEIVED - AND IF NOT THEN ADD IT OTHERWISE IGNORE.
+
+MOST IMPORTANTLY - DO NOT DIRECTLY USE THIS VARIABLE AND INSTEAD USE PrintOnceAtTheEndOfExecutionf METHOD OR OTHER AVAILABLE
+ABSTRACTIONS. THE LONG TERM PLAN IS EITHER TO EXTRACT THIS OUT INTO ITS OWN PACKAGE LIKE stacktrace OR PASS IT DOWN TO
+INSTRUCTIONS ETC. I AM LEANING TOWARDS THE FIRST ONE, BECAUSE WARNING MESSAGES IS AN SEPARATE ENTITY AND DO NOT SEE ANY VALUE
```
h4ck3rk3y: put (PK) in paranthesis in the I

core/server/api_container/server/startosis_engine/starlark_warning/warning.go
```diff
@@ -0,0 +1,65 @@
...
+	if !stored {
+		logrus.Tracef("Error occurred while adding warning to the set with message: %v", formattedMessage)
+	}
+}
+
+// GetContentFromWarningSet This method retrieves and deletes all the warnings from set. The main idea
+// is that after calling this method the set will be empty and make sure that for next starlark run - we
+// start with empty set
+func GetContentFromWarningSet() []string {
+	once.Do(initialize)
```
h4ck3rk3y, line 48: cool; so for multiple runs we still have the same set so we don't pop the same warning twice

h4ck3rk3y (approved):

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
h4ck3rk3y: sure. can we pop in a unit test if its like <15 minutes of work? otherwise i have approved this!

core/server/api_container/server/startosis_engine/starlark_warning/warning.go
```diff
@@ -0,0 +1,65 @@
...
+
+	var warnings []string
+	warningMessageSet.Range(func(key interface{}, value interface{}) bool {
```
h4ck3rk3y: value could be `_`

core/server/api_container/server/startosis_engine/startosis_warning/warning.go
adschwartz: absolutely, should have a test

core/server/api_container/server/startosis_engine/starlark_warning/warning.go
```diff
@@ -0,0 +1,65 @@
...
+		warnings = append(warnings, fmt.Sprint(key))
+		// once the key is read, remove the map
+		warningMessageSet.Delete(key)
```
h4ck3rk3y, line 53: oh we remove from the map so that if the APIC is run again we don't print old warnings but then we might print the old warnings anyway if we find a warning while interpretation/execution of the second run

we could also store -  warning(string) -> hasBeenPrinted(value)

if warning isn't in map we add it (simple), with printed set to false
if it is we do nothing

while GetContentFromWarningSet if its not printed, we print it and set the boolean to true

thoughts?

h4ck3rk3y (approved):

core/server/api_container/server/startosis_engine/starlark_warning/warning.go
Peeeekay: I don't think so we should suppress warning; meaning if a user has run the starlark again -- they should still see the same warnings. I don;t think so caching warnings for same enclave is needed atm - but this optimization can happen later.

h4ck3rk3y: That works for me man. Agree, we can always add it later if people find it noisy :)

PR #565, opened 2023-04-25
fix: random error message after execution

Peeeekay:
This closes : https://app.zenhub.com/workspaces/engineering-636cff9fc978ceb2aac05a1d/issues/gh/kurtosis-tech/kurtosis/529


cli/cli/out/file_logger.go
```diff
@@ -41,7 +42,7 @@ func GetFileLogger() *logrus.Logger {
 
 func RemoveLogFiles() {
 	files, err := os.ReadDir(kurtosisCliLogsRootFolder)
```
h4ck3rk3y: Instead can we check if it's not set before doing the read dir?

h4ck3rk3y (approved):

leeederek:
Hey @Peeeekay, this looks like it's user facing. I've added the label now but for next time, can you add this detail in the PR description? Thank you!

Peeeekay:
@leeederek I do not know if it is; this bug was found internally and it's just cosmetic bug. It does not need to be added in the changelog in my opinion.

PR #585, opened 2023-04-28
fix: revert networking change that limited us to a private ip address range

h4ck3rk3y:
## Description:

This seems to fix network partitioning for the Ethereum (locally) but this will break weaviate inside AutoGPT as that doesn't like non private networks

The Starlark for docker compose here https://weaviate.io/developers/weaviate/installation/docker-compose will stop working

Essentially reverts 003f190af636f76009fac34899d8b51ef5dad901

YES


h4ck3rk3y:
Closing this in favor of the recent 172.x changes

PR #598, opened 2023-05-09
fix: Enable autocomplete for the `files upload` path argument

### LINKED ISSUE #254: [bug]: The `path` element of `kurtosis files upload` doesn't tab-complete files

0.68.10

Papercut, this bug is frustrating, but I have a workaround.

Enter `kurtosis files upload SOME_ENCLAVE` and try to tab-complete the path. Nothing happens.

### Relevant log output

_No response_

### Desired behavior

Tab-completion of the path variable works

laurentluce:
## Description:
Make the `files upload` path arg a `NewFilepathOrDirpathArg` instance to support autocomplete in addition to validation.

NO

Closes #254 
The remaining commands will be addressed in #601


cli/cli/commands/files/upload/upload.go
```diff
@@ -51,10 +52,11 @@ var FilesUploadCmd = &engine_consuming_kurtosis_command.EngineConsumingKurtosisC
...
 			isEnclaveIdArgGreedy,
 		),
-		{
-			Key:            pathArgKey,
-			ValidationFunc: validatePathArg,
-		},
+		file_system_path_arg.NewFilepathOrDirpathArg(
+			pathArgKey,
+			isPathArgOptional,
+			nil,
```
mieubrisse: can we swap the `nil` magic constant for a named constant per https://github.com/kurtosis-tech/kurtosis-internal-docs/blob/main/general-style-guide.md#replace-magic-constants-with-constants-named-for-their-purpose-not-their-value so it's clear what the arg is?

cli/cli/commands/files/upload/upload.go
```diff
@@ -97,15 +99,3 @@ func run(
 	logrus.Infof("Files package '%v' uploaded with UUID: %v", fileArtifactName, filesArtifactUuid)
 	return nil
 }
-
-func validatePathArg(_ context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error {
```
mieubrisse, line 101: Are we removing this because the filepath arg already does this?

docs/docs/cli-reference/files-upload.md
```diff
@@ -7,8 +7,11 @@ slug: /files-upload
 Files can be stored as a [files artifact][files-artifacts] inside an enclave by uploading them:
 
 ```bash
-kurtosis files upload $PATH_TO_FILES
+kurtosis files upload $THE_ENCLAVE_IDENTIFIER $PATH_TO_FILES
```
mieubrisse, line 10: niceeeee

mieubrisse (approved):
approved pending comments

cli/cli/commands/files/upload/upload.go
laurentluce: > Are we removing this because the filepath arg already does this?

Yes

cli/cli/commands/files/upload/upload.go
laurentluce: I added a function set to nil to be used for this use case.  @mieubrisse Let me know if that works.

PR #603, opened 2023-05-09
refactor: upgraded quickstart docs now service readiness is checked by the automatic ports wait feature

### LINKED ISSUE #587: Upgrade the quickstart docs and the quickstart gitpod package with the automatic port wait
### What's your CLI version?

0.76.3

We can reduce the quickstart example code after we release the automatic port wait

### Desired behavior

The quickstart readme docs with fewer code lines

### LINKED ISSUE #588: Add a header before the service logs when the automatic por wait fails
### What's your CLI version?

0.76.3

When the automatic port wait fails the service's logs are printed in order to identify the root cause, but it isn't easy to detect where the execution error text ends and the service's logs start

### Desired behavior

Same error but with a `service's logs headers` between the two texts

leoporoli:
## Description:
upgraded quickstart docs now service readiness is checked by the automatic ports wait feature

YES

Fixes #587 
Fixes #588

### DIFF
--- a/docs/docs/quickstart.md
+++ b/docs/docs/quickstart.md
@@ -244,6 +244,7 @@ def run(plan, args):
     )
 
     # Load the data into Postgres
+    postgres_flags = ["-U", POSTGRES_USER,"-d", POSTGRES_DB]
     plan.exec(
         service_name = "postgres",
         recipe = ExecRecipe(command = ["pg_restore"] + postgres_flags + [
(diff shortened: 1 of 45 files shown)


go.work.sum
```diff
@@ -0,0 +1,10 @@
+cloud.google.com/go/compute/metadata v0.2.0/go.mod h1:zFmK7XCadkQkj6TtorcaGlCW1hT1fIilQDwofLpJ20k=
```
mieubrisse, line 1: Not sure what this file is, so just double-check if it's supposed to be here or not (maybe 1.18?)

mieubrisse (approved):
Really awesome job; great detail-orientedness

docs/docs/quickstart.md
```diff
@@ -579,12 +543,11 @@ UUID           Name
 ========================================== User Services ==========================================
 UUID           Name        Ports                                                Status
 45b355fc810b   postgres    postgres: 5432/tcp -> postgresql://127.0.0.1:59821   RUNNING
-80987420176f   api         http: 3000/tcp                                       STOPPED
 ```
 
-From the above, the problem is clear now: the PostgREST service (named: `api`) status is `STOPPED`, rather than `RUNNING`. 
+From the above, we can infer that the PostgREST service (named: `api`) is not created, so there were an error during the creation face.
```
mieubrisse: s/there were/there was/

s/face/phase/

mieubrisse: Oop also, I think it's been created but it just crashed

go.work.sum
leoporoli: This file pinned some module's dependencies declared in the module's go.mod files, more here: https://go.googlesource.com/proposal/+/master/design/45713-workspace.md#files This file was autogenerated when building the modules after @laurentluce added the Go [workspace feature here](https://github.com/kurtosis-tech/kurtosis/pull/596) and I'm adding the file in the VCS because we do same for mdoule's go.sum files, but we could ignore it if you @mieubrisse  think that it's the best.

leoporoli: if you @mieubrisse think that we should ignore it, let me know and I'll do the change in a next PR

docs/docs/quickstart.md
leoporoli: I rephrase the sentence

PR #608, opened 2023-05-11
feat: Add Windows support for CLI

victorcolombo:
## Description:

After this PR, CLI can work natively on Windows

So we didn't have to deal with cross compilation, Windows binary is compiled with `CGO_ENABLED=0`, and LSP specific code is hidden using `go:build` tag.

This breaks LSP (see https://github.com/kurtosis-tech/kurtosis/issues/609) on that OS.

YES


cli/cli/.goreleaser.yml
```diff
@@ -10,11 +10,17 @@ builds:
...
     goarch:
       - amd64
       - arm64
     env:
-      - CGO_ENABLED=1
+      - >-
+          {{- if eq .Os "windows" }}
+            CGO_ENABLED=0
+          {{- else }}
+            CGO_ENABLED=1
```
h4ck3rk3y: we need CGO_ENABLED For the LSP to work :(

victorcolombo: Added more information on the PR

api/typescript/scripts/build.sh
```diff
@@ -11,5 +11,6 @@ lang_root_dirpath="$(dirname "${script_dirpath}")"
 # ==================================================================================================
 cd "${lang_root_dirpath}"
 yarn install
-CGO_ENABLED=0 yarn test
+# There are no unit tests currently
+# CGO_ENABLED=0 yarn test
```
h4ck3rk3y, line 15: even if there were why have CGO_ENABLED in yarn??

h4ck3rk3y (approved):
Godspeed

api/typescript/scripts/build.sh
victorcolombo: my guess is that this was a product of some copy and paste of go build scripts
https://github.com/kurtosis-tech/kurtosis/blob/be52f9e73c5cd63e09f5c2343add165886bd7313/api/golang/scripts/build.sh#L15-L20

PR #613, opened 2023-05-11
refactor: removed replace statements

leoporoli:
## Description:
Since we activated Go workspaces adding the `go.work` file in the root the replace statements are no longer needed, so I removed all of them and I have to replace the module's versions because these were using v0.0.0, which is not an acceptable version, and fails the build, the versions now are pinned but the workspace config allows us to get the module from the local instead of the remote.

NO

[Workspaces added here](https://github.com/kurtosis-tech/kurtosis/pull/596)


leoporoli:
hey, @victorcolombo could you confirm if this change helps to fix this: https://app.zenhub.com/workspaces/engineering-636cff9fc978ceb2aac05a1d/issues/gh/kurtosis-tech/kurtosis/356 ?

victorcolombo:
Question for you @leoporoli: Now it seems we are *pinning* packages to a specific version `https://github.com/kurtosis-tech/kurtosis/pull/613/files#diff-23f0a84994f6bf6a2925d4759fbe3f163bb44a5ebeb0a754fb98e7a707142942R7`.

My understanding is that workspaces, when built locally, will replace that version with the local version. But what the behavior when someone pull the API package using `go get`? Do they get the latest version of the dependencies? Or do they get the version we defined on `go.mod`

leoporoli:
> Question for you @leoporoli: Now it seems we are _pinning_ packages to a specific version `https://github.com/kurtosis-tech/kurtosis/pull/613/files#diff-23f0a84994f6bf6a2925d4759fbe3f163bb44a5ebeb0a754fb98e7a707142942R7`.
> 
> My understanding is that workspaces, when built locally, will replace that version with the local version. But what the behavior when someone pull the API package using `go get`? Do they get the latest version of the dependencies? Or do they get the version we defined on `go.mod`

@victorcolombo This is a good point and the answer is: `The behavior will be that they will get the pinned version instead the last one and yes it's not the best,` by the way, the current state does not allow users to `go get the modules` because these fails (probably it was on purpose I don't remember what was the plan) for instance if you try `go get github.com/kurtosis-tech/kurtosis/core/files_artifacts_expander` in another repository you will see that it can download the v0.0.0 

```
go-module-test ❯ go get github.com/kurtosis-tech/kurtosis/core/files_artifacts_expander                                                                                                   
go: downloading github.com/kurtosis-tech/kurtosis/core/files_artifacts_expander v0.0.0-20230512122355-3ccbd7368cfa
go: downloading github.com/kurtosis-tech/kurtosis/api/golang v0.0.0
go: downloading github.com/kurtosis-tech/kurtosis/grpc-file-transfer/golang v0.0.0
github.com/kurtosis-tech/kurtosis/core/files_artifacts_expander imports
	github.com/kurtosis-tech/kurtosis/api/golang/core/kurtosis_core_rpc_api_bindings: reading github.com/kurtosis-tech/kurtosis/api/golang/api/golang/go.mod at revision api/golang/v0.0.0: unknown revision api/golang/v0.0.0
github.com/kurtosis-tech/kurtosis/core/files_artifacts_expander imports
	github.com/kurtosis-tech/kurtosis/grpc-file-transfer/golang/grpc_file_streaming: reading github.com/kurtosis-tech/kurtosis/grpc-file-transfer/golang/grpc-file-transfer/golang/go.mod at revision grpc-file-transfer/golang/v0.0.0: unknown revision grpc-file-transfer/golang/v0.0.0
```

So, I think the best decision here will be to freeze this PR until we create a script or something that upgrades the mono repo Kurtosis dependencies versions before releasing a new version. 
What we will miss?? I think that we will miss having intermediate versions between the releases, and I don't know how to fix this (probably with pre-release versions??)

Let me know your thoughts.

PR #626, opened 2023-05-16
fix: Remove the GRPC proxy port from the engine and from the APIC

### LINKED ISSUE #560: The engine image is still listening on port 9711 even though the GRPC proxy is long dead
### What's your CLI version?

0.75.9

`kurtosis engine start` shows an image listening on 9711 (because the CLI is starting it that way), but there is no more grpc proxy

### Desired behavior

The engine image doesn't listen on port 9711

laurentluce:
## Description:
The engine container listens on the GRPC proxy port even if this port is not used anymore.  This PR removes the GRPC proxy port from the engine and the code supporting it.  We took the opportunity to also remove the GRPC proxy port support from the APIC and the GRPC-WEB support since they are not used anymore.

NO

Closes #560


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -55,46 +54,28 @@ func CreateEngine(
...
 		privateGrpcPortSpec,
-		consts.KurtosisInternalContainerGrpcProxyPortId,
-		privateGrpcProxyPortSpec,
 	)
 	if err != nil {
 		return nil, stacktrace.Propagate(
 			err,
-			"An error occurred getting the engine server container attributes using GUID '%v', grpc port num '%v', and "+
-				"grpc proxy port num '%v'",
+			"An error occurred getting the engine server container attributes using GUID '%v'and grpc port num %v'",
```
leoporoli:
```suggestion
			"An error occurred getting the engine server container attributes using GUID '%v' and GRPC port num '%v'",
```

leoporoli (approved):

PR #634, opened 2023-05-23
feat: Return error on SDK if Starlark run on any step 

victorcolombo:
## Description:
Currently Starlark run remote package, run package and run script blocking calls on the SDK have the following behavior:
1. If an error happens on the APIC side, we return `(*StarlarkRunResult == nil, error != nil)`
2. If an error happens on the Starlark side, we return `(*StarlarkRunResult != nil, error == nil)`, with either `StarlarkRunResult.[InterpretationError, ValidationErrors, ExecutionError] != nil`
3. If no errors happen on the Starlark side, we return `(*StarlarkRunResult != nil, error == nil)` with all `StarlarkRunResult.[InterpretationError, ValidationErrors, ExecutionError] == nil`

This behavior is not very ergonomic, given that most people are only interested if the Starlark succeeded or not, they should be able to do this with a simple `err != nil` check, which is the Go idiomatic way of dong it.

If the user wants to investigate further the interpreted instructions, the breakdown of which phase failed, etc, this will still be accessible via the `StarlarkRunResult`.

YES

https://github.com/ava-labs/subnet-evm/pull/649

### DIFF
--- a/api/golang/core/lib/enclaves/enclave_context.go
+++ b/api/golang/core/lib/enclaves/enclave_context.go
@@ -30,6 +30,7 @@ import (
 	"google.golang.org/protobuf/types/known/emptypb"
 	"io"
 	"path"
+	"strings"
 )
 
 type EnclaveUUID string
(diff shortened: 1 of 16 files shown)


api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -96,9 +97,10 @@ func (enclaveCtx *EnclaveContext) RunStarlarkScript(ctx context.Context, seriali
 func (enclaveCtx *EnclaveContext) RunStarlarkScriptBlocking(ctx context.Context, serializedScript string, serializedParams string, dryRun bool, parallelism int32) (*StarlarkRunResult, error) {
```
h4ck3rk3y, line 97: can you add this to RunRemotesStarlarkScriptBlocking?

victorcolombo: Good catch, done

h4ck3rk3y (approved):
LGTM - please add this to RunRemoteStarlarkScriptBlocking

PR #663, opened 2023-06-01
ci: Apply gofmt to container-engine-lib and api

gbouv:
## Description:
Runs gofmt as a GH action. Not required for now, TBD if we requite it in the future

NO


.github/workflows/gofmt.yml
```diff
@@ -0,0 +1,27 @@
...
+    name: gofmt
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-go@v3
+      with:
+        go-version: 1.19
+    - uses: actions/checkout@v3
+    - uses: Jerome1337/gofmt-action@v1.0.5
+      with:
+        gofmt-path: './api/golang/ ./container-engine-lib/'
```
gbouv, line 26: I'll submit other PRs to fix the other modules and add them here. The ultimate goal is to be able to run it with `./`

leoporoli (approved):

PR #676, opened 2023-06-02
Update README.md

galenmarchetti:
what do you think of this slight modification with extra details

YES/NO


leeederek (approved):

leeederek:
@galenmarchetti too wordy but i see why you did it

PR #680, opened 2023-06-06
docs: make SDK more obvious

leeederek:
## Description:
Some users have expressed confusion and difficulty when trying to find our SDK. We named it "client libraries" in our docs and people are searching for "SDK". This PR renames it back to SDK 

YES


docs/docs/explanations/architecture.md
```diff
@@ -92,13 +92,13 @@ def run(plan, args):
 
 For a list of all the Kurtosis Starlark instructions, [see here][starlark-code-reference].
 
-Client Libraries
+Client Libraries or SDK
```
mieubrisse: I say if we go all in on, SDKs, we should do it and just call this SDK

mieubrisse (approved):

docs/docs/explanations/architecture.md
leeederek:
```suggestion
SDK
```

PR #687, opened 2023-06-07
feat: added min resource constraint for kubernetes

Peeeekay:
YES/NO

### DIFF
--- a/container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/start_user_services.go
+++ b/container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/start_user_services.go
@@ -517,7 +517,7 @@ func getUserServicePodContainerSpecs(
 	memoryAllocationMegabytes uint64,
 	minCpuAllocationMilliCpus uint64,
 	minMemoryAllocationMegabytes uint64,
-) ([]apiv1.Container, error, ) {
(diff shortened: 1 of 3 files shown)


container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/start_user_services.go
```diff
@@ -532,16 +536,37 @@ func getUserServicePodContainerSpecs(
 
 	resourceLimitsList := apiv1.ResourceList{}
 	resourceRequestsList := apiv1.ResourceList{}
+
+	err = checkIfResourcesAreSetProperly(minCpuAllocationMilliCpus, cpuAllocationMillicpus, apiv1.ResourceCPU)
+	if err != nil {
+		return nil, err
+	}
+
+	err = checkIfResourcesAreSetProperly(minMemoryAllocationMegabytes, cpuAllocationMillicpus, apiv1.ResourceMemory)
```
Peeeekay: added this so that we don;t need make request to kubernetes api, and gives me more control around error message.

laurentluce: If we think this validation will rarely return an error, it is probably fine to hit the API and rely on it for validation.  Adding our own validation does not help much if we are checking the same things than the k8s API.

laurentluce (approved):

PR #690, opened 2023-06-08
feat: Add `kurtosis service exec` command

### LINKED ISSUE #688: Kurtosis shell exec does not return the return value of command
### What's your CLI version?

0.77.0

```
$ kurtosis enclave add --name test-enclave
$ kurtosis service add test-enclave test2 httpd --ports http=80
$ kurtosis service shell test-enclave test2 --exec "exit 1"
echo $? # prints 0
```

This is critical for usage in CI systems that check exit code to report failure

### Desired behavior

`echo $?` should print 1

This behavior happens because we use `GetConnectionWithUserService` to act both as a attach terminal and as an exec with the param `commandToRunInsteadOfBash`. https://github.com/kurtosis-tech/kurtosis/blob/68a1ad1d9ed9b6d9abdd1fcebc9bb30d8ec37af9/container-engine-lib/lib/backend_interface/kurtosis_backend.go#L271

I believe `GetConnectionWithUserService` should be reserved for interactive operations, since we are attaching a connection. So getting an interactive shell or even some interactive programs like `cat` (equivalent to `-it` flag in Docker)

For non interactive ones, like `echo`, could simply use `RunUserServiceExecCommands` when running `kurtosis service shell ... --exec ...`, since that command and blocking, and returns all information about the program execution (return status), but does not offer interactivity (maybe `kurtosis service exec` to match Dockers terminology?)

gbouv:
## Description:
Users can now run `kurtosis service exec <ENCLAVE_ID> <SERVICE_ID> 'any sh command'`. The command will be executed inside the designated user service and the result will be printed in the CLI output.

The main different with `kurtosis service shell <ENCLAVE_ID> <SERVICE_ID> --exec 'any sh command'` is that the former expects a zero exit code for the command otherwise throws an exception, whereas the latter returns gracefully whatever the exit code of the command was.

In short, `kurtosis service shell` is designed for interactive sessions and will always return gracefully as long as the session terminated is ended successfully. `kurtosis service exec` provides a way to run a single command and returns immediately, taking the status of the command into account.

Fixes #688 

YES

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -687,10 +687,12 @@ jobs:
       - run: "${KURTOSIS_BINPATH} service add test-enclave test1 httpd --ports http=80"
       - run: "${KURTOSIS_BINPATH} service add test-enclave test2 httpd --ports http=80"
 
-      # Shell
+      # Shell & Exec
       # Emulates GitHub Actions
       - run: "${KURTOSIS_BINPATH} service shell test-enclave test1 --exec 'echo 1' < /dev/null > /dev/null"
       - run: "${KURTOSIS_BINPATH} service shell test-enclave test1 --exec 'echo 1'"
+      - run: "${KURTOSIS_BINPATH} service exec test-enclave test1 'echo hello'"
+      - run: "! ${KURTOSIS_BINPATH} service exec test-enclave test1 'echo hello; exit 1'"
 
       # File commands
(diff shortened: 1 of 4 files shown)


gbouv:
Screenshot 2023-06-08 at 11 42 10

leoporoli (approved):

PR #693, opened 2023-06-08
feat!: Added `main-file` and `main-function-name` flags to the `kurtosis run` CLI command. These new options were also added in the  `RunStarlarkScript`, `RunStarlarkPackage` and  the `RunStarlarkRemotePackage` SDKs methods, users will have to update the calls.

leoporoli:
## Description:
Added `main-file` and `main-function-name` flags to the `kurtosis run` CLI command

YES


api/protobuf/core/api_container_service.proto
```diff
@@ -176,6 +176,9 @@ message RunStarlarkScriptArgs {
 
   // Defaults to 4
   optional int32 parallelism = 4;
+
+  // The name of the main function, the default value is "run"
```
gbouv, line 180: If there a default, should we make this argument optional? Or we consider empty string means no value? Maybe that's fine

gbouv (approved):

api/protobuf/core/api_container_service.proto
leoporoli: the empty string here means use the default value, which is 'run'

I think both versions are the same because when you use an `optional string` like for example `optional string subnetwork ` then when you call the `GetSubnetwork()` it returns an empty string if the value is nil. 

image

image

but I could be missing some pattern here, so i'm all ears

PR #703, opened 2023-06-09
feat: random function args

leoporoli:
## Description:
random function args

YES


core/server/api_container/server/startosis_engine/startosis_interpreter_test.go
```diff
@@ -56,6 +58,34 @@ def run(plan):
 	validateScriptOutputFromPrintInstructions(t, instructions, expectedOutput)
 }
 
+func TestStartosisInterpreter_RandomMainFunctionAndParams(t *testing.T) {
```
leoporoli, line 61: I was able to implement receive random function args, but it still missing to implement that one of these arguments is a dictionary, in order to accomplish the HugoBytes case, which is run this method: https://github.com/HugoByte/chain-package/blob/development/services/jvm/icon/src/node-setup/contract_deploy.star#LL5C1-L5C72

Peeeekay:
LGTM

Peeeekay (approved):

PR #712, opened 2023-06-13
feat: Remove `--exec` flag for `kurtosis service shell`

gbouv:
## Description:
`--exec` flag for `kurtosis service shell` command is getting deleted in favour of the new `kurtosis service exec` command.

YES


victorcolombo (approved):
This is a breaking change because if someone used this in their CI, this will break them. Do you mind changing this to `refactor!:` so we bump the minor?

PR #714, opened 2023-06-13
fix: accept run as the default main function

h4ck3rk3y:
## Description:
runRemoteStarlarkPackage would break if a user passed "run" as the mainFunctionName instead of ""
This fixes it

YES


h4ck3rk3y:
Context - https://kurtosistech.slack.com/archives/C032ZB39AHH/p1686632988891109

gbouv (approved):

PR #716, opened 2023-06-13
feat: Automatically inject the plan object if the first argument of the main function is `plan`

leoporoli:
## Description:
Automatically inject the plan object if the first argument of the main function is `plan`

YES


core/server/api_container/server/startosis_engine/startosis_interpreter.go
```diff
@@ -124,16 +122,21 @@ func (interpreter *StartosisInterpreter) Interpret(
...
 	var kwArgs []starlark.Tuple
 
-	if shouldInjectPlanArg || (isUsingDefaultMainFunction && mainFunction.NumParams() >= minimumParamsRequiredForPlan) {
-		if paramName, _ := mainFunction.Param(planParamIndex); paramName != planParamName {
-			return "", nil, startosis_errors.NewInterpretationError(unexpectedArgNameError, planParamIndex, planParamName, paramName).ToAPIType()
+	mainFuncParamsNum := mainFunction.NumParams()
+
+	if mainFuncParamsNum >= minimumParamsRequiredForPlan {
+		// the plan object will always be injected if the first argument name is 'plan'
+		firstParamName, _ := mainFunction.Param(planParamIndex)
```
h4ck3rk3y, line 129: cool the plan param can only be the first param?

h4ck3rk3y: can we for loop a non run functions params and see which one is plan?

h4ck3rk3y (approved):

core/server/api_container/server/startosis_engine/startosis_interpreter.go
leoporoli: yes, I think we could do that, I'll do it on the next PR. I want to have this merged

PR #719, opened 2023-06-13
fix: add `continue` to avoid segfault on service failing to register

adschwartz:
## Description:
When a service fails to register we wrongfully continue to process and eventually call `.GetUUID` on a nulled service registration which in turn causes a segfault.

NO


gbouv (approved):
nice catch!

PR #722, opened 2023-06-13
chore: TestSuite added on the Startosis request wait assert test package

leoporoli:
YES/NO


internal_testsuites/golang/.golangci.yml
```diff
@@ -0,0 +1,19 @@
...
+        - typecheck
+    - path: "testsuite/startosis_package_test"
+      linters:
+        - typecheck
+    - path: "testsuite/startosis_ports_wait_test"
+      linters:
+        - typecheck
+    - path: "testsuite/startosis_subpackage_test"
+      linters:
+        - typecheck
```
gbouv: newline

internal_testsuites/golang/testsuite/startosis_add_service_test/startosis_add_service_testsuite_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/stretchr/testify/suite"
+	"testing"
+)
+
+const (
+	name                  = "startosis-add-service"
+	isPartitioningEnabled = false
+)
+
+type StartosisAddServiceTestSuite struct {
```
gbouv, line 18: That's super cool the concept of suite here, I think we could probably do something similar with the SL framework test we have

gbouv (approved):
I like the concept of TestSuite here, this is quite nice 👏  Sad that is breaks the encapsulation of each tests, but maybe for now it's fine.

Worst case if we can't find a good solution to speed up enclave creation, maybe we can implement an endpoint on the APIC that manually removes all containers and all files artifact from the enclave, and we can call this endpoing after each test. That being said, removing all resources will slow down each test in the end, and maybe we will be back to bad perf again... So, if we ever consider doing this, we should try it carefully first

As a side note, I think we don't have all those tests in TS, but mayybe we should so something similar for the few we have?  Might not be worth the investment though, I'll let you take the decision here

PR #734, opened 2023-06-15
chore: Run exec in parallel in the backend

gbouv:
## Description:
The backend now supports running exec command in parallel

NO


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/user_services_functions/run_user_service_exec_commands.go
```diff
@@ -23,57 +24,110 @@ func RunUserServiceExecCommands(
...
+		execOperation := createExecOperation(ctx, serviceUuid, userServiceDockerResource, commandArg, dockerManager)
+		execOperations[execOperationId] = execOperation
+	}
+
+	successfulOperations, failedOperations := operation_parallelizer.RunOperationsInParallel(execOperations)
+	for operationUuid, operationResult := range successfulOperations {
+		serviceUuid := service.ServiceUUID(operationUuid)
+		execResult, ok := operationResult.(*exec_result.ExecResult)
+		if !ok {
+			return nil, nil, stacktrace.NewError("An error occurred processing the result of the exec command "+
```
h4ck3rk3y, line 95: Makes sense to exit here as this isn't a failed exec but something else completely.

h4ck3rk3y: Or should we add it to the `failedExecs` dictionary if there are others that might have succeeded?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/run_user_service_exec_commands.go
```diff
@@ -46,69 +49,108 @@ func RunUserServiceExecCommands(
...
-			return nil, nil, stacktrace.NewError(
+
+		userServiceKubernetesService := userServiceKubernetesResource.Service
+		if userServiceKubernetesService == nil {
+			failedExecs[serviceUuid] = stacktrace.NewError(
 				"Cannot execute command '%+v' on service '%v' because the service is not started yet",
 				commandArgs,
-				guid,
+				serviceUuid,
 			)
```
h4ck3rk3y, line 90: should there be a continue in here? `userServiceKubernetesService` can be a nil

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/run_user_service_exec_commands.go
```diff
@@ -46,69 +49,108 @@ func RunUserServiceExecCommands(
...
-			return nil, nil, stacktrace.NewError(
+		if userServiceKubernetesService.GetStatus() != container_status.ContainerStatus_Running {
+			failedExecs[serviceUuid] = stacktrace.NewError(
 				"Cannot execute command '%+v' on service '%v' because the service status is '%v'",
 				commandArgs,
-				guid,
-				serviceObj.GetStatus().String(),
+				serviceUuid,
+				userServiceKubernetesService.GetStatus().String(),
 			)
```
h4ck3rk3y, line 99: similarly here. should we have a `continue`?

Victor told me about a bug in another one of our parallel operations where we ran into seg faults because of this yesterday 😅

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -841,87 +842,95 @@ func (network *DefaultServiceNetwork) UnpauseService(ctx context.Context, servic
...
-		}
-		return 0, "", stacktrace.NewError(
-			"One or more errors occurred attempting to exec command(s) in the service(s): \n%v",
-			strings.Join(
-				serviceExecErrs,
-				"\n\n",
-			),
-		)
+
+	if execResult, found := successfulExecCommands[serviceUuid]; found {
```
h4ck3rk3y, line 871: For a user to see this they will have to be looking at the logs in APIC? When would this happen?

Is this something we should just fail on instead?

h4ck3rk3y: Curious when line 872 would be true compared to line 880 being true?

core/server/api_container/server/startosis_engine/kurtosis_instruction/run_sh/run_sh.go
```diff
@@ -205,14 +206,14 @@ func (builtin *RunShCapabilities) Execute(ctx context.Context, _ *builtin_argume
...
-	code, output, err := builtin.serviceNetwork.ExecCommand(ctx, builtin.name, createDefaultDirectory)
+	createDefaultDirectoryResult, err := builtin.serviceNetwork.RunExec(ctx, builtin.name, createDefaultDirectory)
 	if err != nil {
 		return "", stacktrace.Propagate(err, fmt.Sprintf("error occurred while executing one time task command: %v ", builtin.run))
 	}
 
 	result := map[string]starlark.Comparable{
-		runshOutputKey: starlark.String(output),
-		runshCodeKey:   starlark.MakeInt(int(code)),
+		runshOutputKey: starlark.String(createDefaultDirectoryResult.GetOutput()),
```
h4ck3rk3y, line 215: I suppose we can add `GetStdError` or something to the execResult obj in the future or pull it from there into starlark

h4ck3rk3y (approved):

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/user_services_functions/run_user_service_exec_commands.go
gbouv: I took some inspiration from what we do in add_services here because I was conflicted the same way you are.
I think failing hard is right here. Basically if you follow the codepath downstream, there's no real reason why one would fail for this reason and not the other (because the ExecResult object is computed the exact same way for all execs that a run in parallel). So, I believe that's the reasoning behind failing hard here, and it kind of made sense to me. I can add this as a comment in the code to make it clearer

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/run_user_service_exec_commands.go
gbouv: Yup you're totally right

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/run_user_service_exec_commands.go
gbouv: Huh, yes here too, good catch!

core/server/api_container/server/service_network/default_service_network.go
gbouv: In this case this would be a Kurtosis internal bug, and I can't really find a reason why we would end up in this situation. But in the code I replaced before, we were doing something similar, so I guess that guided me toward being extra conservative. I would be fine throwing an error here. Again, this should never happen

To your second question, this would be hit with the following maps returned by the backend:
```
successfulExecCommands -> {
    serviceUuid -> ExecResult(...)
},
failedExecCommands -> {
    [serviceUuid OR someOtherRandomUuidOutOfNowhere] -> err
}
```

whereas the second clause line 880 would be hit with:
```
successfulExecCommands -> {},
failedExecCommands -> {}
```

So, slightly different, but overall both would indicate a pretty serious and unexpected bug, so I'll fail in this case as well

core/server/api_container/server/startosis_engine/kurtosis_instruction/run_sh/run_sh.go
gbouv: I think right now we write StdOut and StdErr to the same buffer, but yes if we want to differentiate between the two at some point it should be doable

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/user_services_functions/run_user_service_exec_commands.go
h4ck3rk3y: I can live with that 👍 !

core/server/api_container/server/service_network/default_service_network.go
h4ck3rk3y: Sounds good to me :)

core/server/api_container/server/startosis_engine/kurtosis_instruction/run_sh/run_sh.go
h4ck3rk3y: Oh yes you are right! I have seen the red out put at the end. We don't print new lines properly on it and I see `\n` but I do see the errors. 

I'll [create a ticket](https://github.com/kurtosis-tech/kurtosis/issues/735) for the exec error being a bit messy (new lines not rendered properly) separately

gbouv: Yup we can track that separately, shouldn't be too difficult to fix, hopefully

PR #736, opened 2023-06-15
fix: Support for reconnects in the Gateway port forwarder

### LINKED ISSUE #726: The gateway port forwarder should handle network interruptions
### What's your CLI version?

0.78.4

The port forwarder connections drop when there are network interruptions which is expected.  There is no reconnect mechanism though.

### Desired behavior

The port forwarder connections management should support reconnects so network interruptions are transparent to the user.

laurentluce:
## Description:
The Gateway port forwarding logic does not try to reconnect when there are network interruptions.  The current workaround is to restart the Gateway.  This PR adds reconnect logic around the Gateway port forwarder.  The retries will stop when a service is removed via `RemoveService` or a Starlark script.

YES

Closes #726


cli/cli/kurtosis_gateway/connection/connection.go
```diff
@@ -34,14 +35,11 @@ type GatewayConnectionToKurtosis interface {
 type gatewayConnectionToKurtosisImpl struct {
 	localAddresses []string
 	localPorts     map[string]*port_spec.PortSpec
-	// kubectl command to exec to to start the tunnel
-	portforwarder *portforward.PortForwarder
```
laurentluce, line 38: Remove unused attributes.

cli/cli/kurtosis_gateway/connection/connection.go
```diff
@@ -147,7 +181,8 @@ func newLocalPortToPodPortConnection(kubernetesRestConfig *k8s_rest.Config, podP
 }
 
 func (connection *gatewayConnectionToKurtosisImpl) Stop() {
-	connection.portforwarder.Close()
+	logrus.Infof("Closing connection to pod: %s", connection.urlString)
+	close(connection.portforwarderStopChannel)
```
laurentluce, line 195: We close the port forwarder using the stop channel which is the recommended way.

cli/cli/kurtosis_gateway/server/api_container_gateway/api_container_gateway_service_server.go
```diff
@@ -432,6 +434,28 @@ func (service *ApiContainerGatewayServiceServer) idempotentKillRunningConnection
...
+
+	serviceUuids := map[string]bool{}
+	for serviceId, serviceInfo := range serviceInfos {
+		if err := service.writeOverServiceInfoFieldsWithLocalConnectionInformation(serviceInfo); err != nil {
+			return stacktrace.Propagate(err, "Expected to be able to write over service info fields for service '%v', instead a non-nil error was returned", serviceId)
+		}
+		serviceUuids[serviceInfo.GetServiceUuid()] = true
+	}
+
+	if cleanupRemovedServices {
```
laurentluce, line 423: This logic closes the port forwarder for services removed during a Starlark script run.  We have to do that to stop the reconnect process.  It also makes things cleaner compare to what we have today where removed services stay forever in the user service to local connection map.

cli/cli/kurtosis_gateway/connection/connection.go
```diff
@@ -147,7 +182,8 @@ func newLocalPortToPodPortConnection(kubernetesRestConfig *k8s_rest.Config, podP
 }
 
 func (connection *gatewayConnectionToKurtosisImpl) Stop() {
-	connection.portforwarder.Close()
+	logrus.Infof("Closing connection to pod: %s", connection.urlString)
```
adschwartz, line 185: @laurentluce this may be nitpicking a bit, but maybe we should generalize the language: 
"Closing connection to pod" => "Closing connection to service"
Our main use of the gateway is for k8s, but we could potentially apply this to other backends.

cli/cli/kurtosis_gateway/connection/connection.go
```diff
@@ -89,17 +87,56 @@ func newLocalPortToPodPortConnection(kubernetesRestConfig *k8s_rest.Config, podP
...
+							portStrings = append(portStrings, portString)
+						}
+					} else {
+						if retries == 0 {
+							// Exit the retry logic if the first try to connect fails
+							logrus.Errorf("Expected to be able to start forwarding local ports to remote ports, instead our portforwarder has returned a non-nil err:\n%v", err)
+							return
+						}
+						logrus.Debugf("Error trying to forward ports:\n%v", err)
+					}
```
adschwartz, line 137: @laurentluce I'm probably missing something here but I struggle with this logic:
If the error is not `ErrLostConnectionToPod` and retries >0 we then basically continue to retry indefinitely? Is that the intention? 

What types of errors would we get if not `ErrLostConnectionToPod`? My assumption is that it'll be a set of more or less unrecoverable issues.  I guess I would expect the logic to be: If error is not `ErrLostConnectionToPod` we log an error.

Let me know your thoughts

adschwartz (approved):
@laurentluce Looks good to me. Have you tested this change in a k8s cluster (or some other way) where you can test the impact of the change, and verify that the reconnection indeed works? Running in GKE showed many instances of pods being rescheduled so I imagine we can try it there if nothing else.

cli/cli/kurtosis_gateway/connection/connection.go
laurentluce: @adschwartz `ErrLostConnectionToPod` is returned when a previously working connection breaks.  Other errors will be connection initiation failures.  Once the connection breaks, the logic tries to reconnect every 5 seconds and we are going to get connection initiation failures until the pod is accessible again.  I chose the debug level here to not flood the gateway output until we are able to reconnect.

laurentluce:
> @laurentluce Looks good to me. Have you tested this change in a k8s cluster (or some other way) where you can test the impact of the change, and verify that the reconnection indeed works? Running in GKE showed many instances of pods being rescheduled so I imagine we can try it there if nothing else.

I ran some tests against minikube by modifying the pod restart policy, killing the container and issuing some requests to the locally mapped port.  The gateway is able to reconnect when a new container is up.  The gateway is also re-using the same local ports.  I also tested with services removal and made sure the reconnect logic does not happen.

PR #741, opened 2023-06-18
Prototype idempotent run

gbouv:
YES/NO


gbouv:
Does not work, re-worked in #754

PR #744, opened 2023-06-19
refactor: Remove `PauseService` and `UnpauseService` APIC endpoints

gbouv:
## Description:
Removes PauseService and UnpauseService APIC endpoints. Those endpoints were not used, and were incompatible with k8s anyway. Removing them now will prevent us from using them in the future.

NO


h4ck3rk3y (approved):
Burn it with fire.

PR #749, opened 2023-06-20
feat!: Applying RFC-1123 standard to service names

adschwartz:
## Description:
This is a followup to https://github.com/kurtosis-tech/kurtosis/pull/713. The main change is to apply the RFC-1123 standard to how services are named. This is to ensure compatibility when launching services on k8s.

I've run this on all our packages in `awesome-kurtosis` and all the packages referenced in there and fixed a few packages that had noncompliant naming standard.

Since this is a breaking release we should wait to coordinate with @leeederek.

YES


container-engine-lib/lib/backend_interface/objects/service/service.go
```diff
@@ -8,12 +8,20 @@ import (
...
-	ServiceNameRegex            = "[a-zA-Z0-9-_]+"
-	wordWrappedServiceNameRegex = "^" + ServiceNameRegex + "$"
+	// ServiceNameRegex implements RFC-1123 for naming services, namely:
+	// * contain at most 63 characters
+	// * contain only lowercase alphanumeric characters or '-'
+	// * start with an alphanumeric character
+	// * end with an alphanumeric character
+	// The adoption of RFC-1123 is to maintain compatability with current Kubernetes service and pod naming standards:
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
+	ServiceNameRegex            = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
```
gbouv: I _think_ the following should work:
```
^[a-z0-9][-a-z0-9]{0,61}[a-z0-9]$
```
and it has the advantage of encoding the length constraint inside the regexp itself, so that you don't need to have this separate function `isServiceNameCorrectLength` below. This would be a bit cleaner IMO

core/server/api_container/server/startosis_engine/kurtosis_instruction/add_service/add_service_shared.go
```diff
@@ -85,6 +85,16 @@ func validateSingleService(validatorEnvironment *startosis_validator.ValidatorEn
 	return nil
 }
 
+func invalidServiceNameErrorText(
+	serviceName service.ServiceName,
+) string {
+	return fmt.Sprintf(
+		"Service name '%v' is invalid as it contains disallowed characters. Service names must adhere to the RFC 1123 standard, specifically implementing this regex and be 1-63 characters long: %s. This means the service name must only contain lowercase alphanumeric characters or '-', and must start and end with a lowercase alphanumeric character.",
```
gbouv, line 92: nit: Wondering if it's worth printing the regexp here, especially as you're providing the details already. Maybe hiding the regexp would be better. No strong opinion though

core/server/api_container/server/startosis_engine/kurtosis_instruction/add_service/add_service.go
```diff
@@ -36,7 +36,7 @@ func NewAddService(serviceNetwork service_network.ServiceNetwork, runtimeValueSt
 					IsOptional:        false,
 					ZeroValueProvider: builtin_argument.ZeroValueProvider[starlark.String],
 					Validator: func(value starlark.Value) *startosis_errors.InterpretationError {
-						return builtin_argument.NonEmptyString(value, ServiceNameArgName)
```
gbouv: Do you need to validate it twice here? If the validation was cheap, I'd say it doesn't hurt, but given that you have to write 20 lines of code here, I'd just keep the `NonEmptyString` default validator.
The service names are going to be validated at the validation step anyway

gbouv (approved):

container-engine-lib/lib/backend_interface/objects/service/service.go
adschwartz: @gbouv I struggled with this myself and thought adding the length in the regex was better as well. _But_ the challenge is that with this regex you are requiring a minimum of 2 characters, while the spec says that the name length could be a single character. I thought about just making it so that _we_ would have a minimum of 2 chars, but didn't want to diverge from the kubernetes standard. I felt that it was simpler to just check the length. Let me know if you see it differently but I weighed alignment with kubernetes and RFC 1123 as more important.

gbouv: Oh sorry, I forgot the `?`, I meant:
```
^[a-z0-9]([-a-z0-9]{0,61}[a-z0-9])?$
```
I ran it against all your test cases and this one seems to work

adschwartz: Perfect, I missed the `?` at the end as an option. Thanks for suggesting this!

core/server/api_container/server/startosis_engine/kurtosis_instruction/add_service/add_service.go
adschwartz: @gbouv I'm thinking that it's a better practice to move the validation as early stage as possible, i.e. interpretation stage as it provides the result as soon as possible. But it does have the implication that we then validate twice. If you feel it's a tradeoff worth making I can also go with removing it. Let me know.

gbouv: Generally speaking I'd agree that failing as early as possible is a good default strategy, but if it comes at the cost of adding 20 lines of code, given that it adds no value at all (interpretation is lightening fast), I would personally not do it.

adschwartz: @gbouv Cool, I'm open to remove it. But for the record, I don't agree that it doesn't add any value, as it promotes the best practice of applying validation as early as possible, which can be useful particularly if the validation stage takes a long time to complete. However, there are always tradeoffs to be made and I'm not rigid about this one.

PR #750, opened 2023-06-20
feat: added wait to run_sh task.

Peeeekay:
Added wait field to run_sh task.
The intention is that user can set a wait time-out and if the command is not completed before the time-out, it will fail. In this PR, I have also fixed the wait and ready-condition bug observed on k8s with exec-recipe.


Peeeekay:
This is the output with wait set to 40s and command that takes 50s

Screen Shot 2023-06-20 at 3 16 36 PM

Peeeekay:
Also fixed the wait for exec recipe ( this is an example with ready conditions)

Sample Code:
```
recipe = ExecRecipe(
        command = ["sh", "-c", "sleep 50s"]
    )
    old_pattern = plan.add_service(
        name = "old-pattern",
        config = ServiceConfig(
            image = "ethpandaops/ethereum-genesis-generator:1.0.14",
            entrypoint = ["tail", "-f", "/dev/null"],
            ready_conditions=ReadyCondition(recipe=recipe, field="code", assertion="==", target_value="0", timeout="25s")
        )
    )
```

Output:
Screen Shot 2023-06-20 at 3 18 26 PM

core/server/api_container/server/startosis_engine/kurtosis_instruction/run_sh/run_sh.go
```diff
@@ -21,16 +22,19 @@ import (
...
 	"strings"
+	"time"
 )
 
 const (
 	RunShBuiltinName = "run_sh"
 
 	ImageNameArgName = "image"
 	RunArgName       = "run"
+	StoreFilesName   = "store"
```
laurentluce: Can we rename those to include `Arg` so it is consistent with the ones above?  `StoreFilesArgName` for example.

core/server/api_container/server/startosis_engine/kurtosis_instruction/run_sh/run_sh.go
```diff
@@ -332,13 +364,51 @@ func getCommandToRun(builtin *RunShCapabilities) (string, error) {
...
+	// Wait is set to None
+	if builtin.wait == DisableWaitTimeoutDurationStr {
+		return builtin.serviceNetwork.RunExec(ctx, builtin.name, commandToRun)
+	}
+
+	resultChan := make(chan *exec_result.ExecResult, 1)
+	errChan := make(chan error, 1)
+
+	timoutStr := builtin.wait
+	parsedTimeout, parseErr := time.ParseDuration(timoutStr)
```
laurentluce: Should this validation be done earlier when the arguments are parsed?

Peeeekay: This is extra pre-caution; i could remove it.. because I already do the validation check for wait..
```
Validator: func(value starlark.Value) *startosis_errors.InterpretationError {
						return builtin_argument.DurationOrNone(value, WaitArgName)
					},
```
@laurentluce

laurentluce: If it cannot be invalid then let's remove the second check.

laurentluce (approved):

PR #755, opened 2023-06-21
feat: Implemented rename enclave method in container engine lib

leoporoli:
## Description:
Implemented rename enclave method in container engine lib

NO

This is part of the enclave pool project, more context here: https://www.notion.so/kurtosistech/Enclave-Pool-Project-7273552e53e44b0f931a57568cd170a0


gbouv (approved):
I'm too lazy to dive into the code, or even try it on minikube, but is the namespace the only thing that we tag with the enclave name label? We don't tag the APIC pod for example?

leoporoli:
> I'm too lazy to dive into the code, or even try it on Minikube, but is the namespace the only thing that we tag with the enclave name label? We don't tag the APIC pod for example.

@gbouv good question, these are the enclave's labels and annotations:
image

I think that it will be necessary to update these: `kurtosistech.com/enclave-id, kurtosistech.com/guid, and kurtosistech.com/id` which in fact have the same value

And the APIC pod also contains the enclave-id, so I think that will be necessary to update this too:
image

And the APIC service

image


mmm thinking on it again, I think it's not necessary to update the enclave-uuid it can continue being the same, so won't be necessary to update any other K8s resource if I'm not wrong

PR #694, opened 2023-06-09
feat: Start and Stop service Starlark instructions for Docker

laurentluce:
## Description:
Add `plan.start_service` and `plan.stop_service` Starlark instructions for the Docker backend.  Support for the K8S backend first requires an update to the Gateway to manage the port forwarder connection during stop and start operations.  The k8s backend code itself is ready and the changes (small) are in this PR.

No new backend methods are introduced, the goal being to re-use the current code as much as possible.  All the changes under `api/` and `cli/` are related to the renaming of the API method `StartServices` to `AddServices`.  Most of the code added is unit and integration tests.  The map of service registrations at the service network layer is used as part of the existing start service process so I am using this map to store the service status and config required for the restart process.  I thought about using a map of created `Services` but it is not as clean as leveraging the map of service registrations.

Added some PR comments to help with the review.

Follow-up work in subsequent PRs:

- Update gateway to manage the port forwarder connection during stop and start operations and enable support for k8s.
- Update documentation.

## Is this change user facing?
YES

## References (if applicable):
#673

PR #756, opened 2023-06-22
feat: Start and Stop service Starlark instructions for K8S

### LINKED ISSUE #673: Services should be stoppable and restartable on Docker and K8S
### Background & motivation

There is no way today to just stop a service without completely removing it including its network configuration.  The service then needs to be re-added to start it again.  More details [here](https://www.notion.so/kurtosistech/Starting-and-Stopping-a-Service-b601dc1f60c847fca9cfaf832d424046).

### Desired behaviour

I should be able to stop and start a service using stop and start primitives in Starlark and in the API.

laurentluce:
## Description:
Most of the work for the K8S backend was done as part of #694.  In addition, the port forwarder reconnect feature reduced the amount of work we had to finish.  Here are the changes made in this PR:
- We keep the K8S service untouched when we stop the user service so it can be restarted easily by creating a new pod.  This change does not impact the remove service expected behavior.
- We don't try to create a port forwarder connection if the service is stopped.

A service stopped here is a k8s service with no pod running.  I think it is time to return the service status (registered, started, stopped) as part of the GetServices response so we can simplify the second item above.  I will do that in a follow-up PR.

YES

Closes #673 
#694 
#736


cli/cli/kurtosis_gateway/connection/provider.go
```diff
@@ -93,10 +93,12 @@ func (provider *GatewayConnectionProvider) ForEnclaveApiContainer(enclaveInfo *k
 	return apiContainerConnection, nil
 }
 
-func (provider *GatewayConnectionProvider) ForUserService(enclaveId string, serviceUuid string, servicePortSpecs map[string]*port_spec.PortSpec) (GatewayConnectionToKurtosis, error) {
-	podPortforwardEndpoint, err := provider.getUserServicePodPortforwardEndpoint(enclaveId, serviceUuid)
+func (provider *GatewayConnectionProvider) ForMaybeUserService(enclaveId string, serviceUuid string, servicePortSpecs map[string]*port_spec.PortSpec) (GatewayConnectionToKurtosis, error) {
```
laurentluce: Rename some methods to include `Maybe` to indicate that this might not happen since it does not happen for stopped services.  Added a comment to `writeMaybeOverServiceInfoFieldsWithLocalConnectionInformation` to also clarify things.  We can revert those changes once the service status (registered, stopped, started) is part of the `GetServices` response since we can easily check if we need to call `writeOverServiceInfoFieldsWithLocalConnectionInformation`.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/stop_user_services.go
```diff
@@ -46,22 +45,6 @@ func StopUserServices(
 			}
 		}
 
-		kubernetesService := resources.Service
```
laurentluce, line 49: Keep the k8s service untouched when the user service is stopped so we can restart it easily by attaching a new pod.  This change does not impact the remove service expected behavior.

internal_testsuites/golang/testsuite/startosis_start_service_test/startosis_start_service_test.go
```diff
@@ -42,17 +37,19 @@ def run(plan):
 	plan.add_service(name = DATASTORE_SERVICE_NAME, config = config)
 	plan.print("Service " + DATASTORE_SERVICE_NAME + " deployed successfully.")
 `
-	// We start the service we created through the script above with a different script
-	startScript = `
+	
+	// We stop and restart the service we created through the script above with a different script
+	stopAndStartScript = `
```
laurentluce, line 42: Test with a stop service followed by a start service instruction in the same Starlark script.

cli/cli/kurtosis_gateway/connection/provider.go
gbouv: suuuuper nit: I like the `Maybe` thing, and we've been doing that in a few place already to signify that the result might be `nil`. However, we have a tendency of putting the `Maybe` as a suffix to the function. So here it would be `ForUserServiceMaybe`. Ive also seen `Optional` suffix in a few places. I think the wording doesn't really matter a lot, as long as it is clear, but I kind of like having it as a suffix. Just for consistency, I'd keep it this way. WDYT?

cli/cli/kurtosis_gateway/server/api_container_gateway/api_container_gateway_service_server.go
```diff
@@ -322,9 +326,9 @@ func (service *ApiContainerGatewayServiceServer) writeOverServiceInfoFieldsWithL
 	return nil
 }
 
-// startRunningConnectionForKurtosisService starts a port forwarding process from kernel assigned local ports to the remote service ports specified
+// startMaybeRunningConnectionForKurtosisService starts a port forwarding process from kernel assigned local ports to the remote service ports specified
 // If privatePortsFromApi is empty, an error is thrown
-func (service *ApiContainerGatewayServiceServer) startRunningConnectionForKurtosisService(serviceUuid string, privatePortsFromApi map[string]*kurtosis_core_rpc_api_bindings.Port) (*runningLocalServiceConnection, error) {
+func (service *ApiContainerGatewayServiceServer) startMaybeRunningConnectionForKurtosisService(serviceUuid string, privatePortsFromApi map[string]*kurtosis_core_rpc_api_bindings.Port) (*runningLocalServiceConnection, error) {
```
gbouv: Though my comment above doesn't really apply here. For those methods with long names, the Maybe is a bit unclear, we don't really know to which part of the name it applies. Maybe something like `startRunningConnectionForKurtosisServiceIfRunning` or `*Present`?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/stop_user_services.go
gbouv: Oh interesting, so when you say it doesn't affect the remove service behaviour, you mean you moved this piece of logic to the remove service function? I.e. just want to make sure that when a "kurtosise user service is removed", we don't leave any dangling resource in kubernetes

docs/docs/starlark-reference/plan.md
```diff
@@ -548,12 +541,6 @@ plan.stop_service(
 )
 ```
 
-:::caution
-
-`stop_service` is only available with the Docker backend.
-
-:::
-
```
gbouv, line 567: Nice 💪

gbouv (approved):
Yay!

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/stop_user_services.go
laurentluce: The code I removed set the k8s service selector to nil to not associate itself to a pod anymore.  The k8s service stayed behind after removing the user service.  I did not change that behavior.  The k8s services are cleaned up only when the k8s namespace is deleted during enclave deletion.  I now keep the selector set to be able to associate a new pod.  This change is transparent to the user since the pod is still removed.

cli/cli/kurtosis_gateway/server/api_container_gateway/api_container_gateway_service_server.go
laurentluce: I was trying to be as consistent as possible and I found function names starting with `getMaybe`.  I renamed the updated functions using a suffix and `IfRunning`.

PR #757, opened 2023-06-22
feat: Persist enclave plan in the Starlark executor memory

gbouv:
## Description:
The Starlark executor now keeps track of all instructions that were executed within this enclave. It doesn't take into account the manual modification that might have been made, such as manual file uploads via the CLI.

NO


core/server/api_container/server/startosis_engine/instructions_plan/instructions_plan.go
```diff
@@ -0,0 +1,69 @@
...
+
+	scheduledInstructionUuid := ScheduledInstructionUuid(generatedUuid)
+	scheduledInstruction := NewScheduledInstruction(scheduledInstructionUuid, instruction, returnedValue)
+
+	plan.scheduledInstructionsIndex[scheduledInstructionUuid] = scheduledInstruction
+	plan.instructionsSequence = append(plan.instructionsSequence, scheduledInstructionUuid)
+	return nil
+}
+
+func (plan *InstructionsPlan) AddScheduledInstruction(scheduledInstruction *ScheduledInstruction) *ScheduledInstruction {
```
adschwartz, line 44: Wondering what the difference is between an instruction and scheduled instruction. Might be worth adding a comment about the difference in these functions?

adschwartz (approved):

core/server/api_container/server/startosis_engine/instructions_plan/instructions_plan.go
gbouv: Yeah I had difficulties finding a name for this one. A ScheduledInstruction is basically an Starlark Instruction that is part of a plan (either the current enclavePlan or the plan that just got generated from an interepretation). If you can find a better name for it, I'd gladly take it :D

gbouv: But I'll add a comment in the code for posterity

PR #760, opened 2023-06-22
docs: remove dupe doc reindexing

h4ck3rk3y:
seems like main runs anyway; so why run this?

## Description:
`main` and `tag` both run on release; but main should be enough

NO


.github/workflows/reindex-website-main.yml
```diff
@@ -15,3 +15,27 @@ jobs:
...
+  check-index:
+    runs-on: ubuntu-latest
+    name: Checks that the latest index is still working
+    steps:
+      - uses: fjogeleit/http-request-action@v1
+        id: search
+        with:
+          url: "https://NTSX40VZB8-dsn.algolia.net/1/indexes/kurtosis/query"
+          method: 'POST'
+          customHeaders: '{"X-Algolia-API-Key": "4269c726c2fea4e6cddfeb9a21cd3d4e", "X-Algolia-Application-Id": "NTSX40VZB8"}'
```
gbouv, line 27: Not super familiar with Algolia API, but is this API key and app ID safe to share in our public monorepo? Shouldn't it be secrets like the other params above?

.github/workflows/reindex-website-main.yml
```diff
@@ -15,3 +15,27 @@ jobs:
...
+          data: '{ "params": "query=add_service&hitsPerPage=2&getRankingInfo=1" }'
+      - name: Show Response
+        run: |
+          #!/bin/bash
+          set -eou pipefail
+
+          if (( ${{ fromJson(steps.search.outputs.response).nbHits }} > 100 )); then
```
gbouv, line 34: again, not familiar with Algolia API, but more than 100 hits for `add_service` query? Is that because of all the different versions we have or something else?

.github/workflows/reindex-website-release.yml
```diff
@@ -1,17 +0,0 @@
-name: "Reindex website with Algolia Docsearch for Releases"
-on:
-  push:
-    tags:
-      - '[0-9]+.[0-9]+.[0-9]+'
-jobs:
-  reindex-website-release:
-    runs-on: ubuntu-latest
```
gbouv, line 8: Do we have that because releases have a different version number for example, or something of this kind? Ultimately I trust you on this, just trying to think of things that would have lead us to have a different workflow for releases 🤷

gbouv (approved):

.github/workflows/reindex-website-main.yml
h4ck3rk3y: Yep you can see these on docs.kurtosis.com

.github/workflows/reindex-website-main.yml
h4ck3rk3y: Yep there's about 600 of them thanks to the different versions. Can adjust to something lower if this starts failing

.github/workflows/reindex-website-release.yml
h4ck3rk3y: The contents of the crawl are always the same, the release runs when release happens in addition to main. But main runs when release happens as well so I think its a dupe. I'll merge this tomorrow and test new releaes

PR #762, opened 2023-06-23
feat: stops the task service once it's completed

Peeeekay:
Removes the service when the one time command is completed.


core/server/api_container/server/startosis_engine/kurtosis_instruction/run_sh/run_sh.go
```diff
@@ -263,6 +263,11 @@ func (builtin *RunShCapabilities) Validate(_ *builtin_argument.ArgumentValuesSet
 //   Create an mechanism for other services to retrieve files from the task container
 //   Make task as its own entity instead of currently shown under services
 func (builtin *RunShCapabilities) Execute(ctx context.Context, _ *builtin_argument.ArgumentValuesSet) (string, error) {
+	var err error
+	defer func() {
+		err = removeService(ctx, builtin)
```
h4ck3rk3y, line 301: Should we check if err is not nil? Otherwise something else that returns an error downstream will get overwritten with a nil error.

h4ck3rk3y: Perhaps just log it following the guide instead of changing the returned error

https://github.com/kurtosis-tech/kurtosis-internal-docs/blob/main/style-guide/picking-up-your-toys.md

h4ck3rk3y:
lets merge this after #913 is in

PR #769, opened 2023-06-26
feat: Implement idempotent run v0

gbouv:
## Description:
This feature saves the current state of the enclave after each starlark run and tries to optimize the next execution to execute only the instructions that changed compare to previous run. For this V0, the optimisation is "append-only". I.e. the resolution of instructions will succeed only if what changed is at the end of the instruction queue

NO


core/server/api_container/server/startosis_engine/instructions_plan/resolver/instructions_plan_mask.go
```diff
@@ -0,0 +1,43 @@
...
+	readIdx               int
+	scheduledInstructions []*instructions_plan.ScheduledInstruction
+	isValid               bool
+}
+
+func NewInstructionsPlanMask(size int) *InstructionsPlanMask {
+	return &InstructionsPlanMask{
+		readIdx:               0,
+		scheduledInstructions: make([]*instructions_plan.ScheduledInstruction, size),
+		isValid:               true, // the mask is considered valid until it's proven to be invalid
```
adschwartz, line 15: @gbouv What made choose to assume it's valid (assuming something became simpler if you did that)? I guess my only concern is that if no validation is performed it would be a false positive?

adschwartz (approved):

core/server/api_container/server/startosis_engine/instructions_plan/resolver/instructions_plan_mask.go
gbouv: That's part of the heuristic I explain in the Interpreter code. We tentatively generate plans we think could be valid, then run an interpretation round with this plan to validate it's indeed valid. If it's proven to be invalid, then we continue with another tentative, and so on and so on until we run out of plan, in which case nothing can be cached and we just default to the behaviour we currently have (i.e. run starlark instructions on top of whatever exist in the current enclave).
This is why we initialise isValid to true. I thought the big comment I wrote in the interpreter would clarify this, but maybe it's not clear enough. Let me know if you can think of something more explicit

PR #775, opened 2023-06-26
feat: Add minimal support for feature flags in APIC

gbouv:
## Description:
This adds an `--experimental` flags to `kurtosis run` command to optionally turn on experimental features when running a Starlark package.

YES


cli/cli/commands/run/run.go
```diff
@@ -172,6 +176,14 @@ var StarlarkRunCmd = &engine_consuming_kurtosis_command.EngineConsumingKurtosisC
 			Type:    flags.FlagType_String,
 			Default: mainFunctionNameDefaultValue,
 		},
+		{
+			Key: experimentalFeaturesFlagKey,
```
h4ck3rk3y, line 180: Where do I see a list of features? How does this compare to reading features from a config file?


For instruction caching I will have to pass the flag with every run; which can get annoying but if I write it once to config then my life becomes easier. Thoughts?

cli/cli/commands/service/add/add.go
```diff
@@ -112,6 +112,7 @@ var (
 		portIdSpecDelimiter,
 		serviceAddSpec,
 	)
+	noExperimentalFeature []kurtosis_core_rpc_api_bindings.KurtosisFeatureFlag
```
h4ck3rk3y, line 115: So with just 'add' I can't add the same service twice?

Makes sense I guess. It's just one instruction.

docs/docs/cli-reference/run-starlark.md
```diff
@@ -69,6 +69,7 @@ This command has options available to customize its execution:
 1. The `--verbosity` flag can be used to set the verbosity of the command output. The options include `BRIEF`, `DETAILED`, or `EXECUTABLE`. If unset, this flag defaults to `BRIEF` for a concise and explicit output. Use `DETAILED` to display the exhaustive list of arguments for each command. Meanwhile, `EXECUTABLE` will generate executable Starlark instructions. 
 1. The `--main-file` flag can be used to set the main file filepath, the "main" file is a file for the main method (i.e. the package's entrypoint) which will be executed first; the filepath has to be relative to the package's root. The default value is 'main.star'. This flag is only used for running packages. Example: if your main file is located in a path like this `github.com/my-org/my-package/src/internal/my-file.star` you should set `src/internal/my-file.star` as the relative path.
 1. The `--main-function-name` flag can be used to set the main function name, which will be executed first as the entrypoint of the package or the module. The default value is 'run'.
+1. The `--experimental` flag can be used to enable experimental or incubating features. Please reach out to Kurtosis team if you wish to try any of those.
```
h4ck3rk3y, line 72: Answers my previous question around list of experimental features. It's on an asking basis

h4ck3rk3y (approved):
LGTM. Have a question around UX somewhere but looks alright 😀

cli/cli/commands/run/run.go
gbouv: Yes, that's what I wanted to do initially, but then I realized that it's a decent amount more complex. 1. You have to write it in the config file and make sure we support the config file format. Second it's going to drastically complexify Leandro's feature of enclave pooling, because the feature flag can changes between the time the enclave is created and the time it's allocated by the user.
I am not sure this is something we want to keep around, might just be temporary, so I biased toward the easiest. I think anyway those flags are going to be used mostly by Kurtosians, so we can probably eat this amount of pain for a reasonable amount of time. WDYT?

cli/cli/commands/service/add/add.go
gbouv: Yes for now add will not use instruction caching. Once we remove the feature flag and it becomes broadly available, we will start doing caching here as well

docs/docs/cli-reference/run-starlark.md
gbouv: Yes, I primarily see feature flags as a way for Kurtosians to be beta tester (or maybe users we're very close to), but I don't think at this point we want to advertise all feature flags in our docs for a broad usage

cli/cli/commands/run/run.go
h4ck3rk3y: I think I agree with you here. Let's go with the simplest thing possible; the current implementation and see how it gets received!

PR #777, opened 2023-06-26
chore: added the execute file permission to the Rust build script

leoporoli:
## Description:
Added the execute file permission to the Rust build script and removed an unused file that had the list of the API languages

NO


victorcolombo (approved):
Thanks @leoporoli !

PR #778, opened 2023-06-26
feat: added port print functionality in cli for users to quickly check how to access port.

Peeeekay:
Using this command: `kurtosis port print enclave-id service-id port-id` users can access the information about a port. An example is show below

HERE Is the sample output:

Screen Shot 2023-06-26 at 5 56 04 PM


cli/cli/commands/port/print/print.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	serviceCtx, err := enclaveCtx.GetServiceContext(serviceIdentifier)
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred while getting service context for service with identifier '%v'", serviceIdentifier)
+	}
+
+	publicPorts := serviceCtx.GetPublicPorts()
+
+	if publicPorts[portIdentifier] == nil {
```
leoporoli: nit: you could also check if the `portIdentifier` key is in the map before:
```
publicPort, found := publicPorts[portIdentifier]
if !found {
 return an error
 }
```

leoporoli (approved):

PR #787, opened 2023-06-27
feat: Added enclave pool for improving performance on enclave creation 

leoporoli:
## Description:
- Added enclave pool for improving performance on enclave creation
- The `pool-size` new flag was added in the `engine start and engine restart` commands

## What is pending? 
- reuse idle enclaves when the engine is restarted

YES

[Docs are here](https://www.notion.so/kurtosistech/Enclave-Pool-Project-7273552e53e44b0f931a57568cd170a0?d=2f870450d01d4ad7b8af6d0045602a2a#a20862960dad4868b0b9b6eb320b6178)


docs/docs/cli-reference/engine-start.md
```diff
@@ -13,4 +13,5 @@ This command will do nothing if the Kurtosis engine is already running.
 
 You may optionally pass in the following flags with this command:
 * `--log-level`: The level that the started engine should log at. Options include: `panic`, `fatal`, `error`, `warning`, `info`, `debug`, or `trace`. The engine logs at the `info` level by default.
-* `--version`: The version (Docker tag) of the Kurtosis engine that should be started. If not set, the engine will start up with the default version.
\ No newline at end of file
+* `--version`: The version (Docker tag) of the Kurtosis engine that should be started. If not set, the engine will start up with the default version.
+* `--pool-size`: The size of the Kurtosis engine enclave pool. The enclave pool is a component of the Kurtosis engine that allows us to create and maintain 'n' number of idle enclaves for future use. This functionality allows to improve the performance for each new creation enclave request.
```
gbouv: Should we add that this is only available for Kubernetes backend? BTW what happens if I do `kurtosis engine start --pool-size 2` while using a Docker backend?

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -0,0 +1,314 @@
...
+	//TODO the reuse logic is not enable yet because we ned to store the APIC version on the APIContainer object in container-engine-lib
+	//TODO in order to using it for comparing it with the expected version
+
+	// iterate on all the existing enclaves in order to find idle enclaves already created
+	// and reuse or destroy them if these were created from old Kurtosis version
+	if err := enclavePool.destroyPreviousIdleEnclaves(); err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred destroying previous idle enclave before creating the enclave pool")
+	}
+
+	go enclavePool.run()
```
gbouv: Will that automatically be killed if we stop the engine? Or do we need a close function of some sort we can call when stopping the engine?
This would also have the advantage of killing the idle enclaves when we stop the engine, not when we start a new one

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -0,0 +1,314 @@
...
+			if fails >= maxRetryAfterFails {
+				logrus.Debugf("The enclave pool sub-routine has been canceled after reaching the max retries value '%v'", maxRetryAfterFails)
+				break
+			}
+			logrus.Debugf("Next retry will be executed on '%v' seconds", timeBetweenRetries.Seconds())
+			time.Sleep(timeBetweenRetries)
+		}
+		enclaveUUID := enclave.EnclaveUUID(newEnclaveInfo.EnclaveUuid)
+
+		pool.idleEnclavesChan <- enclaveUUID
```
gbouv: So, you start by creating the enclave, and THEN you push it to the channel?? Does this work in practice? Won't you end up with one more enclave running than the desired size of the pool?

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -0,0 +1,314 @@
...
+	if !found {
+		return nil, stacktrace.NewError("Expected to find an enclave with UUID '%v' in enclave map '%+v', but none was found; this is a bug in Kurtosis", enclaveUUID, enclaves)
+	}
+
+	return enclaveObj, nil
+}
+
+// The enclave pool feature is only available for Kubernetes so far, and it will be activated
+// only if users require this when setting the pool-size value
+func isEnclavePoolAllowedForThisConfig(
```
gbouv: Where is this used?

engine/server/engine/enclave_manager/enclave_pool.go
gbouv: Ah yes that's the comment you wrote in CreateEnclavePool function. I understand that the enclave UUID can be retrieved only AFTER the enclave is created, therefore you have to create the enclave first and only after push it to the channel. Can't we use a distinct channel for the "pooling" mechanism, and store the enclave UUIDs in a simple array for example?
IDK, this seems like an extra complexity here

gbouv (approved):
Looks generally good, not a huge fan of the "create enclave first and then push it to the channel" logic though, I'd prefer the "push to channel" and "enclave creation" operation to happen as closely as possible

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: yes, it's pending to implement a `StopEngine` engine endpoint for being used when users execute `engine stop` and the killing idle enclave process in the start if for when the engine crashes and engine restarts (it's pending to implement the reuse idle enclave mechanism also)

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: @gbouv I'm not really sure if I understand your concern. What error do you think that could happen with the current state?

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: @gbouv It's used in the `EnclaveManager` here:
image
but now we a second though I think that I'll move these validations inside the `CreateEnclavePool` method and I'll add the validations in the CLI also

docs/docs/cli-reference/engine-start.md
leoporoli: good catch! I added the CAUTION on the docs and a validation on the CLI

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: it was refactored, now the channel length is equal to the pool size

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: it was refactored as I described in the previous comment

PR #788, opened 2023-06-28
feat: make the docker network attachable

h4ck3rk3y:
## Description:
Sets the docker network to be attachable to allow users to attach local services to the services inside the enclave

YES

https://discord.com/channels/783719264308953108/783719264308953111/1123504466340098078

### DIFF
--- a/container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
+++ b/container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
@@ -113,7 +113,6 @@ const (
 
 	emptyNetworkAlias = ""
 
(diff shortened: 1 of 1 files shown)


container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -112,6 +112,9 @@ const (
 	successfulExitCode = 0
 
 	emptyNetworkAlias = ""
+
+	// setting this to true to enable "satrian" on Discord to do his workflow
```
adschwartz: I think we should leave this comment out of the code base, but other wise good with the change

adschwartz (approved):

PR #798, opened 2023-06-29
chore: stop the enclave pool, which includes removing idle enclaves, when the engine is stopped

leoporoli:
## Description:
 stop the enclave pool, which includes removing idle enclaves, when the engine is stopped

NO

[Docs are here](https://www.notion.so/kurtosistech/Enclave-Pool-Project-7273552e53e44b0f931a57568cd170a0?d=2f870450d01d4ad7b8af6d0045602a2a#a20862960dad4868b0b9b6eb320b6178)

Related to #787

### DIFF
--- a/engine/server/engine/enclave_manager/enclave_manager.go
+++ b/engine/server/engine/enclave_manager/enclave_manager.go
@@ -473,6 +473,13 @@ func (manager *EnclaveManager) getEnclavesWithoutMutex(
 
 }
 
+func (manager *EnclaveManager) Close() error {
+	if err := manager.enclavePool.Close(); err != nil {
+		return stacktrace.Propagate(err, "An error occurred closing the enclave pool")
+	}
+	return nil
+}
+
 func getEnclaveByEnclaveIdFilter(enclaveUuid enclave.EnclaveUUID) *enclave.EnclaveFilters {
 	return &enclave.EnclaveFilters{
 		UUIDs: map[enclave.EnclaveUUID]bool{
--- a/engine/server/engine/enclave_manager/enclave_pool.go
+++ b/engine/server/engine/enclave_manager/enclave_pool.go
@@ -35,8 +35,9 @@ type EnclavePool struct {
 }
 
 // CreateEnclavePool will do the following:
-// 1- Wil create a new enclave pool object, if pool size > 1, return nil if pool size = 0, or return an error
-// 2- Will remove idle enclaves from previous engine runs
+// 1- Will remove idle enclaves from previous engine runs even if the pool is not activated (this is for removing
+// any resource leak after an engine restar without this feature enabled or after an engine crash)
+// 2- Wil create a new enclave pool object, if pool size > 1, return nil if pool size = 0, or return an error
 // 3- Will start a subroutine in charge of filling the pool
 func CreateEnclavePool(
 	kurtosisBackend backend_interface.KurtosisBackend,
@@ -45,6 +46,16 @@ func CreateEnclavePool(
 	poolSize uint8,
 	engineVersion string,
 ) (*EnclavePool, error) {
+
+	//TODO the current implementation only removes the previous idle enclave, it's pending to implement the reusable feature
+	//TODO the reuse logic is not enable yet because we ned to store the APIC version on the APIContainer object in container-engine-lib
+	//TODO in order to using it for comparing it with the expected version
+
+	// iterate on all the existing enclaves in order to find idle enclaves already created
+	// and reuse or destroy them if these were created from old Kurtosis version
+	if err := destroyIdleEnclaves(kurtosisBackend); err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred destroying previous idle enclave before creating the enclave pool")
+	}
 
 	// validations
 	// poolSize = 0 means that the Enclave Pool won't be activated, it returns nil with no error
@@ -78,16 +89,6 @@ func CreateEnclavePool(
 		fillChan:                fillChan,
 		engineVersion:           engineVersion,
 		cancelSubRoutineCtxFunc: cancelCtxFunc,
-	}
-
-	//TODO the current implementation only removes the previous idle enclave, it's pending to implement the reusable feature
-	//TODO the reuse logic is not enable yet because we ned to store the APIC version on the APIContainer object in container-engine-lib
-	//TODO in order to using it for comparing it with the expected version
-
-	// iterate on all the existing enclaves in order to find idle enclaves already created
-	// and reuse or destroy them if these were created from old Kurtosis version
-	if err := enclavePool.destroyIdleEnclaves(); err != nil {
-		return nil, stacktrace.Propagate(err, "An error occurred destroying previous idle enclave before creating the enclave pool")
 	}
 
 	go enclavePool.run(ctxWithCancel)
@@ -150,7 +151,7 @@ func (pool *EnclavePool) GetEnclave(
 			idleEnclavesToRemove := map[enclave.EnclaveUUID]bool{
 				enclaveUUID: true,
 			}
-			if err := pool.destroyEnclaves(ctx, idleEnclavesToRemove); err != nil {
+			if err := destroyEnclaves(ctx, pool.kurtosisBackend, idleEnclavesToRemove); err != nil {
 				logrus.Errorf(
 					"Something fail while getting an enclave from the pool, we tried to destroy the "+
 						"enclave that was taken from it, for avoiding a resource leak, but this also fails, "+
@@ -192,7 +193,7 @@ func (pool *EnclavePool) Close() error {
 	pool.cancelSubRoutineCtxFunc()
 
 	// destroy all the idle enclaves
-	if err := pool.destroyIdleEnclaves(); err != nil {
+	if err := destroyIdleEnclaves(pool.kurtosisBackend); err != nil {
 		return stacktrace.Propagate(err, "An error occurred destroying idle enclave")
 	}
 
@@ -208,64 +209,6 @@ func (pool *EnclavePool) init(poolSize uint8) {
 	for i := uint8(0); i < poolSize; i++ {
 		pool.fillChan <- fill
 	}
-}
-
-func (pool *EnclavePool) destroyIdleEnclaves() error {
-	ctx := context.Background()
-
-	filters := &enclave.EnclaveFilters{
-		UUIDs: map[enclave.EnclaveUUID]bool{},
-		Statuses: map[enclave.EnclaveStatus]bool{
-			enclave.EnclaveStatus_Running: true,
-		},
-	}
-
-	enclaves, err := pool.kurtosisBackend.GetEnclaves(ctx, filters)
-	if err != nil {
-		return stacktrace.Propagate(err, "An error occurred getting enclaves using filters '%+v'", filters)
-	}
-
-	idleEnclavesToRemove := map[enclave.EnclaveUUID]bool{}
-
-	for enclaveUUID, enclaveObj := range enclaves {
-		enclaveName := enclaveObj.GetName()
-		// is it an idle enclave from a previous run?
-		if strings.HasPrefix(enclaveName, idleEnclaveNamePrefix) {
-			idleEnclavesToRemove[enclaveUUID] = true
-		}
-	}
-
-	if err := pool.destroyEnclaves(ctx, idleEnclavesToRemove); err != nil {
-		return stacktrace.Propagate(err, "An error occurred destroying enclaves with UUIDs '%v'", idleEnclavesToRemove)
-	}
-
-	return nil
-}
-
-func (pool *EnclavePool) destroyEnclaves(ctx context.Context, enclavesToRemove map[enclave.EnclaveUUID]bool) error {
-	destroyEnclaveFilters := &enclave.EnclaveFilters{
-		UUIDs:    enclavesToRemove,
-		Statuses: map[enclave.EnclaveStatus]bool{},
-	}
-
-	_, destroyEnclaveErrs, err := pool.kurtosisBackend.DestroyEnclaves(ctx, destroyEnclaveFilters)
-	if err != nil {
-		return stacktrace.Propagate(err, "An error occurred destroying enclaves using filters '%v'", destroyEnclaveFilters)
-	}
-	if len(destroyEnclaveErrs) > 0 {
-		logrus.Errorf("Errors occurred removing the following enclaves...")
-		var removalErrorStrings []string
-		for enclaveUUID, destroyEnclaveErr := range destroyEnclaveErrs {
-			logrus.Errorf("Enclave wit UUID '%v' error:\n %v", enclaveUUID, destroyEnclaveErr.Error())
-			resultErrStr := fmt.Sprintf(">>>>>>>>>>>>>>>>> ERROR on Enclave %v <<<<<<<<<<<<<<<<<\n%v", enclaveUUID, destroyEnclaveErr.Error())
-			removalErrorStrings = append(removalErrorStrings, resultErrStr)
-		}
-		logrus.Errorf("...you should have to manually remove all these errored enclaves.")
-		errorSeparator := "\n"
-		joinedRemovalErrors := strings.Join(removalErrorStrings, errorSeparator)
-		return stacktrace.NewError("Following errors occurred while removing idle enclaves :\n%v", joinedRemovalErrors)
-	}
-	return nil
 }
 
 // run is executed in a subroutine and wait for any of these two signals:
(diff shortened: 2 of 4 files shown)


engine/server/engine/enclave_manager/enclave_manager.go
```diff
@@ -473,6 +473,13 @@ func (manager *EnclaveManager) getEnclavesWithoutMutex(
 
 }
 
+func (manager *EnclaveManager) Stop() error {
```
gbouv: nit: I've seen `Close` more often that `Stop`. I think in Go, the `Close()` method makes it implement the `Closer` interface, which seem to indicate it's more common to use Close.
If you feel `Stop` is better, let's go for `Stop`, no super strong opinion TBH

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -205,27 +226,33 @@ func (pool *EnclavePool) destroyPreviousIdleEnclaves() error {
...
-		newEnclaveInfo, err := pool.createNewEnclave()
-		if err != nil {
-			// Retry strategy if something fails
-			logrus.Errorf("An error occurred creating a new idle enclave. Error\n%v", err)
-			fails++
-			if fails >= maxRetryAfterFails {
-				logrus.Debugf("The enclave pool sub-routine has been canceled after reaching the max retries value '%v'", maxRetryAfterFails)
-				break
+		select {
+		case <-pool.closeChan:
```
gbouv: Does this really work??? O.o 
Below, the call `pool.idleEnclavesChan <- enclaveUUID` should be blocking right? So, if the code is blocked at this line when a value is pushed to `closeChan`, there is no reason why we would end up in this `case`. Am I misunderstanding something here?

engine/server/engine/main.go
```diff
@@ -125,6 +125,12 @@ func runMain() error {
 		serverArgs.DidUserAcceptSendingMetrics,
 		logsDatabaseClient,
 	)
+	defer func() error {
+		if err := engineServerService.Stop(); err != nil {
+			return stacktrace.Propagate(err, "An error occurred stopping the engine server service")
```
gbouv: I am not entirely sure what throwing inside a deferred function does, have you tested it? Usually when closing a resource does not work, we just log the error saying "hey, we did our best to clean our stuffs up but for some reasons it failed, you might need to take manual actions to clean things up yourself".
That's really the best we can do, I don't think throwing is a good thing here, I am actually wondering whether we would see this error in the logs

engine/server/engine/server/engine_server_service.go
```diff
@@ -216,6 +216,13 @@ func (service *EngineServerService) GetServiceLogs(
 
 }
 
+func (service *EngineServerService) Stop() error {
```
gbouv: Same here about Stop VS Close

gbouv (commented):
I'm a bit dubious about the pattern you're using for the `closeChan`, see my comment below. Playing the code in my head, I think there as cases where it might not work

engine/server/engine/enclave_manager/enclave_manager.go
leoporoli: I'll use `Close` at the beginning I used `Close` but then I had some doubts, but now that you are also mentioning it, I'll rollback it

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: I've changed the logic

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -62,17 +62,28 @@ func CreateEnclavePool(
 		)
 	}
 
-	// The amount of idle enclaves is equal to the chan capacity + one enclave (this one will be waiting in the queue until the channel is unblocked)
-	// chanCapacity = 0  means that it will be an unbuffered, so the communication between the channels will succeed when both (sender and receiver) are ready
-	chanCapacity := poolSize - oneEnclave
+	// this channel is the repository of idle enclave UUIDs
```
gbouv, line 73: Do we need a channel for a simple "repository", as you say? Especially a buffered array, those are complex to reason about (need to think of when a call will be blocking, when it's not, etc), it adds complexity to the code and if we don't need such a structure, let's use something simpler. 
What about a simple list? You can add an element with a simple `append`, and to remove one, just do something like:
```
if len(idleEnclavesUuids) > 0 {
    enclaveUuid = idleEnclavesReservoir[0]
    idleEnclavesUuids = idleEnclavesUuids[1:]
    return enclaveUuid
}
```

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -62,17 +62,28 @@ func CreateEnclavePool(
...
+	idleEnclavesChan := make(chan enclave.EnclaveUUID, poolSize)
+
+	// this channel is for communicate to the sub-routine that one
+	// idle enclave has been picked from the pool, so it can trigger the
+	// fill mechanism again, it has the capacity = poolSize for not blocking
+	// the caller
+	fillChan := make(chan bool, poolSize)
```
gbouv, line 79: yes here it makes sense to use a channel. However i don't understand why it needs to be buffered. If it's indeed used a signal to send to sub-routine that it needs to start a new idle enclave. I don't think this should ever get more than one value, unless i'm missing something

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -62,17 +62,28 @@ func CreateEnclavePool(
 		)
 	}
 
-	// The amount of idle enclaves is equal to the chan capacity + one enclave (this one will be waiting in the queue until the channel is unblocked)
-	// chanCapacity = 0  means that it will be an unbuffered, so the communication between the channels will succeed when both (sender and receiver) are ready
-	chanCapacity := poolSize - oneEnclave
+	// this channel is the repository of idle enclave UUIDs
+	idleEnclavesChan := make(chan enclave.EnclaveUUID, poolSize)
+
+	// this channel is for communicate to the sub-routine that one
```
gbouv: Improve comment to make it more clear here. Maybe something like:
```
This channel is used as a signal to tell to the sub-routine that one idle enclave has been allocated from the pool ....
```

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -121,11 +135,18 @@ func (pool *EnclavePool) GetEnclave(
 		return nil, nil
 	}
 
+	// If there is no idle enclave in the pool returns nil
+	// for not to block the caller
```
gbouv, line 134: Exactly, here you have some overhead because you're using a channel, I think a list would be enough here

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -121,11 +135,18 @@ func (pool *EnclavePool) GetEnclave(
...
+	if len(pool.idleEnclavesChan) < oneIdleEnclave {
+		return nil, nil
+	}
 	enclaveUUID, ok := <-pool.idleEnclavesChan
 	if !ok {
 		return nil, stacktrace.NewError("A new enclave can't be returned from the pool because the internal channel is closed, it shouldn't happen; this is a bug in Kurtosis")
 	}
 
+	pool.fillChan <- fill
```
gbouv, line 144: nit: Maybe add a comment here explaining why you need to do that. It's not super explicit by just reading the code

gbouv: Yes so here, you should be expecting that as soon as you push a value to this channel, it get picked by the subroutine to start a new idle enclave, right? So, if it's picked immediately, you don't need a buffered channel

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -147,12 +168,30 @@ func (pool *EnclavePool) GetEnclave(
...
 
+// Close stop the EnclavePool sub-routine in charge of filling the pool
+// and removes all the idle enclaves already created
+func (pool *EnclavePool) Close() error {
+
+	// will terminate running processes in the sub-routine
+	pool.cancelCtxFunc()
+
+	// closing this channel will quit the sub-routine
+	close(pool.closeChan)
```
gbouv: Where are you closing the other channels? That's also why it's an overhead to use a channel for the enclaveUuids. You need to make sure it's closed, etc.

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -203,34 +242,58 @@ func (pool *EnclavePool) destroyPreviousIdleEnclaves() error {
 }
 
 // run method is in charge of filling the enclave pool channel with new idle enclaves
-func (pool *EnclavePool) run() {
+func (pool *EnclavePool) run(ctx context.Context) {
 
-	defer close(pool.idleEnclavesChan)
+	defer func() {
+		close(pool.idleEnclavesChan)
```
gbouv: Don't you need to close the `fillChan` at some point also?

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -62,17 +62,28 @@ func CreateEnclavePool(
...
+	// idle enclave has been picked from the pool, so it can trigger the
+	// fill mechanism again, it has the capacity = poolSize for not blocking
+	// the caller
+	fillChan := make(chan bool, poolSize)
 
-	idleEnclavesChan := make(chan enclave.EnclaveUUID, chanCapacity)
+	// this channel if for signaling the close event to the sub-routine
+	closeChan := make(chan bool)
+
+	ctxWithCancel, cancelCtxFunc := context.WithCancel(context.Background())
```
gbouv, line 81: Using a cancel context is clever here, but then it raises the question of why you need the closeChan, aren't those two redundant?

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -203,34 +242,58 @@ func (pool *EnclavePool) destroyPreviousIdleEnclaves() error {
...
+				if err == context.Canceled {
+					logrus.Debug("The sub-routine context has been canceled")
+				} else {
+					logrus.Errorf("An error occurred filling the enclave pool. Error\n%v", err)
+				}
 				break
 			}
-			logrus.Debugf("Next retry will be executed on '%v' seconds", timeBetweenRetries.Seconds())
-			time.Sleep(timeBetweenRetries)
+		case <-pool.closeChan:
```
gbouv: Can't you pass in the cancelCtx and replace this with `case <-cancelCtx.Done()` ? This way you don't even need the closeChan, as soon as you call the `cancelFunc` in your Close() function, it will send a signal here

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -203,34 +242,58 @@ func (pool *EnclavePool) destroyPreviousIdleEnclaves() error {
...
 	for {
-		newEnclaveInfo, err := pool.createNewEnclave()
-		if err != nil {
-			// Retry strategy if something fails
-			logrus.Errorf("An error occurred creating a new idle enclave. Error\n%v", err)
-			fails++
-			if fails >= maxRetryAfterFails {
-				logrus.Debugf("The enclave pool sub-routine has been canceled after reaching the max retries value '%v'", maxRetryAfterFails)
+		// wait until receive the re-fill signal or the close signal
+		select {
```
gbouv, line 219: YES, now your `for ... select` loop makes a lot more sense, it waits on both channels and the first one to receive a value enters the clause. That's much better

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -203,34 +242,58 @@ func (pool *EnclavePool) destroyPreviousIdleEnclaves() error {
...
+		case <-pool.closeChan:
+			return
+		}
+	}
+}
+
+func (pool *EnclavePool) fill(ctx context.Context) error {
+
+	for {
+		if len(pool.idleEnclavesChan) == cap(pool.idleEnclavesChan) {
```
gbouv: Huh... You'll get some concurrency issues doing that here. What happens if an enclave get picked WHILE it's being filled? With this for loop checking on the channel length, you're basically defeating the purpose of using channels, aren't you?

I think another approach here could be to have this function create just ONE enclave, and then you use a fillChan buffered channel that you start by filling entirely. This way, when you start the subroutine, it will start by calling X number of time this function to empty the fillChan completely, and continue from there. If you do that, then it makes total sense to use a buffered channel for fillChan

gbouv (commented):
I'm still dubious this is completely bug-proof 😬 

A couple of suggestions:
- First, get rid of the closeChan. You're already using a cancelCtx (which is clever) and you can use cancelCtx.Done() signal to replace your closeChan
- Second, right now you don't need fillChan to be buffered. And the reason you don't need that is because your `fill` function is creating as many enclaves as needs to fill the pool _every time_ it's being called. However, I think this exposes us to weird concurrency issues. I'd much prefer to indeed use a buffered channel for fillChan, BUT have your current `fill` method create only one enclave at a time. You can send start by pushing as many value as needed to the `fillChan` to effectively fill the pool, and then once the pool is filled then values will be pushed one a a time to fillChan, but this is fine.
- Lastly, don't use a channel to track enclaveUuids if you don't need one, and I think in this case you don't need one

Happy to chat more if it's unclear. I can also push a commit to help clarify my ideas

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: yes, good catch!

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: yes they are, good catch!

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: replaced! good catch!

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: refactored! good suggestion!

leoporoli: in this version, the race condition was handled by the Select statement in the run method, because it will be executing one call to fill at a time, but your suggestion I think is more reasonable and legible, so I'll do the refactor.

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: Added the close

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: the buffered channel is for if we have concurrent requests

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: it's buffered for handling concurrent requests and not blocking any of them.

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: I think that we need a channel because the `get` and `fill` methods are executed in different routines, if we use a list we could have race conditions when we append a new element from both sides. I think that a channel is better in this case, but we can discuss it, probably I could missing something. I even tried to add a FIFO queue with a list but then I realized that I should add a mutex or something to avoid race conditions which will block the client from getting an enclave from the pool.

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: I put my comment about this above, we can discuss it

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -47,49 +47,60 @@ func CreateEnclavePool(
 	engineVersion string,
 ) (*EnclavePool, error) {
 
+	//TODO the current implementation only removes the previous idle enclave, it's pending to implement the reusable feature
+	//TODO the reuse logic is not enable yet because we ned to store the APIC version on the APIContainer object in container-engine-lib
+	//TODO in order to using it for comparing it with the expected version
+
+	// iterate on all the existing enclaves in order to find idle enclaves already created
+	// and reuse or destroy them if these were created from old Kurtosis version
+	if err := destroyIdleEnclaves(kurtosisBackend); err != nil {
```
leoporoli, line 54: I moved this before this condition `if poolSize == 0 {` because it has to be executed always

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -144,93 +178,82 @@ func (pool *EnclavePool) GetEnclave(
...
+// ====================================================================================================
+//
+//	Private helper methods
+//
+// ====================================================================================================
+func (pool *EnclavePool) init(poolSize uint8) {
+	for i := uint8(0); i < poolSize; i++ {
+		pool.fillChan <- fill
+	}
+}
```
gbouv, line 210: Nice nice !

engine/server/engine/enclave_manager/enclave_pool.go
gbouv: Got it got it, yeah using a channel might be easier than a list and mutex. I'm in sync with you here

engine/server/engine/enclave_manager/enclave_pool.go
```diff
@@ -121,10 +132,33 @@ func (pool *EnclavePool) GetEnclave(
 		return nil, nil
 	}
 
+	// If there is no idle enclave in the pool returns nil
+	// for not to block the caller
+	if len(pool.idleEnclavesChan) < oneIdleEnclave {
```
gbouv: nit: I think you can do `== 0` here, using a const is a bit overkill IMO

gbouv: Returning `nil, nil` usually frightens me. Are we 100% certain this won't cause an NPE downstream? And BTW what will happen in this case? Are we creating an encalve on the fly?

gbouv (approved):
It feels a lot more natural reading this code. Thanks for making it a lot simpler to understand 👍 This looks very good to me now 💪

engine/server/engine/enclave_manager/enclave_pool.go
leoporoli: I am also not too happy with returning `nil, nil` but after reviewing it several times I didn't find a better cleaner way to solve this use case (one alternative was returning another value for specifying that not enclave was found in the pool, but I though it wasn't to clean), I added a comment about it on the method for this reason. And yes, it's actually handled from the caller's side

image

Let me know if you have any other suggestion I could refactor it in a next PR

PR #807, opened 2023-07-03
chore: filter idle enclaves on `EnclaveManager`

leoporoli:
## Description:
filter idle enclaves on `EnclaveManager` 
This work is part of the enclave pool project, and we need to filter idle enclaves for instance when users get the enclaves from the engine or when users call to the clean endpoint

NO


leoporoli:
@gbouv it's failing when the CI runs the `kurtosis cluster set minikube`  so you can omit the review until I fix this.  I'll review it and fix it tomorrow, so I commented on the code for debugging purposes

if you have a couple of free minutes and you are curious about the error, you can see the logs here: https://app.circleci.com/pipelines/github/kurtosis-tech/kurtosis/4399/workflows/19a7ca1e-5c89-4639-86a2-2fb7388ba502/jobs/55444  I think that could be related to the wait-for-port-using-netstat timeout

gbouv (approved):

gbouv:
Wow, this is a weird error, never seen that one before. It seems the engine pod was never created

PR #816, opened 2023-07-05
fix: fixed the output for port print

Peeeekay:
it just returns the url associated with port id.


cli/cli/commands/port/print/print.go
```diff
@@ -108,7 +108,7 @@ func run(
 	publicPort, found := publicPorts[portIdentifier]
 	if !found {
 		return stacktrace.NewError(
-			fmt.Sprintf("Port Identifier: %v is not found for service: %v in enclave %v", portIdentifier, serviceIdentifier, enclaveIdentifier),
+			fmt.Sprintf("Port Identifier: %q is not found for service: %q in enclave %q", portIdentifier, serviceIdentifier, enclaveIdentifier),
```
h4ck3rk3y: hmm, what about `fmt.Errorf`? Should we put the values in `''`

h4ck3rk3y (approved):

cli/cli/commands/port/print/print.go
Peeeekay: %q should do that for us.

PR #817, opened 2023-07-05
chore: pass the name of the package instead of the path when using a local package

### LINKED ISSUE #811: Parse the Kurtosis.yml to send the actual package id of a package run using .
### Background & motivation

We get hashed values for `.` or `./` while we could just hash the package locator present in the `kurtosis.yml`

### Desired behaviour

We hash the `kurotsis.yml` package locator

h4ck3rk3y:
## Description:
We would send `.` or `./` or others while we could send hashed `packageNames`. We're doing that now

NO


gbouv (approved):
Nice nice

PR #833, opened 2023-07-06
feat: exposing kurtosis frontend v0

Peeeekay:
This PR just exposes the current to the user:
1. I have frontend directory in engine/fronted/...
2. the build artifacts generated by the frontend using npm build are stored in engine/server/engine/dist
3. these are copied over to engine dockerfile, and are exposed on port 9711..

The frontend build artifacts step is manual at the moment; will include this in build process in upcoming PR(s). 
This is only working with docker backend at the moment; will make it work with k8s in upcoming PRs .

This is not backward compatible; that is frontend would not work with enclaves created before 0.80.8


engine/server/webapp/index.html
```diff
@@ -0,0 +1 @@
+<!doctype html><html lang="en"><head><meta charset="utf-8"/><link rel="icon" href="/favicon.ico"/><meta name="viewport" content="width=device-width,initial-scale=1"/><meta name="theme-color" content="#000000"/><meta name="description" content="Web site created using create-react-app"/><link rel="apple-touch-icon" href="/logo192.png"/><link rel="manifest" href="/manifest.json"/><title>React App</title><script defer="defer" src="/static/js/main.5ae6b736.js"></script><link href="/static/css/main.7b6afa70.css" rel="stylesheet"></head><body><noscript>You need to enable JavaScript to run this app.</noscript><div id="root"></div></body></html>
```
adschwartz: may want to scrub the create-react-app stuff

adschwartz:
is there a reason you chose javascript over typescript? I'd think we would prefer ts unless there's a strong reason why not?

Peeeekay:
> is there a reason you chose javascript over typescript? I'd think we would prefer ts unless there's a strong reason why not?

The frontend code is pretty simple, and adding typescript will be just an overkill imo. I would def. need to refactor some of the js code to make it more maintainable at that time, I can use typescript.
 For initial versions, I believe javascript is fine as in my experience typescript is very good for backend but for most frontends ( generally  aren't super complex,  mostly async requests and state management for those cases, javascript works well.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -18,6 +18,7 @@ import (
 )
 
 const (
+	frontendPortSpec                            = 9711
```
leoporoli, line 22: This should be an argument of the CreateEngine method like the `grpcPortNum` in order to centralize this number in a place as we did for the grpcPortNum

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -55,6 +56,16 @@ func CreateEngine(
 		)
 	}
 
+	httpPortSpec, err := port_spec.NewPortSpec(uint16(frontendPortSpec), consts.EngineTransportProtocol, consts.HttpApplicationProtocol, defaultWait)
+	if err != nil {
+		return nil, stacktrace.Propagate(
+			err,
+			"An error occurred creating the engine's http port spec object using number '%v' and protocol '%v'",
+			9711,
```
leoporoli: replace the number with the new argument

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -74,8 +85,14 @@ func CreateEngine(
 		return nil, stacktrace.Propagate(err, "An error occurred transforming the private grpc port spec to a Docker port")
 	}
 
+	httpDockerPort, err := shared_helpers.TransformPortSpecToDockerPort(httpPortSpec)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred transforming the private grpc port spec to a Docker port")
```
leoporoli: this error msg should be updated to match the port name

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -55,6 +56,16 @@ func CreateEngine(
...
+			err,
+			"An error occurred creating the engine's http port spec object using number '%v' and protocol '%v'",
+			9711,
+			consts.EngineTransportProtocol.String(),
+		)
+	}
+
 	engineAttrs, err := objAttrsProvider.ForEngineServer(
 		engineGuid,
 		consts.KurtosisInternalContainerGrpcPortId,
```
leoporoli, line 72: you should add the `httpPortSpec` also here in order to store this number in Docker labels as we do for `privateGrpcPortSpec `

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -74,8 +85,14 @@ func CreateEngine(
...
 	}
 
+	httpDockerPort, err := shared_helpers.TransformPortSpecToDockerPort(httpPortSpec)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred transforming the private grpc port spec to a Docker port")
+	}
+
 	usedPorts := map[nat.Port]docker_manager.PortPublishSpec{
 		privateGrpcDockerPort: docker_manager.NewManualPublishingSpec(grpcPortNum),
+		httpDockerPort:        docker_manager.NewManualPublishingSpec(uint16(9711)),
```
leoporoli: replace the number with the new argument

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -74,8 +85,14 @@ func CreateEngine(
...
 	}
```
leoporoli, line 98: I think that we should also wait for this new port to become available before returning the engine object. You could even do it in parallel with the other wait or refactor the `WaitForPortAvailabilityUsingNetstat` in order to accept more than one portSpec

engine/frontend/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "frontend",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
+    "@testing-library/jest-dom": "^5.16.5",
```
leoporoli, line 7: are all these dependencies necessary for prod? just checking

engine/frontend/public/index.html
```diff
@@ -0,0 +1,39 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8" />
+    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
```
leoporoli: Do we have this icon? otherwise, it will be a broken link

engine/frontend/public/index.html
```diff
@@ -0,0 +1,39 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#000000" />
+    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
```
leoporoli: Same here. Do we have this image? otherwise, it will be a broken link

engine/frontend/public/index.html
```diff
@@ -0,0 +1,39 @@
...
+    <!--
+      manifest.json provides metadata used when your web app is installed on a
+      user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
+    -->
+    <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
```
leoporoli: Same here. Do we have this manifest.json file? otherwise, it will be a broken link

engine/frontend/src/App.css
```diff
@@ -0,0 +1,38 @@
+.App {
```
leoporoli, line 1: could we use lowercase for the filename?

engine/frontend/src/App.js
```diff
@@ -0,0 +1,24 @@
+import './App.css';
```
leoporoli, line 1: Same here for the filename, all our files are lowercase

engine/frontend/src/App.test.js
```diff
@@ -0,0 +1,8 @@
+import { render, screen } from '@testing-library/react';
```
leoporoli, line 1: Same here for the filename in lowercase

engine/server/engine/main.go
```diff
@@ -131,6 +134,40 @@ func runMain() error {
 		},
 	)
 
+	go func() {
+		staticPath := "/run/webapp"
```
leoporoli: `staticPath` and i`ndexPath` should be`serverArgs` fields instead

engine/frontend/package.json
Peeeekay: probably not; but can remove the un-needed ones in upcoming PR(s). I just kept testing ones as they may get handy for me to add some tests later on

engine/server/engine/main.go
```diff
@@ -42,6 +44,7 @@ const (
 	logMethodAlongWithLogLine = true
 	functionPathSeparator     = "."
 	emptyFunctionName         = ""
+	webappPortAddr            = ":9711"
```
leoporoli, line 47: I think that this value should be put in a centralized place as we did for the `grpcPortNum`, so we can see on the same file that the port number `9710` is for GRPC and `9711` is taken for the HTTP server and if you have to change the number, so change it only in one place.

engine/frontend/src/App.css
Peeeekay: i will keep it uppercase for now; App is usually uppercase as per react standards.

engine/frontend/src/App.js
Peeeekay: replied above.

engine/server/engine/main.go
```diff
@@ -131,6 +134,40 @@ func runMain() error {
...
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+
+			fileServer.ServeHTTP(w, r)
+		})
+
+		err := http.ListenAndServe(webappPortAddr, handler)
+		if err != nil {
+			logrus.Debugf("error while starting the webapp: \n%v", err)
```
leoporoli, line 167: if something fails here the engine will start without the HTTP server, I think it should fail loudly if the HTTP server can't start. So, you can pass an error channel to this subroutine and wait (in the main thread) until there is no error in order to continue

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
Peeeekay: will leave this for now; will create a ticket. good point. Will do it next PR - while adding this for k8s

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
Peeeekay: good catch; will create a ticket for this and add the label for frontend.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
Peeeekay: this is fine for now; because they will won't be able to only open ui on the browser and it's on v0 prototype phase - so don;t want to add this additional check.. - everything should still be operational regardless fe is up or not.

engine/frontend/src/App.test.js
Peeeekay: react `App` is generally uppercase.

engine/server/engine/main.go
```diff
@@ -131,6 +134,40 @@ func runMain() error {
...
+		}
```
leoporoli, line 168: I think that you shutdown gracefully the HTTP server when the engine is interrupted by calling the `server.Shutdown` method

engine/server/engine/main.go
Peeeekay: i don't think so i need to this atm because it's a prototype and it shouldn't impact the normal flow. When we productize fully, I agree we can fail loudly.

engine/server/engine/main.go
leoporoli: you could even use a defer call I think, like `defer server.Shutdown(ctx)`

Peeeekay: sorry; why do we need this though? Because kurtosis engine is running inside docker container; so we always stop/delete old and and start new one? so it will be abruptly close anyways? @leoporoli maybe I am missing something here?

leoporoli: @Peeeekay stopping the server gracefully will return the received HTTP request before the engine stops

Peeeekay: @leoporoli Ah I see; I don't think it is absolutely necessary to do that frontend - maybe overkill here - because as soon as an engine gets up - it should work fine. I don;t really want to impact the current behaviour which is kill the engine as soon as  grpcServer is gracefully stopped. Though this is one-liner change and can be done later on.

engine/frontend/src/components/LeftView.js
```diff
@@ -0,0 +1,27 @@
+export const LeftView = ({heading, renderList}) => {
+    // const renderServices = (services, handleClick) => {
```
leoporoli, line 2: reminder  for a next PR, do not forget to delete the commented code

engine/frontend/src/components/LogView.js
```diff
@@ -0,0 +1,36 @@
+export const LogView = ({heading, logs, classAttr}) => {
+    // const renderServices = (services, handleClick) => {
```
leoporoli, line 2: reminder  for a next PR, do not forget to delete the commented code

engine/frontend/src/components/Services.js
```diff
@@ -0,0 +1,146 @@
...
+
+    const [services, setServices] = useState([])
+    const [fileArtifacts, setFileArtifacts] = useState([])
+
+    const {state} = useLocation();
+    const {enclaves} =  state;
+    
+    useEffect(() => {
+        const fetch = async () => {
+            //const response = await axios.get(`http://localhost:5050/enclaves/${name}/services`)
```
leoporoli, line 83: reminder  for a next PR, do not forget to delete the commented code

engine/frontend/src/components/Services.js
```diff
@@ -0,0 +1,146 @@
...
+                <div className="flex flex-col space-y-2 overflow-auto h-1/3 border-8">
+                    <div className="flex flex-col space-y-2 overflow-auto text-2xl text-center justify-center pt-2">
+                        Files Artifacts
+                    </div>
+                    <div className="overflow-auto">
+                        {renderFileArtifacts(fileArtifacts)}
+                    </div>
+                </div>  
+            </div>
+            {/* <div className="col-span-5 grid grid-rows-6">
```
leoporoli, line 126: reminder  for a next PR, do not forget to delete the commented code

engine/server/engine/main.go
Peeeekay: not sure what you mean here?

engine/server/engine/main.go
Peeeekay: yepp; right - lets do that in next PR. for this it's fine to have it hard-coded.

engine/frontend/src/components/Services.js
Peeeekay: i purposely left this commented code; because i am going uncomment them and add these things functionalities to UI.

leoporoli (approved):

engine/server/engine/main.go
leoporoli: that these values can be engine's arguments, like the `serverArgs.GrpcListenPortNum` but we can do it in a next PR

PR #844, opened 2023-07-07
docs: Add highlevel explanation of idempotent runs to "Concepts" section to increase discoverability

mieubrisse:
## Description:
Retitles the current "idempotent runs" docs to be more explicit about what it is (a deep explanation of how idempotent runs work), and added a higher-level page to the "Concepts" section

YES


leeederek (approved):
I've added a backlink from the explanation to the concepts-reference page. As long as you're ok with this change, so am I.

Issue #848, opened 2023-07-07
A failed `run_sh` prints the task output without newlines


mieubrisse:
### What's your CLI version?

0.80.6

### Description & steps to reproduce

1. Run a `run_sh` command that prints multiple lines of output, but will fail
2. Observe that it puts everything on one line, e.g.: `Caused by: error occurred and shell command: "cd /workspace && bazel build --output_user_root=/output //src/..." exited with code 2 with output "Extracting Bazel installation...\nStarting local Bazel server and connecting to it...\nERROR: --output_user_root=/output :: Unrecognized option: --output_user_root=/output\n"`

### Desired behavior

The output is exactly as the user would have seen it if they were running the command locally, e.g.:
```
Caused by: Shell command 'cd /workspace && bazel build --output_user_root=/output //src/...' exited with non-zero exit code 2 with output:
  Extracting Bazel installation...
  Starting local Bazel server and connecting to it...
  ERROR: --output_user_root=/output :: Unrecognized option: --output_user_root=/output
```

### What is the severity of this bug?

Papercut; this bug is frustrating, but I have a workaround.

PR #852, opened 2023-07-10
fix: improve the error message

Peeeekay:
https://github.com/kurtosis-tech/kurtosis/issues/848

Made the error log same as we have when we run `plan.exec()`


internal_testsuites/golang/testsuite/startosis_run_sh_task_test/run_task_sh_task_test.go
```diff
@@ -63,7 +63,7 @@ func TestStarlark_RunshTaskFileArtifact(t *testing.T) {
 func TestStarlark_RunshTaskFileArtifactFailure(t *testing.T) {
 	ctx := context.Background()
 	runResult, _ := test_helpers.SetupSimpleEnclaveAndRunScript(t, ctx, runshTest, runshStarlarkFileArtifactFailure)
-	expectedErrorMessage := "error occurred and shell command: \"cat /tmp/kurtosis.txt\" exited with code 1 with output \"cat: can't open '/tmp/kurtosis.txt': No such file or directory\\n"
+	expectedErrorMessage := "cat: can't open '/tmp/kurtosis.txt': No such file or directory"
```
laurentluce, line 66: Just curious why you are only checking a subset of the error message and not the entire one.

laurentluce: From @Peeeekay : "Because the error also prints out the path of the file; which is non deterministic as it can change depending upon whether the tests are runnin locally or on ci"

laurentluce (approved):

PR #861, opened 2023-07-11
feat: added support for users to only copy the content and not the parent dir for run_sh task

Peeeekay:
The idea is that users can specify whether they want the parent directory or not in the file-artifact while using run_sh and store_service_files by using special symbol '*'

```
plan.store_service_files(
        service_name = "old-pattern",
        src = "/data/*" // will only store the content of the `data` directory as a file artifact
   )
```


container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/copy_files_from_user_service.go
```diff
@@ -60,15 +64,27 @@ func CopyFilesFromUserService(
...
+			srcPath,
+		)
+	} else {
+		commandToRun = fmt.Sprintf(
+			commandString,
+			srcPathDir,
+			srcPathBase,
+			srcPath,
+		)
+	}
```
laurentluce:
```suggestion
	if srcPathBase == doNotIncludeParentDirInArchiveSymbol {
		srcPathBase = ignoreParentDirInArchiveSymbol
	}
	commandToRun = fmt.Sprintf(
		commandString,
		srcPathDir,
		srcPathBase,
		srcPath,
	)
```

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/copy_files_from_user_service.go
```diff
@@ -60,15 +64,27 @@ func CopyFilesFromUserService(
 	srcPath = filepath.Clean(srcPath)
 	// we get the base dir | file
 	srcPathBase := filepath.Base(srcPath)
-	// we get the dir that holds base the dir | file
+	//// we get the dir that holds base the dir | file
```
laurentluce:
```suggestion
	// we get the dir that holds base the dir | file
```

docs/docs/starlark-reference/plan.md
```diff
@@ -432,15 +432,17 @@ The `files` dictionary argument accepts a key value pair, where `key` is the pat
...
 
 ```python
 
     result = plan.run_sh(
         run = "mkdir -p task && cd task && echo kurtosis > test.txt",
         store = [
             "/task",
-            "/task/test.txt",
+            # using '*' will only copy the contents of the parent directory and not the directory itself to file artifact
+            # in this case, only test.txt will be stored and task directory will be ignored
```
laurentluce, line 444: I find this a bit confusing since `/task` is included just above.  Can this example be updated to use two different dir names?

Peeeekay: The idea is to show - how `/task` and `/task/*` behaves differently so decided to have them besides each other; if this is super confusing I can divide this into 2.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/copy_files_from_user_service.go
Peeeekay: ah man; that's neater - didn't even realized it. 💯

docs/docs/starlark-reference/plan.md
laurentluce: Got it.

laurentluce (approved):

PR #867, opened 2023-07-12
Pk/add connect go

Peeeekay:
YES/NO


engine/frontend/src/api/enclave.js
```diff
@@ -15,22 +20,27 @@ const createApiPromiseClient = (apiClient) => {
     return "";
 }
 
+
 export const getEnclavesFromKurtosis = async () => {
-    const respFromGrpc = await engineClient.getEnclaves(new google_protobuf_empty_pb.Empty, null);
-    const response = respFromGrpc.toObject()
+    const response = await axios.post("http://localhost:9710/engine_api.EngineService/GetEnclaves", {"field":""}, {"headers":{'Content-Type': "application/json"}})
```
Peeeekay, line 25: This is an example for rest endpoint.

engine/server/engine/server/engine_connect_server_service.go
```diff
@@ -0,0 +1,133 @@
+package server
```
Peeeekay, line 1: This is the new way of creating the server; need to do the same api_container. pretty straight-forward imo. The complexity is with streams.

PR #868, opened 2023-07-12
refactor: Use smarter FetchImage rather than PullImage

gbouv:
NO


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
```diff
@@ -159,8 +159,7 @@ func (backend *DockerKurtosisBackend) CreateAPIContainer(
 		labelStrs,
 	).Build()
 
-	// Best-effort pull attempt
-	if err = backend.dockerManager.PullImage(ctx, image); err != nil {
+	if err = backend.dockerManager.FetchImage(ctx, image); err != nil {
```
h4ck3rk3y, line 162: This would still pull the latest image even if the previous image is available locally? Commenting on this instance but this applies to other instances of the replacement in this PR as well

Say I am on 0.80.7 and I upgrade to 0.80.8; I have images from 0.80.7. Will this still work?

h4ck3rk3y (approved):

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
gbouv: I think so yeah, in those cases we're passing the image name + version, so I believe we should pull it. That's my understanding of the code by reading it, but there's no easy way to test that here because by building it locally, we get the image with custom version from local docker. So I'd say we can merge it and if we see it's not happening, we can revert.

h4ck3rk3y: Works for me man. That was my read as well but I hadn't dived deep enough some of them definitely have args and if all of them do thats better :)

PR #869, opened 2023-07-12
refactor: Docker client timeout and preventively check that Portal is running when using a remote context

gbouv:
NO


container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
```diff
@@ -9,10 +9,12 @@ import (
 	"github.com/sirupsen/logrus"
 	"os"
 	"path"
+	"time"
 )
 
 const (
-	urlScheme = "tcp"
+	urlScheme           = "tcp"
+	dockerClientTimeout = 10 * time.Second
```
adschwartz: @gbouv What happens when the timeout expires? Do we log an error or does it reconnect without incident to the user?
If we break the connection then I think we can do 30s in the remote context. When you are on the internet I can imagine scenarios where we could potentially trigger a false positive if we go as low as 10s.

adschwartz (approved):
@gbouv looks good, just one question to address.

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
gbouv: It just throws an error.
10s to _just_ establish the connection seems extremely slow to me. I mean, if you need 10 second to just reach remote docker, there's no way you're gonna be able to decently run any Kurtosis package. So I'd say failing early is still better.
I'll bump it to 30 seconds if you feel better about it, I don't really mind TBH

adschwartz: 10s is slow, but I think 30s is a good practice when dealing with internet based resources. A load balancer could momentarily get in a bad state while it's switching between instances. It's not super uncommon.

PR #870, opened 2023-07-12
feat: added a command that opens the Kurtosis Web UI

h4ck3rk3y:
## Description:
Opens the Web UI via `kurtosis web`

Yes


docs/docs/cli-reference/web.md
```diff
@@ -0,0 +1,11 @@
+---
```
Peeeekay, line 1: sorry; could you add `beta` here too. thanks :D I would have done it, but I am not near my laptop atm.

h4ck3rk3y: Done :)

Peeeekay (approved):

PR #879, opened 2023-07-12
feat: added connect-go for engine

Peeeekay:
This PR will include the connect-go updates only related to engine; I am relying on the CI pipeline tests to verify it's correctness - which atm looks good. This PR will also include frontend changes that will allow me to connect to engine.


.circleci/config.yml
```diff
@@ -108,7 +108,7 @@ parameters:
     default: 20.10.7
   api-go-version:
     type: string
-    default: "1.18"
+    default: "1.19"
```
Peeeekay: Even though doc says below; connect-go fails if go --version is not 1.19.
```
You'll need one of the [last two major releases](https://golang.org/doc/devel/release) of Go, with a minimum of Go 1.18. See Go's [Getting Started](https://golang.org/doc/install) guide for installation instructions.
```

gbouv: Bumping the Go version of the API package means anyone using the SDK on 1.18 will be unable to use it anymore. I know at some point it was a bit of a concern to force users to use recent versions of Go, not sure where we stand those days, but worth being extra loud about it if we end up merging this

api/scripts/protobuf-bindings-generator.sh
```diff
@@ -185,6 +185,8 @@ generate_golang_bindings() {
                 "${go_grpc_out_flag}" \
                 "${go_grpc_module_flag}" \
                 "${go_grpc_unimplemented_servers_flag}" \
+                "${connect_go_grpc_flag}" \
+                "${connect_go_module_flag}" \
```
gbouv, line 187: So, do we need to generate both the connect_go bindings and the "normal" bindings? I haven't looked at the generated files. Are connect_go bindings a layer on top of the normal bindings?

connect-server/server.go
```diff
@@ -0,0 +1,82 @@
...
+
+	<-stopper
+	serverStoppedChan := make(chan interface{})
+	go func() {
+		httpServer.Shutdown(context.Background())
+		serverStoppedChan <- nil
+	}()
+	select {
+	case <-serverStoppedChan:
+		logrus.Debug("gRPC server has exited gracefully")
```
gbouv: nit: is `gRPC` the right word here?

engine/server/engine/main.go
```diff
@@ -121,19 +120,6 @@ func runMain() error {
 
 	logsDatabaseClient := kurtosis_backend.NewKurtosisBackendLogsDatabaseClient(kurtosisBackend)
 
-	engineServerService := server.NewEngineServerService(serverArgs.ImageVersionTag, enclaveManager, serverArgs.MetricsUserID, serverArgs.DidUserAcceptSendingMetrics, logsDatabaseClient)
```
gbouv: If we're not instantiating that guy anymore, can we remove the associated file? Or it's used somewhere else?

engine/server/engine/server/engine_connect_server_service.go
```diff
@@ -0,0 +1,244 @@
...
+			shouldFollowLogs,
+		)
+	}
+	defer func() {
+		if cancelCtxFunc != nil {
+			cancelCtxFunc()
+		}
+	}()
+
+	for {
```
gbouv: I assume all this logic in this file is mostly copied from engine_server_service.go, is that correct? Haven't reviewed the code thoroughly here

gbouv (commented):
It doesn't look too controversial from my perspective. Though it's worth discussing the bump to go 1.19

api/scripts/protobuf-bindings-generator.sh
Peeeekay: Yes; we would need gRPC and connect bindings for Go. Because our clients still going to be using go-gRPC. Once we switch them to use connect, we can remove go-grpc bindings. Conenct just creates a new bindings that is backward compatible with go-grpc.

engine/server/engine/main.go
Peeeekay: yepp; will do that.

engine/server/engine/server/engine_connect_server_service.go
Peeeekay: yes; i have updated the how we handle request and response; the only thing to look for is streams.. There is no notion of stream.Context() so needed to change that a bit for `GetServiceLogs`

.circleci/config.yml
Peeeekay: @gbouv have message the folks at buf to see what they say.. i agree with you;; will ping the team regarding this.

adschwartz:
@Peeeekay this is still a draft, should I review this?

Peeeekay:
> @Peeeekay this is still a draft, should I review this?

Just wanted an initial review to make sure there are no glaring red flags - will open this to review in a bit as it seems like it's good to proceed with. @adschwartz

adschwartz:
Hey @Peeeekay . Do we need some instructions on how to get the `protoc-gen-connect-go` installed and running? When I run the regenerate script it fails like this:
```
kurtosiss-MacBook-Pro:scripts ads$ /usr/bin/env bash /Users/ads/code/kurtosis-tech/kurtosis-bravo/api/scripts/regenerate-protobuf-bindings.sh
protoc-gen-connect-go: program not found or is not executable
Please specify a program using absolute path or make sure the program is available in your PATH system variable
--connect-go_out: protoc-gen-connect-go: Plugin failed with status code 1.
```

adschwartz:
> Hey @Peeeekay . Do we need some instructions on how to get the `protoc-gen-connect-go` installed and running? When I run the regenerate script it fails like this:
> 
> ```
> kurtosiss-MacBook-Pro:scripts ads$ /usr/bin/env bash /Users/ads/code/kurtosis-tech/kurtosis-bravo/api/scripts/regenerate-protobuf-bindings.sh
> protoc-gen-connect-go: program not found or is not executable
> Please specify a program using absolute path or make sure the program is available in your PATH system variable
> --connect-go_out: protoc-gen-connect-go: Plugin failed with status code 1.
> ```

@Peeeekay It may be this instruction that we need?
https://connectrpc.com/docs/go/getting-started/#install-tools
https://pkg.go.dev/github.com/bufbuild/connect-go@v1.9.0/cmd/protoc-gen-connect-go#pkg-overview

connect-server/server.go
adschwartz: Fixed

adschwartz (approved):

adschwartz:
@Peeeekay I touched up a few comments and added 2 extra functions to better enable controlling the CORS settings which we need for the KLOUD:
https://github.com/kurtosis-tech/kurtosis/pull/879/commits/391f488e951b66111fe290a8a3cfbcd35a869bb1
https://github.com/kurtosis-tech/kurtosis/pull/879/commits/4d4d47cc1c76063e411ff34ea0507a5189fa2244

Let's get this guy merged when you are back from your break.

engine/frontend/src/api/enclave.js
```diff
@@ -15,50 +26,61 @@ const createApiPromiseClient = (apiClient) => {
...
+    }
+}
+
+export const getServiceLogs = async (enclaveName, serviceUuid) => {
+    const args = {
+        "enclaveIdentifier": enclaveName,
+        "serviceUuidSet": {
+            [serviceUuid]: true
+        },
+        followLogs: false,
```
Peeeekay: @adschwartz ideally we want this to be true; but seeing some weird behaviours where stream requests get stuck on pending;; however having it as false works fine - we get the logs but don;t tail it.. in the interest of time, i can look into that in the next pr ( known bug) .. while this can get merged for e2e enclave-manager. My hunch is that it's happening because I am not correctly cancelling the requests when I unmount the stream from react page; but will have better idea tomorrow hopefully.

adschwartz: @Peeeekay agreed, good idea to look into if we are following the right procedure when handling the streams. 

If you need another day to merge this, then that can work too. 

Just for my curiosity, what does `followLogs=true` do?

Peeeekay: It allows us to tail logs and see the new logs  in real-time on the UI instead of refreshing the page

Peeeekay: it works now! used AbortController.

PR #882, opened 2023-07-12
feat: add `kurtosis cloud load to CLI`

adschwartz:
## Description:
In this PR we are adding the `kurtosis cloud load <instance_id>` option that enables the user to load the config from the cloud and automatically switch the context.

To use the command the user must export an API key to the env var: `KURTOSIS_CLOUD_API_KEY`

YES


api/golang/engine/lib/cloud/cloud.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	api "github.com/kurtosis-tech/kurtosis-cloud-backend/api/golang/kurtosis_backend_server_rpc_api_bindings"
+	"github.com/kurtosis-tech/stacktrace"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+func CreateCloudClient(connectionStr string) (api.KurtosisCloudBackendServerClient, error) {
+	conn, err := grpc.Dial(connectionStr, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
adschwartz, line 11: This should be a secure connection

cli/cli/kurtosis_config/resolved_config/kurtosis_config.go
```diff
@@ -15,6 +15,9 @@ const (
...
...NlcnZpY2VzIFJvb3QgQ2VydGlmaWNhdGUgQXV0aG9yaXR5IC0gRzIwggEi\nMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDVDDrEKvlO4vW+GZdfjohTsR8/\ny8+fIBNtKTrID30892t2OGPZNmCom15cAICyL1l/9of5JUOG52kbUpqQ4XHj2C0N\nTm/2yEnZtvMaVq4rtnQU68/7JuMauh2WLmo7WJSJR1b/JaCTcFOD2oR0FMNnngRo\nOt+OQFodSk7PQ5E751bWAHDLUu57fa4657wx+UX2wmDPE1kCK4DMNEffud6QZW0C\nzyyRpqbn3oUYSXxmTqM6bam17jQuug0DuDPfR+uxa40l2ZvOgdFFRjKWcIfeAg5J\nQ4W2bHO7ZOphQazJ1FTfhy/HIrImzJ9ZVGif/L4qL8RVHHVAYBeFAlU5i38FAgMB\nAAGjgfAwge0wDwYDVR0TAQH/BAUwAwEB/zAOBgNVHQ8BAf8EBAMCAYYwHQYDVR0O\nBBYEFJxfAN+qAdcwKziIorhtSpzyEZGDMB8GA1UdIwQYMBaAFL9ft9HO3R+G9FtV\nrNzXEMIOqYjnME8GCCsGAQUFBwEBBEMwQTAcBggrBgEFBQcwAYYQaHR0cDovL28u\nc3MyLnVzLzAhBggrBgEFBQcwAoYVaHR0cDovL3guc3MyLnVzL3guY2VyMCYGA1Ud\nHwQfMB0wG6AZoBeGFWh0dHA6Ly9zLnNzMi51cy9yLmNybDARBgNVHSAECjAIMAYG\nBFUdIAAwDQYJKoZIhvcNAQELBQADggEBACMd44pXyn3pF3lM8R5V/cxTbj5HD9/G\nVfKyBDbtgB9TxF00KGu+x1X8Z+rLP3+QsjPNG1gQggL4+C/1E2DUBc7xgQjB3ad1\nl08YuW3e95ORCLp+QCztweq7dp4zBncdDQh/U90bZKuCJ/Fp1U1ervShw3WnWEQt\n8jxwmKy6abaVd38PMV4s/KCHOkdp8Hlf9BRUpJVeEXgSYCfOn8J3/yNTd126/+pZ\n59vPr5KW7ySaNRB6nJHGDn2Z9j8Z3/VyVOEVqQdZe4O/Ui5GjLIAZHYcSNPYeehu\nVsyuLAOQ1xk4meTKCRlb/weWsKh/NEnfVqn3sF/tM+2MR7cwA130A4w=\n-----END CERTIFICATE-----"
```
adschwartz: This is the public certificate chain for `cloud.kurtosis.com:8080`. I think we'll need a better way to store this, especially because they could change in the future. But putting it here for now

adschwartz:
Note, the GitGuardian complaint about an API key being found can be ignored. It was a test key that's burned and no api keys are in the source code

cli/cli/command_str_consts/command_str_consts.go
```diff
@@ -23,6 +22,8 @@ const (
 	ContextLsCmdStr         = "ls"
 	ContextRmCmdStr         = "rm"
 	ContextSwitchCmdStr     = "switch"
+	CloudCmdStr             = "cloud"
+	CloudLoadCmdStr         = "load"
```
gbouv: Can we keep the alphabetical ordering here?

api/golang/engine/lib/cloud/cloud.go
```diff
@@ -0,0 +1,28 @@
...
+	"crypto/tls"
+	"crypto/x509"
+	api "github.com/kurtosis-tech/kurtosis-cloud-backend/api/golang/kurtosis_backend_server_rpc_api_bindings"
+	"github.com/kurtosis-tech/stacktrace"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+func CreateCloudClient(connectionStr string, caCertChain string) (api.KurtosisCloudBackendServerClient, error) {
+	f := []byte(caCertChain)
```
gbouv: nit: We tend to give explicit names to variables

cli/cli/commands/cloud/load/load.go
```diff
@@ -0,0 +1,134 @@
...
+const (
+	instanceIdentifierArgKey      = "instance-id"
+	instanceIdentifierArgIsGreedy = false
+	kurtosisCloudApiKeyEnvVarArg  = "KURTOSIS_CLOUD_API_KEY"
+)
+
+var LoadCmd = &lowlevel.LowlevelKurtosisCommand{
+	CommandStr:       command_str_consts.CloudLoadCmdStr,
+	ShortDescription: "Load a Kurtosis Cloud instance",
+	LongDescription: "Load a remote Kurtosis Cloud instance into the local context by providing the instance id." +
```
gbouv: "... into the _local_ context ..."?? What does that mean?

cli/cli/commands/cloud/load/load.go
```diff
@@ -0,0 +1,134 @@
...
+		return stacktrace.Propagate(err, "Failed to base64 decode context config")
+	}
+
+	parsedContext, err := add.ParseContextData(decodedConfigBytes)
+	if err != nil {
+		return stacktrace.Propagate(err, "Unable to decode context config")
+	}
+
+	contextsConfigStore := store.GetContextsConfigStore()
+	// We first have to remove the context incase it's already loaded
```
gbouv, line 88: Really? Would throwing be better in that case? Erasing some data silently is a bit borderline IMO. We could add a `-f` option to force the override, but by default I'd just return with a nice message

cli/cli/commands/cloud/load/load.go
```diff
@@ -0,0 +1,134 @@
...
+	err = contextsConfigStore.RemoveContext(parsedContext.Uuid)
+	if err != nil {
+		return stacktrace.Propagate(err, "While attempting to reload the context with uuid %s an error occurred while removing it from the context store", parsedContext.Uuid)
+	}
+	err = add.AddContext(parsedContext)
+	if err != nil {
```
gbouv: You can inline the function call in those `if err := nil`, here and above

cli/cli/kurtosis_config/overrides_objects/v3/kurtosis_config_v3.go
```diff
@@ -0,0 +1,26 @@
...
+//  2. it's easier to write
+//  3. it's consistent with previous properties and changing the format of an already-written config file is very difficult
+type KurtosisConfigV3 struct {
+	// vvvvvvvvv Every new Kurtosis config version must have this key vvvvvvvv
+	ConfigVersion config_version.ConfigVersion `yaml:"config-version"`
+	// ^^^^^^^^^ Every new Kurtosis config version must have this key ^^^^^^^^
+
+	ShouldSendMetrics *bool                               `yaml:"should-send-metrics,omitempty"`
+	KurtosisClusters  map[string]*KurtosisClusterConfigV3 `yaml:"kurtosis-clusters,omitempty"`
+	CloudConfig       *KurtosisCloudConfigV3              `yaml:"cloud-config,omitempty"`
```
gbouv: Sooo, if that's a pointer with `omitempty` do you need all this migration jazz?? I guess a nullable pointer is backward compatible in this case, right?

cli/cli/kurtosis_config/resolved_config/kurtosis_config.go
```diff
@@ -15,6 +15,11 @@ const (
...
```
gbouv, line 22: Can we put that in a resource file maybe?

gbouv (commented):
Looks good in general, but I think we don't need to declare a V3 for the config object and handle the migration logic. The change to the config you're making is backward compatible.

cli/cli/kurtosis_config/resolved_config/kurtosis_config.go
adschwartz: haha yes it's not pretty here. My thinking is that we'd need to pull this from somewhere outside the code base to make it more dynamic. These certs can potentially change over time so we don't want it to be static,.

adschwartz: I'd prefer to leave it where it is for now and replace this with a more dynamic solution soon

adschwartz: Let me know if you prefer me to make this change or not. I'm leaning towards leaving it and optimizing in later iterations.

cli/cli/commands/cloud/load/load.go
adschwartz: I don't fully agree that we are silently erasing data. We are replacing the config that was already there. Forcing it seems unnecessary in my opinion.

laurentluce (approved):

PR #884, opened 2023-07-13
feat: Starlark package arguments will be parsed as a deep Struct when `"_kurtosis_parser": "struct"` is passed in the arguments JSON

gbouv:
## Description:
Running `kurtosis run github.com/path/to/package '{"_kurtosis_parser": "struct", "arg1": "val1", "arg2": "val2", ...}'` will parse the arguments as a Starlark struct, rather than a simple dictionary.

YES

### DIFF
--- a/core/server/api_container/server/startosis_engine/package_io/package_io_test.go
+++ b/core/server/api_container/server/startosis_engine/package_io/package_io_test.go
@@ -1,8 +1,10 @@
 package package_io
 
 import (
+	"fmt"
 	"github.com/stretchr/testify/require"
 	"go.starlark.net/starlark"
+	"go.starlark.net/starlarkstruct"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


h4ck3rk3y (approved):

core/server/api_container/server/startosis_engine/package_io/package_io_test.go
```diff
@@ -1,8 +1,10 @@
 package package_io
 
 import (
+	"fmt"
```
h4ck3rk3y, line 4: can we add an integration test?

h4ck3rk3y: Sweet. Question, we had some problems where a dictionary inside the struct was still a dictionary; did we sort that out?

h4ck3rk3y (approved):

core/server/api_container/server/startosis_engine/package_io/package_io_test.go
gbouv: Do we need an integration test? It's not a lot of effort to add one, but IMO the feature is well covered with the unit test, and integrations test are really slow so we should add one only when it's necessary

gbouv: And to your question, yes, a nested dictionary in the tope level JSON dict will become a nested struct inside the top level struct

PR #885, opened 2023-07-13
feat: Add file artifact inspect API do APIC

victorcolombo:
## Description:

Screenshot 2023-07-13 at 12 19 37

YES/NO


core/server/api_container/server/api_container_service.go
```diff
@@ -130,6 +137,36 @@ func (apicService ApiContainerService) UploadStarlarkPackage(server kurtosis_cor
...
+	filesArtifact, err := apicService.filesArtifactStore.GetFile(artifactIdentifier)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred getting files artifact '%v'", artifactIdentifier)
+	}
+
+	fileDescriptions, err := getFileDescriptionsFromArtifact(filesArtifact.GetAbsoluteFilepath())
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred getting file descriptions from '%v'", artifactIdentifier)
+	}
+	if args.GetFilePath() != "" {
```
Peeeekay: just to confirm; we first find descriptions for all the files; and then check if it exists for the file in question. We are already fetching all the descriptions, wouldn't it be worth for us to just return all of it back to users? Otherwise they would need to make multiple requests to inspect say more than 3 files?

victorcolombo: I did this out of simplicity, but you are right. Let's chat offline on the best API for us both (CLI and FE)

api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -345,6 +345,20 @@ func (enclaveCtx *EnclaveContext) DownloadFilesArtifact(ctx context.Context, art
 	return fileContent, nil
 }
 
+func (enclaveCtx *EnclaveContext) InspectFilesArtifact(ctx context.Context, artifactName services.FileArtifactName) (*kurtosis_core_rpc_api_bindings.InspectFilesArtifactContentsResponse, error) {
+	// TODO(vcolombo): Add a more intuitive return type to this call instead of returning the RPC response
+	response, err := enclaveCtx.client.InspectFilesArtifactContents(ctx, &kurtosis_core_rpc_api_bindings.InspectFilesArtifactContentsRequest{
```
Peeeekay, line 350: I think we could potentially return all the file-artifacts; but not urgent or needed today.

api/protobuf/core/api_container_service.proto
```diff
@@ -428,3 +430,21 @@ message FilesArtifactNameAndUuid {
...
+message InspectFilesArtifactContentsRequest {
+  FilesArtifactNameAndUuid file_names_and_uuid = 1;
+}
+
+message InspectFilesArtifactContentsResponse {
+  repeated FileArtifactContentsFileDescription file_descriptions = 1;
+}
+
+message FileArtifactContentsFileDescription {
+  string path = 1;
```
Peeeekay, line 448: what is the path relative to? the path of the file in the tar ball we create?

core/server/api_container/server/api_container_service.go
```diff
@@ -130,6 +138,33 @@ func (apicService ApiContainerService) UploadStarlarkPackage(server kurtosis_cor
 	return nil
 }
 
+func (apicService ApiContainerService) InspectFilesArtifactContents(_ context.Context, args *kurtosis_core_rpc_api_bindings.InspectFilesArtifactContentsRequest) (*kurtosis_core_rpc_api_bindings.InspectFilesArtifactContentsResponse, error) {
+	artifactIdentifier := ""
+	if args.GetFileNamesAndUuid().GetFileUuid() != "" {
```
Peeeekay: nit: `""` this can be const.

core/server/api_container/server/api_container_service.go
```diff
@@ -651,3 +686,64 @@ func (apicService ApiContainerService) runStarlark(
...
+
+	tarReader := tar.NewReader(gzipReader)
+
+	fileDescriptions := []*kurtosis_core_rpc_api_bindings.FileArtifactContentsFileDescription{}
+	for {
+		header, err := tarReader.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
```
Peeeekay, line 711: Curious: could there be edge-case this happens only for just one file? but rest all the files are good and can be returned.
Maybe we could return in future {succesfulFileContents: [...], errorFileContentes:[..]}

core/server/api_container/server/api_container_service.go
```diff
@@ -651,3 +686,64 @@ func (apicService ApiContainerService) runStarlark(
...
+			return nil, stacktrace.Propagate(err, "Failed to get header from artifact path '%v'", artifactPath)
+		}
+
+		filePath := header.Name
+		description := fmt.Sprintf("Size: %v", header.Size)
+		textPreview, err := getTextRepresentation(tarReader, unlimitedLineCount)
+		if err != nil {
+			logrus.Debugf("Failed to get text preview for file '%v' with error '%v'", filePath, textPreview)
```
Peeeekay: i think it would be good to return the file-artifacts that errored out as well; though can be done as a FLUP.

Peeeekay (approved):
looks good; some minor comments.

api/protobuf/core/api_container_service.proto
victorcolombo: Yes. Added a comment for clarification

core/server/api_container/server/api_container_service.go
victorcolombo: Added a TODO

core/server/api_container/server/api_container_service.go
victorcolombo: No, we still TARGZ one file. 1 artifact == 1 tgz, no matter the contents. This error is probably a bug in kurtosis, given that the tar is malformated

core/server/api_container/server/api_container_service.go
victorcolombo: Done

api/golang/core/lib/enclaves/enclave_context.go
victorcolombo: Possibly, but this is easily chain-able using the `ListFilesArtifactNamesAndUuids` call

PR #911, opened 2023-07-17
ci: run the golang tests faster

h4ck3rk3y:
YES/NO


h4ck3rk3y:
no improvement here at all

PR #918, opened 2023-07-17
docs: Add extra documentation on import_module and read_file for rela…

mieubrisse:
…tive imports

## Description:
I wanted relative imports to be front and center in the user's face, so they're not struggling with absolute imports.

YES


h4ck3rk3y (approved):
You can now add this to upload_file too

PR #939, opened 2023-07-19
fix: remove task once completed

Peeeekay:


core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
```diff
@@ -260,3 +260,11 @@ func formatErrorMessage(errorMessage string, errorFromExec string) string {
 	reformattedErrorMessage := strings.Join(splitErrorMessageNewLine, "\n  ")
 	return fmt.Sprintf("%v\n  %v", errorMessage, reformattedErrorMessage)
 }
+
```
h4ck3rk3y, line 263: perhaps we can just do

defer removeService() - and have this log instead of throwing an error.

h4ck3rk3y (approved):
Minor comment looks good otherwise. I am worried about the test being broken though. Is `remove_service` really that slow!?

I will retrigger this but - this took 480s and timed out while normally this suite takes 150s; lets see how the retrigger goes

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
Peeeekay: sorry; did not get your comment;; this log is only used at the moment for exec errors.. Are you suggesting instead of throwing error on remove_service, I should just log it @h4ck3rk3y ?

h4ck3rk3y: you use this function in a defer block upstream and log it; so it never gets "thrown"; so i was wondering if this could just log it. nothing major though.

PR #941, opened 2023-07-19
refactor: Add skipped flag to instruction returned by the runStarlark command

gbouv:
## Description:
Useful in the SDK to know whether an instruction will be executed or not

NO


h4ck3rk3y (approved):
Looks good to me.

PR #947, opened 2023-07-19
feat: Add autocomplete for file path of artifact files inspect

victorcolombo:
## Description:
This PR adds a couple of improvements to the recent `file inspect`:

1. Autocomplete for files inside the artifact
Screenshot 2023-07-19 at 13 16 57
2. Prettier coloring of files that contain a text description
Screenshot 2023-07-19 at 13 25 22
3. Render correctly root level files
4. Drop artifact name on file tree

YES


mieubrisse:
@victorcolombo wdyt if we do the coloring like Unix default? white for regular files, and blue for directories?

cli/cli/commands/files/inspect/inspect.go
```diff
@@ -188,3 +196,48 @@ func humanReadableSize(size uint64) string {
...
+		if err != nil {
+			return nil, stacktrace.Propagate(
+				err,
+				"An error occurred getting the file artifacts",
+			)
+		}
+		fileArtifactContentPaths := []string{}
+		for _, fileArtifactDescription := range fileArtifactContents.GetFileDescriptions() {
+			fileArtifactContentPath := fileArtifactDescription.GetPath()
+			if fileArtifactDescription.TextPreview != nil {
```
h4ck3rk3y, line 236: why is completion only for those without text preview?

h4ck3rk3y (approved):

cli/cli/commands/files/inspect/inspect.go
victorcolombo: Printing file contents for files we have no contents yields no meaningful result. I can add a follow up to print something like "no file contents exist"

leeederek:
Hi we're missing docs for this @victorcolombo

PR #948, opened 2023-07-19
chore: Revert changes (added Testify.TestSuite for reusing enclaves on tests…

### LINKED ISSUE #740: Rollback the changes that breaks the test isolation on the golang internal test suite
### What's your CLI version?

0.78.4

We merged [this PR](https://github.com/kurtosis-tech/kurtosis/pull/722) which is breaking test isolation on some packages inside the golang internal test. We did this because we wanted to improve CI pipeline performance.

There are no conflicts between the tests now, some of them are sharing the same enclave, but we don't want to break the isolation on the tests so we have to rollback these changes once we improve the performance with another approach, right now another approach is the `pool of enclave` project: more context here: https://github.com/kurtosis-tech/kurtosis-private/issues/488

Context about the CI problem here:
https://www.notion.so/kurtosistech/Improve-K8s-tests-performance-cc4d7155430e4aeebb61f7830cdf3d53

### Desired behavior

Rollback this change after improving performance on the CI with another approach

leoporoli:
…uite packages) made on Golang internal test suite on Pr #722

NO

@gbouv it can be reviewed after #956 got merged

Fix #740


.circleci/config.yml
```diff
@@ -47,11 +51,56 @@ steps_prepare_testing_k8s: &steps_prepare_testing_k8s
...
+          
+          docker load -i  "<< pipeline.parameters.workspace-with-cli-binary-and-images-mountpoint >>/<< pipeline.parameters.file-artifacts-expander-image-filename >>"
+          
+          k3d image import "kurtosistech/core:$kurtosis_version" "kurtosistech/engine:$kurtosis_version" "kurtosistech/files-artifacts-expander:$kurtosis_version"
+    # Edit the Kurtosis config file in order to have K3S cluster configuration
+    - run:
+        name: Add K3S cluster config in Kurtosis config file
+        command: |
+          KURTOSIS_CONFIG_FILEPATH=$($KURTOSIS_BINPATH config path)
+          echo -e "config-version: 2\nshould-send-metrics: false\nkurtosis-clusters:\n  docker:\n    type: docker\n  k3d-k3s-default:\n    type: kubernetes\n    config:\n       kubernetes-cluster-name: k3d-k3s-default\n       storage-class: standard\n       enclave-size-in-megabytes: 2048" > "$KURTOSIS_CONFIG_FILEPATH"
```
gbouv: Can you use bash multiline string here? If at some point we need to make a change to this config, it's a real pain to mentally parse it

hola.txt
```diff
@@ -1 +0,0 @@
-hola
```
gbouv: nice

internal_testsuites/golang/testsuite/startosis_package_test/startosis_no_run_in_main_star_test.go
```diff
@@ -2,19 +2,39 @@ package startosis_package_test
...
 
-func (suite *StartosisPackageTestSuite) TestStartosisPackage_NoMainInMainStar() {
+func TestStartosisPackage_NoMainInMainStar(t *testing.T) {
+	t.Parallel()
 	ctx := context.Background()
-	runResult, _ := suite.RunPackage(ctx, packageWithNoMainInMainStarRelPath)
 
-	t := suite.T()
+	// ------------------------------------- ENGINE SETUP ----------------------------------------------
+	enclaveCtx, destroyEnclaveFunc, _, err := test_helpers.CreateEnclave(t, ctx, invalidCaseNoMainInMainStarTestName, isPartitioningEnabled)
```
gbouv: reading those duplicate pieces of code, I'm actually wondering if using the suite was not better ^^ What happens if we just change `SetupSuite` with `SetupTest` and same for `TearDown*`? 
Anyway, that's fine for now, just wondering since we're at it if there'a way to DRY this out a little bit

gbouv (approved):

gbouv:
Maybe add to the PR description that you're introducing k3s as an alternative k8s distro to run the test suites

.circleci/config.yml
leoporoli: done!

internal_testsuites/golang/testsuite/startosis_package_test/startosis_no_run_in_main_star_test.go
leoporoli: good idea! 
Done!

PR #956, opened 2023-07-20
chore: k3s on ci

leoporoli:
NO

It's related to the "improve performance on CI" project


h4ck3rk3y (approved):
Perhaps we can remove support for minikube on circle and make it cleaner. If we need to bring it back we just look at commit history

PR #963, opened 2023-07-24
feat: Start engine remotely with remote backend config when the context is remote

laurentluce:
## Description:
This is the first part to [support the bastion with the docker and k8s remote backend](https://www.notion.so/kurtosistech/Kloud-Bastion-Docker-and-K8S-backends-71a19e630ed0454a984042e1cafb1a88?pvs=4#8f3b10e02c874bdd9b9aad22acc76e14).  The CLI starts the engine remotely when the context is remote.  The engine reads the remote backend config at startup.  The CLI requests the portal client to forward the engine port locally.  The remote backend config supplier and the remote context backend are removed.

Manually tested with local and remote context against a Kloud instance acting as the bastion.

The deployment tool will be updated in a subsequent PR.

YES


cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -114,6 +109,11 @@ func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error
 	}
 	logrus.Infof("Context switched to '%s', Kurtosis engine will now be restarted", contextIdentifier)
 
+	engineManager, err := engine_manager.NewEngineManager(ctx)
```
laurentluce, line 122: Instantiate the engine manager after storing the new context so it can read it.

cli/cli/helpers/engine_manager/engine_existence_guarantor.go
```diff
@@ -46,8 +46,6 @@ type engineExistenceGuarantor struct {
 
 	engineServerKurtosisBackendConfigSupplier engine_server_launcher.KurtosisBackendConfigSupplier
 
-	kurtosisRemoteBackendConfigSupplier *engine_server_launcher.KurtosisRemoteBackendConfigSupplier
```
laurentluce, line 49: The remote backend config supplier is not needed anymore.

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
```diff
@@ -86,29 +78,29 @@ func (clusterConfig *KurtosisClusterConfig) GetClusterType() KurtosisClusterType
...
 				clusterId,
 				clusterType.String(),
 			)
 		}
 
 		backendSupplier = func(_ context.Context) (backend_interface.KurtosisBackend, error) {
-			kurtosisRemoteBackendConfigMaybe, err := kurtosisRemoteBackendConfigSupplier.GetOptionalRemoteConfig()
-			if err != nil {
-				return nil, stacktrace.Propagate(err, "Error building optional remote Kurtosis backend config")
+			var remoteBackendConfigMaybe *configs.KurtosisRemoteBackendConfig
```
laurentluce, line 96: This remote backend config is generated from the remote context and used by the CLI to connect to the bastion docker daemon.  The same object type is also used by the engine to connect to the remote backend docker daemon.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
```diff
@@ -30,6 +42,30 @@ type APIContainerModeArgs struct {
 	APIContainerIP net.IP
 }
 
+var (
+	NoAPIContainerModeArgs *APIContainerModeArgs = nil
+)
+
+func GetDockerKurtosisBackend(
```
laurentluce, line 52: This generic function creates a local or a remote docker backend based on the existence of a remote backend config.  It is the only function that should be called.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -101,6 +107,10 @@ func CreateEngine(
 		consts.DockerSocketFilepath: consts.DockerSocketFilepath,
 	}
 
+	if serverArgs.OnBastionHost {
+		bindMounts[consts.HostEngineConfigDirToMount] = consts.EngineConfigLocalDir
```
laurentluce, line 112: Mount the directory on the bastion host containing the remote backend config.

container-engine-lib/lib/backend_impls/remote_context_backend/backend_creator.go
laurentluce, line 1: The remote context backend is not needed anymore.

container-engine-lib/lib/backend_interface/objects/configs/remote_backend_config.go
```diff
@@ -0,0 +1,58 @@
...
+// TODO: Move to the backend interface dir to be more generic
+const (
+	urlScheme = "tcp"
+)
+
+var (
+	NoRemoteBackendConfig *KurtosisRemoteBackendConfig = nil
+)
+
+type KurtosisRemoteBackendConfig struct {
```
laurentluce, line 21: The remote backend config is backend agnostic with a generic endpoint.  It is currently only used with the remote docker backend.

container-engine-lib/lib/backend_interface/objects/configs/remote_backend_config.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+	endpoint := fmt.Sprintf("%s://%s:%d", urlScheme, remoteContext.GetHost(), remoteContext.GetKurtosisBackendPort())
+	return &KurtosisRemoteBackendConfig{
+		Endpoint: endpoint,
+		Tls:  tlsConfig,
+	}
+}
+
+func NewRemoteBackendConfigFromJSON(
```
laurentluce, line 51: This function is used to parse the remote backend config on the bastion host.

cli/cli/helpers/engine_manager/engine_existence_guarantor.go
```diff
@@ -65,6 +63,9 @@ type engineExistenceGuarantor struct {
 	//TODO This is a temporary hack we should remove it when centralized logs be implemented in the KubernetesBackend
 	kurtosisClusterType resolved_config.KurtosisClusterType
 
+	// Engine on bastion?
+	onBastionHost bool
```
laurentluce, line 67: Thinking more about it, I am wondering if I should supply this information using the `engine_server_launcher.KurtosisBackendConfigSupplier`.

cli/cli/command_framework/highlevel/engine_consuming_kurtosis_command/engine_consuming_kurtosis_command.go
```diff
@@ -147,12 +154,21 @@ func (cmd *EngineConsumingKurtosisCommand) getSetupFunc() func(context.Context)
...
-			if store.IsRemote(currentContext) && !portal_manager.NewPortalManager().IsReachable() {
-				return nil, stacktrace.NewError("Kurtosis is setup to use the remote context '%s' but Kurtosis "+
-					"Portal is unreachable for this context. Make sure Kurtosis Portal is running locally with 'kurtosis "+
-					"%s %s' and potentially 'kurtosis %s %s'. If it is, make sure the remote server is running and healthy",
-					currentContext.GetName(), command_str_consts.PortalCmdStr, command_str_consts.PortalStatusCmdStr,
-					command_str_consts.PortalCmdStr, command_str_consts.PortalStartCmdStr)
+			if store.IsRemote(currentContext) {
+				portalManager := portal_manager.NewPortalManager()
+				if !portalManager.IsReachable() {
+					return nil, stacktrace.NewError("Kurtosis is setup to use the remote context '%s' but Kurtosis "+
```
gbouv, line 155: I'm wondering if at some point we shouldn't just start the portal in the background, making it completely transparent to the user. It's literally a single line of here, and I've already hit this issue multiple times. It can become a bit annoying

cli/cli/command_framework/highlevel/engine_consuming_kurtosis_command/engine_consuming_kurtosis_command.go
```diff
@@ -147,12 +154,21 @@ func (cmd *EngineConsumingKurtosisCommand) getSetupFunc() func(context.Context)
...
+						"Portal is unreachable for this context. Make sure Kurtosis Portal is running locally with 'kurtosis "+
+						"%s %s' and potentially 'kurtosis %s %s'. If it is, make sure the remote server is running and healthy",
+						currentContext.GetName(), command_str_consts.PortalCmdStr, command_str_consts.PortalStatusCmdStr,
+						command_str_consts.PortalCmdStr, command_str_consts.PortalStartCmdStr)
+				}
+				// Forward the remote engine port to the local machine
+				portalClient := portalManager.GetClient()
+				forwardEnginePortArgs := portal_constructors.NewForwardPortArgs(uint32(kurtosis_context.DefaultGrpcEngineServerPortNum), uint32(kurtosis_context.DefaultGrpcEngineServerPortNum), &enginePortTransportProtocol)
```
gbouv: This is going to break if the engine was started on a custom port. That being said, maybe the fact that engine is always running on its default port is taken for granted in other places too. So maybe it's fine, but we're adding a bit more tech debts by doing that

cli/cli/commands/kurtosis_context/context_switch/switch.go
gbouv: Maybe add this as a comment in the code, cannot hurt spreading the context to others as well

cli/cli/helpers/engine_manager/engine_existence_guarantor.go
gbouv: Niiice 💪

cli/cli/helpers/engine_manager/engine_existence_guarantor.go
gbouv: That would make sense to me yes. Keeping all the backend config info at the same place is cleaner, and this boolean looks like a piece of config.
But thinking more about it, do we need this flag? From now, if the context is remote, then the engine will necessarily be on the bastion right? So, this flag to me is duplicate of kurtosisContext.IsRemote. Wondering if we could merge those together instead

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
gbouv: 👍  that makes sense to me, and it's nice we're able to use the same kind of object in those 2 places

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
```diff
@@ -86,29 +78,29 @@ func (clusterConfig *KurtosisClusterConfig) GetClusterType() KurtosisClusterType
...
+			currentContext, _ := store.GetContextsConfigStore().GetCurrentContext()
+			if currentContext != nil {
```
gbouv: Interesting pattern here, but I think go generally prefers doing something like:
```
currentContext, err := store.GetContextsConfigStore().GetCurrentContext()
if err != nil {
    handle the error
} else {
    // you can assume the result object (here `currentContext `) is not nil. 
}
```
And if the first `if err != nil` returns, the `else` is not necessary and you can un-indent the case of `currentContext != nil`

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
```diff
@@ -86,29 +78,29 @@ func (clusterConfig *KurtosisClusterConfig) GetClusterType() KurtosisClusterType
...
+				if store.IsRemote(currentContext) {
+					remoteBackendConfigMaybe = configs.NewRemoteBackendConfigFromRemoteContext(currentContext.GetRemoteContextV0())
+				}
 			}
-			backend, err := remote_context_backend.GetContextAwareKurtosisBackend(kurtosisRemoteBackendConfigMaybe, dockerBackendApiContainerModeArgs)
+			backend, err := backend_creator.GetDockerKurtosisBackend(backend_creator.NoAPIContainerModeArgs, remoteBackendConfigMaybe)
```
gbouv, line 106: Maybe add some comments to this call because there's a lot of things that are implicit. Why do we use `NoAPIContainerModeArgs`, and what will happen if `remoteBackendConfigMaybe` != nil VS == nil

container-engine-lib/README.md
```diff
@@ -1,6 +1,6 @@
 Container Engine Lib
 ====================
-This library abstracts away interactions with the container engine (be it Docker or Kubernetes) via a `KurtosisBackend` interface. Users should call `GetLocalDockerKurtosisBackend` (or the Kubernetes equivalent when it exists).
+This library abstracts away interactions with the container engine (be it Docker or Kubernetes) via a `KurtosisBackend` interface. Users should call `GetDockerKurtosisBackend` (or the Kubernetes equivalent when it exists).
```
gbouv: I think `(or the Kubernetes equivalent when it exists)` is a reminiscence of when kubernetes was a backend plugin. 
And quickly searching the code I can't find an `GetKubernetesDockerBackend` anywhere, so what is this "equivalent" we're talking about here? Since you're touching this line, might be worth cleaning it up a little bit

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
gbouv: Same here, this is super interesting information, let's put it as a comment in the code directly

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
```diff
@@ -30,24 +42,123 @@ type APIContainerModeArgs struct {
 	APIContainerIP net.IP
 }
 
-// GetLocalDockerKurtosisBackend is the entrypoint method we expect users of container-engine-lib to call
+var (
+	NoAPIContainerModeArgs *APIContainerModeArgs = nil
```
gbouv, line 46: This seems unused in this file. Let's move it to the file where it's used and make it private

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
```diff
@@ -30,24 +42,123 @@ type APIContainerModeArgs struct {
...
+		}
+		defer cleanCertFilesFunc()
+		tlsOpt := client.WithTLSClientConfig(
+			path.Join(tlsFilesDir, caFileName),
+			path.Join(tlsFilesDir, certFileName),
+			path.Join(tlsFilesDir, keyFileName))
+		clientOptions = append(clientOptions, tlsOpt)
+	}
+
+	// API version negotiation option
```
gbouv: ... and Timeout here? or move the timeout in the above append

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
```diff
@@ -30,24 +42,123 @@ type APIContainerModeArgs struct {
...
+	if err = os.WriteFile(keyAbsFileName, key, tlsFilesPerm); err != nil {
+		return "", nil, stacktrace.Propagate(err, "Error writing content of key to temporary file at '%s'", keyAbsFileName)
+	}
+
+	cleanDirectoryFunc := func() {
+		if err = os.RemoveAll(tempDirectory); err != nil {
+			logrus.Warnf("Error removing TLS config directory at '%s'. Will remain in the OS temporary files folder until the OS removes it", tempDirectory)
+		}
+	}
+	return tempDirectory, cleanDirectoryFunc, nil
```
gbouv, line 159: This code looks strangely familiar to me. Was it copied or moved from somewhere? If it's used by multiple things, maybe we can put it in a shared lib or something to DRY this out a bit

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/consts/consts.go
```diff
@@ -21,6 +21,10 @@ const (
 	// This needs to be bind-mounted into the engine & API containers so they can manipulate Docker
 	DockerSocketFilepath = "/var/run/docker.sock"
 
+	// The host engine config directory to mount and its local mapping
+	HostEngineConfigDirToMount = "/root/engine"
+	EngineConfigLocalDir       = "/run/engine"
```
gbouv, line 26: nit: Maybe have more explicit directory name. I don't care too much about the directory inside the engine container itself, but for the one in the bastion, let's be more explicit with "engine_config" or something.
Even wondering if we should put it in a /root/.kurtosis/data/ hidden folder or something.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
gbouv: Same here, let's put that context in the code with a comment, explaining why we need this specific mount when we're on the bastion

container-engine-lib/lib/backend_interface/objects/configs/remote_backend_config.go
gbouv: in code

container-engine-lib/lib/backend_interface/objects/configs/remote_backend_config.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	contexts_config_store_generated_api "github.com/kurtosis-tech/kurtosis/contexts-config-store/api/golang/generated"
+	"github.com/kurtosis-tech/stacktrace"
+)
+
+// TODO: Move to the backend interface dir to be more generic
```
gbouv, line 11: You mean the const? I'd say if it's used only here, let's keep it close to its consumer, and private.

container-engine-lib/lib/backend_interface/objects/configs/remote_backend_config.go
```diff
@@ -0,0 +1,58 @@
...
+const (
+	urlScheme = "tcp"
+)
+
+var (
+	NoRemoteBackendConfig *KurtosisRemoteBackendConfig = nil
```
gbouv, line 17: Same, do we need that public? It's just a nil, doesn't take a bunch of memory, if we need it somewhere else, maybe we can just re-create it there

container-engine-lib/lib/backend_interface/objects/configs/remote_backend_config.go
gbouv: in code.

engine/server/engine/main.go
```diff
@@ -107,9 +110,21 @@ func runMain() error {
...
 	}
 
-	remoteContextBackendConfig := serverArgs.KurtosisRemoteBackendConfig
+	var remoteBackendConfigMaybe *configs.KurtosisRemoteBackendConfig
+	if serverArgs.OnBastionHost {
+		// Read remote backend config from the local filesystem
+		remoteBackendConfigPath := filepath.Join(consts.EngineConfigLocalDir, remoteBackendConfigFilename)
+		remoteBackendConfigBytes, err := os.ReadFile(remoteBackendConfigPath)
+		if err != nil {
+			return stacktrace.Propagate(err, "The remote backend config cannot be read")
```
gbouv: nit: maybe add a bit more context to the error message here. Where it's expected to be found, what format it should have, etc.

gbouv (approved):
This matches exactly what I had in mind, love it!

cli/cli/helpers/engine_manager/engine_existence_guarantor.go
laurentluce: This config supplier returns a config which gets mapped to a local backend config object which made things confusing for the bastion host flag.  I kept things as is for now.

cli/cli/command_framework/highlevel/engine_consuming_kurtosis_command/engine_consuming_kurtosis_command.go
laurentluce: Added an issue #970 to track.

cli/cli/command_framework/highlevel/engine_consuming_kurtosis_command/engine_consuming_kurtosis_command.go
laurentluce: The get engine status function of the engine manager makes the assumption and I saw it in one other place.  There is a TODO to change that.  The same const is used though which is good.

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
laurentluce: I had copied from somewhere else.  I agree with you that the pattern lacked consistency with the rest of the code and I changed it.

container-engine-lib/README.md
laurentluce: This is the asymmetry I was referring to last week with no `backend_creator` for K8S but there are similar functions of the form `New*KubernetesKurtosisBackend` which can be used for the K8S backend.  I updated the README.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
laurentluce: It is used in two other files and I wanted to add it to this file since it is used with a function in this file.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
laurentluce: I moved some of this code from the remote context backend.

container-engine-lib/lib/backend_interface/objects/configs/remote_backend_config.go
laurentluce: I had done this TODO so I removed it.

container-engine-lib/lib/backend_interface/objects/configs/remote_backend_config.go
laurentluce: I put it here because it applies to the remote backend config and it is used in couple of files.

PR #967, opened 2023-07-24
feat: folks can now use frontend to view files and it's contents.

Peeeekay:
Once this PR is merged, users can use frontend to view contents of the file artifacts. The content itself is not formatted at the moment but things like yaml formatter and json formatter can be added later on.


engine/frontend/src/api/container.js
```diff
@@ -21,6 +21,62 @@ const getDataFromApiContainer = async (request, process) => {
...
+            const responseFromGrpc = await containerClient.inspectFilesArtifactContents(inspectFileArtifactReq, {})
+            return responseFromGrpc.toObject()
+        } catch (error) {
+            return {
+                files:[]
+            }
+        }
+    }
+
+    const processFileArtifact = (data) => {
```
Peeeekay, line 41: this potentially can be done on the backend - @victorcolombo 
What I am doing is - i am converting the list of files to look like
```
{
  dir : {
     sub_dir1: {
            sub_dir2: {
                    file_name: file
                }
        }
  }
}
```

This allows me to show file structure; that is i will only the sub_dir1 on the ui - only after user clicks the parent folder.

victorcolombo: Good call, do you mind create a ticket for that?

victorcolombo (approved):

PR #968, opened 2023-07-24
feat: Implement V0 of docker import

victorcolombo:
## Description:

This PR implements the `kurtosis import` command in the CLI, that parses a very simple Docker Compose into Starlark and Runs it. Feature set is limited to:
- Image
- Ports
- Entrypoint / Command

Notably does not include:
- Volumes
- Env Vars
YES


cli/cli/commands/import/import.go
```diff
@@ -0,0 +1,144 @@
...
+		serviceStarlarks[serviceConfig.Name] = starlarkConfig
+	}
+	script := "def run(plan):\n"
+	for serviceName, serviceConfig := range serviceStarlarks {
+		script += fmt.Sprintf("\tplan.add_service(name = '%s', config = %s)\n", serviceName, serviceConfig)
+	}
+	return script, nil
+}
+
+func runStarlark(ctx context.Context, starlarkScript string) error {
```
h4ck3rk3y: perhaps we can refactor this later

cli/cli/commands/import/import.go
```diff
@@ -0,0 +1,144 @@
...
+	kurtosisCtx, err := kurtosis_context.NewKurtosisContextFromLocalEngine()
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred connecting to the local Kurtosis engine")
+	}
+
+	enclaveCtx, err := kurtosisCtx.CreateEnclave(ctx, "docker-compose-import", false)
```
h4ck3rk3y: is this a prefix? If you look at `run` you can see how a new one is generated

cli/cli/commands/import/import.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred creating an enclave '%v'", "docker-compose-import")
```
h4ck3rk3y: let's change this when we change the above

cli/cli/commands/import/import.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	starlarkRunResult, err := enclaveCtx.RunStarlarkScriptBlocking(ctx, "", starlarkScript, "{}", false, 1, []kurtosis_core_rpc_api_bindings.KurtosisFeatureFlag{})
```
h4ck3rk3y: lets put these constants in variables

cli/cli/commands/import/import.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return stacktrace.Propagate(err, "An error has occurred when running Starlark to add service")
+	}
+	logrus.Infof("Enclave was built with following output: %s", starlarkRunResult.RunOutput)
```
h4ck3rk3y: can we call inspect at the end like we do from `run`

cli/cli/commands/import/import.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return "", stacktrace.Propagate(err, "Error translating docker compose to Starlark")
+	}
+	return script, nil
+}
+
+// TODO(victor.colombo): Have a better UX letting people know ports have been remapped
+func convertComposeProjectToStarlark(compose *types.Project) (string, error) {
+	serviceStarlarks := map[string]string{}
+	for _, serviceConfig := range compose.Services {
```
h4ck3rk3y, line 128: perhaps we can track TODO for supporting `depends` later on

h4ck3rk3y (approved):

cli/cli/commands/import/import.go
victorcolombo: I'll add a TODO

cli/cli/commands/import/import.go
victorcolombo: Added a flag to autogenerate/set enclave name

cli/cli/commands/import/import.go
victorcolombo: Done

cli/cli/commands/import/import.go
victorcolombo: Added enclave name to facilitate ls later

cli/cli/commands/import/import.go
victorcolombo: Done

PR #987, opened 2023-07-26
fix: Fix error swallowing in DefaultServiceNetwork.destroyService

### LINKED ISSUE #979: `destroyService` swallows error messages
### What's your CLI version?

0.80.15

destroyService: https://github.com/kurtosis-tech/kurtosis/blob/main/core/server/api_container/server/service_network/default_service_network.go#L1438

There are multiple cases where an error later overwrites the earlier error

### Desired behavior

All the errors should be printed (and I think we have a helper for composing multiple errors together)

gbouv:
## Description:
Fixes #979 

NO


core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1445,24 +1444,31 @@ func (network *DefaultServiceNetwork) destroyService(ctx context.Context, servic
 		},
 		Statuses: nil,
 	}
-	_, failedToDestroyUuids, err := network.kurtosisBackend.DestroyUserServices(context.Background(), network.enclaveUuid, userServiceFilters)
+	successfullyDestroyedUuids, failedToDestroyUuids, err := network.kurtosisBackend.DestroyUserServices(context.Background(), network.enclaveUuid, userServiceFilters)
 	if err != nil {
-		errorResult = stacktrace.Propagate(err, "Attempted to destroy the services with UUID '%v' but had no success. You must manually destroy the services. Kurtosis will now try to remove its sidecar if it exists but might it fail as well.", serviceUuid)
+		return stacktrace.Propagate(err, "Attempted to destroy the services with UUID '%v' but had no success. You must manually destroy the services. Kurtosis will now try to remove its sidecar if it exists but might it fail as well.", serviceUuid)
```
h4ck3rk3y, line 1449: in this case we will not remove the sidecar as we are exiting early

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1445,24 +1444,31 @@ func (network *DefaultServiceNetwork) destroyService(ctx context.Context, servic
...
 	}
-	if failedToDestroyErr, found := failedToDestroyUuids[serviceUuid]; found {
-		errorResult = stacktrace.Propagate(failedToDestroyErr, "Attempted to destroy the services with UUID '%v' but had no success. You must manually destroy the services. Kurtosis will now try to remove its sidecar if it exists but might it fail as well.", serviceUuid)
+	if _, found := successfullyDestroyedUuids[serviceUuid]; found {
+		if len(failedToDestroyUuids) != 0 {
+			return stacktrace.NewError("Service '%s' was successfully destroyed but unexpected error were returned by Kurtosis backend. This is unexpected and should be investigated. Here are the errors:\n%v", serviceUuid, failedToDestroyUuids)
+		}
+	} else {
+		if failedToDestroyErr, found := failedToDestroyUuids[serviceUuid]; found {
+			return stacktrace.Propagate(failedToDestroyErr, "Attempted to destroy the services with UUID '%v' but had no success. You must manually destroy the services. Kurtosis will now try to remove its sidecar if it exists but might it fail as well.", serviceUuid)
```
h4ck3rk3y, line 1457: in this case we will not remove the sidecar as we are exiting early

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1445,24 +1444,31 @@ func (network *DefaultServiceNetwork) destroyService(ctx context.Context, servic
...
 
 	// deleting the sidecar
 	networkingSidecar, found := network.networkingSidecars[serviceName]
 	if found {
-		delete(network.networkingSidecars, serviceName)
-		err = network.networkingSidecarManager.Remove(ctx, networkingSidecar)
-		if errorResult == nil && err != nil {
-			errorResult = stacktrace.Propagate(err, "Attempted to clean up the sidecar for service with name '%s' but an error occurred.", serviceName)
+		if err = network.networkingSidecarManager.Remove(ctx, networkingSidecar); err != nil {
+			return stacktrace.Propagate(err, "Service '%s' was successfully detroyed byt an error occurred cleaning up its sidecar. The sidecar must be deleted manually.", serviceName)
```
h4ck3rk3y: s/byt/bus/

h4ck3rk3y (approved):

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1445,24 +1444,31 @@ func (network *DefaultServiceNetwork) destroyService(ctx context.Context, servic
...
+		return stacktrace.Propagate(err, "Attempted to destroy the services with UUID '%v' but had no success. You must manually destroy the services. Kurtosis will now try to remove its sidecar if it exists but might it fail as well.", serviceUuid)
 	}
-	if failedToDestroyErr, found := failedToDestroyUuids[serviceUuid]; found {
-		errorResult = stacktrace.Propagate(failedToDestroyErr, "Attempted to destroy the services with UUID '%v' but had no success. You must manually destroy the services. Kurtosis will now try to remove its sidecar if it exists but might it fail as well.", serviceUuid)
+	if _, found := successfullyDestroyedUuids[serviceUuid]; found {
+		if len(failedToDestroyUuids) != 0 {
+			return stacktrace.NewError("Service '%s' was successfully destroyed but unexpected error were returned by Kurtosis backend. This is unexpected and should be investigated. Here are the errors:\n%v", serviceUuid, failedToDestroyUuids)
+		}
+	} else {
+		if failedToDestroyErr, found := failedToDestroyUuids[serviceUuid]; found {
```
h4ck3rk3y, line 1456: randomly being paranoid; but if the failedToDestroyErr is nil then the propagate at line 1457 will panic.

https://go.dev/play/p/pHL11xwBzE1

^ just confirming whether a dictionary can map to a nil; it can.

core/server/api_container/server/service_network/default_service_network.go
gbouv: But since we did not succeed in removing the service, isn't this the expected behaviour?

core/server/api_container/server/service_network/default_service_network.go
gbouv: Wait... But if found is true, it means there is a value associated to it. This is the contract for `DestroyUserServices`, and if this function can return an error map with a key associated with `nil`, then we should fix this function rather than handling it here; I checked and it cannot

core/server/api_container/server/service_network/default_service_network.go
gbouv: Same remark as above, since we failed to remove the service, I think it is expected for kurtosis to _not_ remove the sidecar.

core/server/api_container/server/service_network/default_service_network.go
h4ck3rk3y: This is expected behaviour yes. But we should change the message here and below. That was the intent of my comment.

PR #988, opened 2023-07-26
feat: validate min cpu & min memory are well under whats available

h4ck3rk3y:
## Description:
This is a very best effort implementation. For docker we get a list of all running containers, and sum them; if one of the containers in the list dies we ignore them.

YES/NO


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -496,6 +500,14 @@ func (backend *DockerKurtosisBackend) DestroyDeprecatedCentralizedLogsResources(
 	return nil
 }
 
+func (backend *DockerKurtosisBackend) GetAvailableCPUAndMemory(ctx context.Context) (uint64, bool, uint64, bool, error) {
```
victorcolombo: can we create types so we don't have to deduce what these numbers means?

`type CpuMilli uint64`

h4ck3rk3y: Done @victorcolombo !

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -496,6 +501,14 @@ func (backend *DockerKurtosisBackend) DestroyDeprecatedCentralizedLogsResources(
 	return nil
 }
 
+func (backend *DockerKurtosisBackend) GetAvailableCPUAndMemory(ctx context.Context) (compute_resources.MemoryInMegaBytes, bool, compute_resources.CpuMilliCores, bool, error) {
```
victorcolombo: why we return `resourceInformationIsComplete` two times?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -1065,6 +1070,16 @@ func (manager *DockerManager) CopyFromContainer(ctx context.Context, containerId
 	return tarStreamReadCloser, nil
 }
 
+// GetAvailableCPUAndMemory returns free memory in megabytes, free cpu in millicores, information on whether cpu information is complete
```
victorcolombo, line 1074: use the same types you defined for the kurtosis backend?

victorcolombo (approved):

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
h4ck3rk3y: fixed :)

PR #999, opened 2023-07-27
fix: Log streaming was timing out on docker

### LINKED ISSUE #969: Looking at logs for a long time via kurtosis seems to time out every now and then though.
### What's your CLI version?

Unknown

Reported by @barnabasbusa on #934 

### Desired behavior

Users should be able to look at logs for Kurtosis for at least a day or more

gbouv:
Fixes #969 

## Description:
The docker client timeout comes into play even for long running queries like streaming logs or downloading large images. Even though we receive messages from docker engine, the timeout is not reset... :( This change uses 2 docker clients, one with and one without a timeout such that for long running queries do not fail in Kurtosis.

NO


container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -828,6 +847,12 @@ func (manager *DockerManager) GetContainerLogs(
 	containerId string,
 	shouldFollowLogs bool,
 ) (io.ReadCloser, error) {
+	// As we're using the docker client with no timeout to be able to follow the logs for a long time, we quickly check
+	// with the client that has  a timeout whether the docker engine is reachable.
```
leoporoli, line 857: nice!!!!! smart decision

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -1095,9 +1120,14 @@ func (manager *DockerManager) isImageAvailableLocally(ctx context.Context, image
 	return numMatchingImages > 0, nil
 }
 
-func (manager *DockerManager) pullImage(context context.Context, imageName string) (err error) {
+func (manager *DockerManager) pullImage(context context.Context, imageName string) error {
+	// As we're using the docker client with no timeout to pull the image, we quickly check with the client that has
+	// a timeout whether the docker engine is reachable.
+	if _, err := manager.dockerClient.Ping(context); err != nil {
```
leoporoli, line 1142: not important to do now, but I wonder how we could force this call always before any call with the Docker client without a timeout, a wrapper maybe? but not important to do now at least you have a quick and easy solution

leoporoli (approved):

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
gbouv: Yes that would be nice, though I can't think a good simple implementation for such a wrapper, b/c docker Client object is a struct, not an interface. We could do it at the docker_manager level though. Maybe if at some point we end up with a lot of endpoints using the NoTimeout client, for now as it is only two maybe it's fine to keep it simple

PR #1001, opened 2023-07-27
feat: ask user for email on first run of Kurtosis

### LINKED ISSUE #895: [FR] Add an optional email-share flow/option upon Kurtosis install and upgrade in the CLI
### Background & motivation

We want to curate an email list for us to push content to engaged Kurtosis users with. I'd like to have this in the product itself. 

### Desired behaviour

This Github issue tracks two related things (the first one is more important):
1.  Add functionality such that when someone installs Kurtosis for the first time, they are prompted to optionally share their email with us
2. Upon upgrade at a certain version (that we mandate one-time, or maybe periodically), we prompt the user to share their email (optionally)

h4ck3rk3y:


h4ck3rk3y:
https://github.com/kurtosis-tech/kurtosis/assets/6670312/d371e60f-17b7-4e11-a29f-3012bfc3b7a0

cli/cli/helpers/prompt_displayer/prompt_displayer.go
```diff
@@ -75,40 +46,9 @@ func DisplayConfirmationPromptAndGetBooleanResult(label string, defaultValue boo
...
-	var validInputsSliceStr []string
-
-	for _, validInput := range validUserInputsSlice {
-		validInputsSliceStr = append(validInputsSliceStr, string(validInput))
+	if input == "" {
+		return nil
 	}
-	validInputsListStr := strings.Join(validInputsSliceStr, `','`)
-	return validInputsListStr
+	_, parsingErr := mail.ParseAddress(input)
```
leoporoli, line 52: nit: missing stack trace here

cli/cli/kurtosis_config/kurtosis_config_provider.go
```diff
@@ -35,6 +36,7 @@ func (configProvider *KurtosisConfigProvider) GetOrInitializeConfig() (*resolved
 		if err != nil {
 			return nil, stacktrace.Propagate(err, "An error occurred executing init interactive config")
 		}
+		email_collector.AskUserForEmailAndLogIt()
```
leoporoli, line 39: I don't have much context but are we going to skip the prompt from current users? is it the idea?

cli/cli/kurtosis_config/email_collector/email_collector.go
```diff
@@ -0,0 +1,68 @@
...
+		logrus.Debugf("an error occurred while getting users metrics id:\n%v", err)
+		return
+	}
+	logger := logrus.StandardLogger()
+
+	metricsClient, metricsClientCloseFunc, err := metrics_client.CreateMetricsClient(
+		source.KurtosisCLISource,
+		kurtosis_version.KurtosisVersion,
+		metricsUserId,
+		resolved_config.DefaultDockerClusterName,
```
leoporoli, line 50: Are you using Docker here because this info is not relevant for this metric? if it is the case, could we add a TODO, and then letter we could create a custom enum for this with a default value, something like "not-relevant" 
or could we pass the right cluster type here?

leoporoli (approved):
I added some comments, one of them is more of a question because I don't have much context from the product side about if we want to collect emails for current users or not

cli/cli/kurtosis_config/kurtosis_config_provider.go
h4ck3rk3y: Yes!

cli/cli/kurtosis_config/email_collector/email_collector.go
h4ck3rk3y: I will add a comment. We are doing docker because at this point there isn't a metrics configuration; this is only happening at first run. I will add comment.

Issue #1007, opened 2023-07-27
Enable linters to preemptly catch bugs and reduce reviewer effort


victorcolombo:
### Background & motivation

Currently we rely on programmer and reviewer implementing according to both our style guide and some tribal knowledge.

### Desired behaviour

We should rely more on https://github.com/golangci/golangci-lint

### How important is this to you?

Painful; the lack of this feature makes using Kurtosis frictionful.

victorcolombo:
Linters tracked:
- [X] [gomnd](https://github.com/tommy-muehle/go-mnd)
- [ ] [nestif](https://golangci-lint.run/usage/linters/#nestif)
- [ ] [nilerr](https://github.com/gostaticanalysis/nilerr)
- [ ] [nonamedreturns](https://github.com/firefart/nonamedreturns)
- [X] exportloopref
- [X] staticcheck
- [X] exhaustruct

Other tasks:
- [x] add linter script for git pre-push
- [x] add Golang Lint on Goland instructions
- [x] remove exceptions to linter rules

h4ck3rk3y:
@victorcolombo where are you with the remaining 3?

PR #1008, opened 2023-07-27
refactor: Enable magic constant linter

victorcolombo:
NO

Part of https://github.com/kurtosis-tech/kurtosis/issues/1007


cli/cli/commands/enclave/inspect/inspect.go
```diff
@@ -166,8 +166,8 @@ func PrintEnclaveInspect(ctx context.Context, kurtosisBackend backend_interface.
 			return stacktrace.NewError("No printing function found for enclave object '%v'; this is a bug in Kurtosis!", header)
 		}
 
-		numRunesInHeader := utf8.RuneCountInString(header) + 2 // 2 because there will be a space before and after the header
-		numPadChars := (headerWidthChars - numRunesInHeader) / 2
+		numRunesInHeader := utf8.RuneLen(' ') + utf8.RuneCountInString(header) + utf8.RuneLen(' ') // there will be a space before and after the header
+		numPadChars := (headerWidthChars - numRunesInHeader) / 2                                   //nolint:gomnd
```
Peeeekay, line 170: why are we not using lint here?

container-engine-lib/lib/network_helpers/network_helpers.go
```diff
@@ -27,8 +29,8 @@ func GetFreeIpAddrFromSubnet(takenIps map[string]bool, subnet *net.IPNet) (net.I
 	// The IP can be either 4 bytes or 16 bytes long; we need to handle both!
 	// See https://gist.github.com/ammario/649d4c0da650162efd404af23e25b86b
 	var intIp uint32
-	if len(subnetIp) == 16 {
-		intIp = binary.BigEndian.Uint32(subnetIp[12:16])
+	if len(subnetIp) == net.IPv6len {
+		intIp = binary.BigEndian.Uint32(subnetIp[12:net.IPv6len])
```
Peeeekay: we can use constant for `12` as well.

core/server/api_container/server/startosis_engine/builtins/kurtosis_module.go
```diff
@@ -15,6 +15,11 @@ const (
 	connectionSubmoduleAllowedAttrName = "ALLOWED"
 )
 
+var (
+	totalPacketLossPercentage = starlark.Float(100) //nolint:gomnd
+	noPacketLossPercentage    = starlark.Float(0)
```
Peeeekay: possibly nit:

packageLostPercentage = 100

and in the code we could just do starlark.Float(packageLostPercentage)

also why don't we need `//nolint` on line `20`

Peeeekay (approved):

container-engine-lib/lib/network_helpers/network_helpers.go
victorcolombo: Done!

cli/cli/commands/enclave/inspect/inspect.go
victorcolombo: no good way to define this two. It's clear that we are averaging and defining `const consantToTakeAverage = 2` seems overkill :)

core/server/api_container/server/startosis_engine/builtins/kurtosis_module.go
victorcolombo: Done!

PR #1010, opened 2023-07-27
feat: Add starlark converter to kurtosis import

victorcolombo:
## Description:

This PR adds the `kurtosis import` CLI command, that brings in your existing `docker-compose.yml` into a Kurtosis enclave by going a Docker Compose -> Starlark translation.

YES


cli/cli/commands/import/import.go
```diff
@@ -110,17 +133,25 @@ func run(
 		return stacktrace.Propagate(err, "Failed to convert compose to starlark")
 	}
 	// TODO(victor.colombo): Make this as pretty as run is
+	if convertOnly {
+		fileBase := filepath.Base(path)
+		fileName := fmt.Sprintf("%s.star", strings.TrimSuffix(fileBase, filepath.Ext(fileBase)))
+		err := os.WriteFile(fileName, []byte(script), readWriteEveryone)
+		if err != nil {
```
h4ck3rk3y: perhaps inline this with line 139

h4ck3rk3y (approved):

cli/cli/commands/import/import.go
victorcolombo: Good catch, done

PR #1011, opened 2023-07-27
fix: Check if a local engine is running before switching to a remote context and let the user know what to do

laurentluce:
## Description:
Tell the user if a local engine is running and he is trying to switch to a remote context since it would conflict with the forwarded remote engine port.  Move the engine port forwarding from the start of the context switch to after the engine is started so we don't forward too early.

YES


adschwartz (approved):

cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -57,6 +56,21 @@ func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error
...
+		engineManager, err := engine_manager.NewEngineManager(ctx)
+		if err != nil {
+			return stacktrace.Propagate(err, "An error occurred creating an engine manager.")
+		}
+		engineStatus, _, _, err := engineManager.GetEngineStatus(ctx)
+		if err != nil {
+			return stacktrace.Propagate(err, "An error occurred retrieving the engine status.")
+		}
+		if engineStatus == engine_manager.EngineStatus_Running {
+			logrus.Infof("An engine is running locally. Stop the engine before switching context so the remote engine port can be forwarded locally.")
```
adschwartz: I would rephrase this as:
Prior to switching context, stop the local engine by running `kurtosis engine stop`

PR #1013, opened 2023-07-27
fix: stream instruction output in chunks of 3Mb

h4ck3rk3y:
## Description:
Previously we would try to send all of it at once as the max message size is inifnite. User would get something like - 

```
Caused by: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (17391758 vs. 4194304)
```

Barnabas ran into it. I ran into it several times mostly with execs.

YES


gbouv:
@tedim52 this might be something we want to take in consideration for your instruction "output streaming" feature you're working on.

h4ck3rk3y:
@gbouv this is still in progress. need to fix a few new lines. Will tag you now but review when I ping y ou!

PR #1015, opened 2023-07-28
feat: add `cloud add`

adschwartz:
## Description:
With this change users can create a new cloud instance, by calling `kurtosis cloud add`. Callers must define the `KURTOSIS_CLOUD_API_KEY` env var.

YES


cli/cli/commands/cloud/add/add.go
```diff
@@ -0,0 +1,73 @@
...
+		UserId: metricsUserId,
+	}
+	result, err := client.CreateCloudInstance(ctx, getConfigArgs)
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred while calling the Kurtosis Cloud API")
+	}
+
+	instanceId := result.GetInstanceId()
+	logrus.Infof("Success! The Kurtosis Cloud instance is being created with id: %s", instanceId)
+	logrus.Infof("The Kurtosis cloud instance is currently being created and it should take about 2-3 mins to become ready. "+
```
laurentluce, line 66: How does the user know that the cloud instance is ready?

laurentluce (approved):
It is really nice to see the user flow coming together.

cli/cli/commands/cloud/add/add.go
adschwartz: Yeah, they don't yet. I will need to add something here to help them. It could be an interactive session where we poll the endpoint or even better push the information down. I haven't yet fully ironed that out but we can add it in later iterations.

PR #1022, opened 2023-07-28
fix: truncate output if greater than 64*1024 characters

### LINKED ISSUE #1019: kurtosis run breaks if the server returns more than 4mb of data
### What's your CLI version?

0.80.12

def run(plan):
    plan.run_python(
        run = """
print("hello world"*2**20)
"""
    )

If you run the above you will get the error. Barnabas ran into this.

### Desired behavior

Kurtosis shouldn't break.

h4ck3rk3y:
## Description:
We would written a lot of data and the client would fail as it wouldn't accept over > 4mb.

going with this limit instead of something lower as a lot of our tests rely on output value matching.

### DIFF
--- a/core/server/api_container/server/startosis_engine/startosis_executor.go
+++ b/core/server/api_container/server/startosis_engine/startosis_executor.go
@@ -16,8 +16,9 @@ import (
 const (
 	progressMsg      = "Execution in progress"
 	ParallelismParam = "PARALLELISM"
-	// we limit the output to 300 characters
-	outputSizeLimit          = 300
+	// we limit the output to 600 characters
+	// TODO(tedi) get rid of this in favor of streaming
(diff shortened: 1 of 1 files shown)


core/server/api_container/server/startosis_engine/startosis_executor.go
```diff
@@ -15,6 +16,10 @@ import (
 const (
 	progressMsg      = "Execution in progress"
 	ParallelismParam = "PARALLELISM"
+	// we limit the output to 600 characters
+	// TODO(tedi) get rid of this in favor of streaming
```
tedim52, line 20: been wanting my name in the codebase, thanks gyani

tedim52 (approved):
LGTM

PR #1023, opened 2023-07-28
feat: Support resource reservations on Docker compose import

victorcolombo:
## Description:

This PR adds support to converting `reservations: cpu` field in Docker Compose into `min_cpu` in ServiceConfig. After this change, users will be able to leverage Starlark Validator to get warnings about if CPU/Memory use.

YES


cli/cli/commands/import/import.go
```diff
@@ -241,6 +251,36 @@ func convertComposeProjectToStarlark(compose *types.Project) (string, map[string
 	return script, requiredFileUploads, nil
 }
 
+func getMemoryMegabytesReservation(deployConfig *types.DeployConfig) int {
+	if deployConfig == nil {
+		return 0
+	}
+	reservation := 0
+	if deployConfig.Resources.Reservations != nil {
```
h4ck3rk3y, line 259: is `Resources` bound to be set or can `Resources` be nil?

cli/cli/commands/import/import.go
```diff
@@ -241,6 +251,36 @@ func convertComposeProjectToStarlark(compose *types.Project) (string, map[string
...
+	}
+	return reservation
+}
+
+func getMilliCpusReservation(deployConfig *types.DeployConfig) int {
+	if deployConfig == nil {
+		return 0
+	}
+	reservation := 0
+	if deployConfig.Resources.Reservations != nil {
```
h4ck3rk3y, line 271: similar is resources bound to be set?
if resources is set is nanocpus definitely set? (probably as its a float?)

h4ck3rk3y (approved):

cli/cli/commands/import/import.go
h4ck3rk3y: lets add a test case for this?

victorcolombo: It is a value not a pointer, so per type system it must be set!

PR #1025, opened 2023-07-28
fix: cluster set doesnt restart engine and doesnt set blank cluster-setting

### LINKED ISSUE #1020: cluster set sets "" in case of failure if there was also a failure in retrieving prior config
### What's your CLI version?

All versions till date

I don't have exact steps here.

### Desired behavior

Sets to a non "" cluster; say a default cluster like Docker

### LINKED ISSUE #1021: cluster set restarts the engine and this fails if there is already an engine running created by a different user on k8s
### What's your CLI version?

0.80.12

have two instances, with two different Kubernetes users; create engine with one. do cluster set on the other; this leads to a restart and the restart fails as the engine is already running.

### Desired behavior

Figure out why we do restarts and if we can skip it all together. Or fix this behavior so that the user switch and multi user k8s workflow works

h4ck3rk3y:
## Description:
Supports multi user k8s flow


cli/cli/commands/cluster/set/set.go
```diff
@@ -60,50 +59,25 @@ func run(ctx context.Context, flags *flags.ParsedFlags, args *args.ParsedArgs) e
...
-	}
-
 	if err = clusterSettingStore.SetClusterSetting(clusterName); err != nil {
 		return stacktrace.Propagate(err, "Failed to set cluster name to '%v'.", clusterName)
 	}
 	defer func() {
 		if clusterUpdateSuccessful {
 			return
 		}
+		if clusterPriorToUpdate == noClusterSetting {
```
leoporoli, line 69: I think that this defer doesn't make any sense anymore, we could leave it if we can prevent someone else forget to add it if they add new lines after `clusterSettingStore.SetClusterSetting` but we could also remove it for making the code simpler, I leave it to you.

leoporoli (approved):

gbouv:
After discussing it with the team, I think what we actually want is #1040 . Closing this one in favour of the other one

PR #1028, opened 2023-07-28
feat: Persistent service directories

gbouv:
## Description:
This PR adds the `Directory` object to Starlark to be used in the `ServiceConfig`'s `files` argument. Services can now have directories that persists over service restarts by doing:
```
plan.add_service(
    ...
    config=ServiceConfig(
        ...
        files={
            "/path/to/directory": Directory(
                persistent_key="directory-persistent-key",
            )
        }
        ...
    )
)
```
Persistent directories are owned by a single service. They cannot be shared across multiple services. Kurtosis keep them as long as the enclave lives. Re-using the same service name and persistent key will return the same (persisted) directory over and over again.

Note: previous usage of the `files` dictionary in `ServiceConfig` for file artifacts is still supported, but should be migrated to use the new `Directory` object:
```
plan.add_service(
    ...
    config=ServiceConfig(
        ...
        files={
            "/path/to/file/artifact": Directory(
                artifact_name=my_artifact_name,
            )
        }
        ...
    )
)
```

Docs PR will follow

YES


core/server/api_container/server/startosis_engine/kurtosis_types/directory/directory.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/kurtosis-tech/kurtosis/core/server/api_container/server/startosis_engine/startosis_errors"
+	"go.starlark.net/starlark"
+)
+
+const (
+	DirectoryTypeName = "Directory"
+
+	ArtifactNameAttr  = "artifact_name"
+	PersistentKeyAttr = "persistent_key"
+	DescriptionAttr   = "desc"
```
leoporoli: is it planned to use this constant later? because now it's not used

leoporoli (approved):

core/server/api_container/server/startosis_engine/kurtosis_types/directory/directory.go
gbouv: Ah yes, it was intended to be used but we don't now. I'll remove, thanks for flagging!

PR #1030, opened 2023-07-31
fix: Remove the temp cert files only after the docker client is initialized

laurentluce:
## Description:
The temp cert files were removed before the docker client got a chance to read them to build a TLS config.  Not sure if this was due to a race condition or a recent change in the docker client.

NO


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
```diff
@@ -86,7 +86,10 @@ func getRemoteDockerKurtosisBackend(
 	optionalApiContainerModeArgs *APIContainerModeArgs,
 	remoteBackendConfig *configs.KurtosisRemoteBackendConfig,
 ) (backend_interface.KurtosisBackend, error) {
-	remoteDockerClientOpts, err := buildRemoteDockerClientOpts(remoteBackendConfig)
+	remoteDockerClientOpts, cleanCertFilesFunc, err := buildRemoteDockerClientOpts(remoteBackendConfig)
+	if cleanCertFilesFunc != nil {
+		defer cleanCertFilesFunc()
+	}
```
gbouv: Here you can just do `defer cleanCertFilesFunc()` after the `if err != nil`.
If `err != nil`, we know the function won't be nil.

gbouv (approved):
Oups, good catch, I changed this last week to fix a small bug, and I didn't realize it would break with the cert files. Thanks for fixing!

PR #1038, opened 2023-08-01
feat!: subnetwork capabilities removed from Kurtosis. Users will have to update their Kurtosis package if they are using subnetwork capabilities

leoporoli:
## Description:
Removing the subnetwork feature also known as network partitioning.
This is a breaking change that demands users to update their Kurtosis package if they are using subnetwork capabilities and all the SDK calls that enable or configure this feature, like the CreateEnclave SDK method.

YES


leeederek (approved):
Looks good from the documentation perspective 👍🏻

victorcolombo (approved):
LGTM 🚀

PR #1040, opened 2023-08-02
fix: Stop local running engine when switching context

gbouv:
NO


adschwartz (dismissed):

adschwartz (commented):
waiting for @laurentluce review instead

laurentluce (approved):

PR #1043, opened 2023-08-02
feat: enable streaming exec output in container engine [stream exec pt. 1]

tedim52:
## Description:
Enables streaming output of exec commands out of docker and k8s as they are being run on services. 

NOT THIS PR (but stream exec pt. 2 will be)


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -278,7 +278,6 @@ func (backend *DockerKurtosisBackend) GetUserServiceLogs(
 	return user_service_functions.GetUserServiceLogs(ctx, enclaveUuid, filters, shouldFollowLogs, backend.dockerManager)
 }
 
-// TODO Switch these to streaming so that huge command outputs don't blow up the API container memory
 // NOTE: This function will block while the exec is ongoing; if we need more perf we can make it async
 func (backend *DockerKurtosisBackend) RunUserServiceExecCommands(
```
gbouv, line 282: I assume this method will go away in your second PR, right?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/user_services_functions/run_user_service_exec_command_streamed.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	userServiceDockerContainer := userServiceDockerResource.ServiceContainer
+
+	execOutputLinesChan, finalExecChan := dockerManager.RunExecCommandWithStreamedOutput(
+		ctx,
+		userServiceDockerContainer.GetId(),
+		cmd)
+	return execOutputLinesChan, finalExecChan, nil
+}
```
gbouv: newline

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +957,106 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+		// Therefore, we ONLY call Attach, without Start
+		attachResp, err := dockerClient.ContainerExecAttach(context, execId, execStartConfig)
+		if err != nil {
+			sendErrorAndFail(execOutputChan, err, "An error occurred starting/attaching to the exec command")
+			return
+		}
+		defer attachResp.Close()
+
+		// Stream output from docker through output channel
+		reader := bufio.NewReader(attachResp.Reader)
```
gbouv, line 1016: Heh, you see, docker sends you a reader back, exactly what we discussed. I think readers in Go are much more idiomatic than returning a channel. But that's fine for now, returning a channel is what we're doing in a bunch of other places

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +957,106 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+		}
+
+		execId := createResp.ID
+		if execId == "" {
+			sendErrorAndFail(execOutputChan, stacktrace.NewError("Got back an empty exec ID when running '%v' on container '%v'", command, containerId), "An error occurred creating the exec process.")
+			return
+		}
+
+		execStartConfig := types.ExecStartCheck{
+			// Because detach is false, we'll block until the command comes back
```
gbouv: I don't understand this comment. Where will we be blocked? Aren't we reading the result stream asynchronously here?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +957,106 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+			Cmd:          command,
+		}
+
+		createResp, err := dockerClient.ContainerExecCreate(context, containerId, execConfig)
+		if err != nil {
+			sendErrorAndFail(execOutputChan, err, "An error occurred creating the exec process.")
+			return
+		}
+
+		execId := createResp.ID
```
gbouv: nit: You can extract all this out of the async function and return a "real" error, rather than one through the stream. Though both works

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +957,106 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+				if err == io.EOF {
+					break
+				} else {
+					sendErrorAndFail(execOutputChan, err, "An error occurred while executing exec command")
+					return
+				}
+			}
+			// Trim end of line
+			execOutputChan <- execOutputLine
+			time.Sleep(500 * time.Millisecond)
```
gbouv: Do you _need_ to sleep here? This seems off, the reader itself should block until it reads the next `\n`

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +957,106 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+			finalOutputString = finalOutputString + execOutputLine
+			if err != nil {
+				if err == io.EOF {
+					break
+				} else {
+					sendErrorAndFail(execOutputChan, err, "An error occurred while executing exec command")
+					return
+				}
+			}
+			// Trim end of line
```
gbouv: Where are you trimming it?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +957,106 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+			sendErrorAndFail(execOutputChan, err, "An error occurred inspecting the exec to get the response code")
+			return
+		}
+		if inspectResponse.Running {
+			sendErrorAndFail(execOutputChan, stacktrace.NewError("Expected exec to have stopped, but it's still running!"), "An error occurred while running the exec command")
+			return
+		}
+		unsizedExitCode := inspectResponse.ExitCode
+		if unsizedExitCode > math.MaxInt32 || unsizedExitCode < math.MinInt32 {
+			sendErrorAndFail(execOutputChan, stacktrace.NewError("Could not cast unsized int '%v' to int32 because it does not fit", unsizedExitCode), "An error occurred while inspecting the exec command")
```
gbouv: nit: This seems weird to have to send 2 messages. Maybe you can make the last one option or something
EDIT: reading the code ot sendErrorAndFail below, just pass in the error, and call Propagate above

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +957,106 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+		}
+		int32ExitCode := int32(unsizedExitCode)
+
+		// Send final result over channel
+		finalExecResultChan <- exec_result.NewExecResult(int32ExitCode, finalOutputString)
+	}()
+	return execOutputChan, finalExecResultChan
+}
+
+func sendErrorAndFail(destChan chan<- string, err error, msg string, msgArgs ...interface{}) {
```
gbouv: Where and it fail, concretely? To me this just sends a error through the channel but that's it

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/run_user_service_exec_command_streamed.go
```diff
@@ -0,0 +1,74 @@
...
+func RunUserServiceExecCommandWithStreamedOutput(
+	ctx context.Context,
+	enclaveId enclave.EnclaveUUID,
+	serviceUuid service.ServiceUUID,
+	cmd []string,
+	cliModeArgs *shared_helpers.CliModeArgs,
+	apiContainerModeArgs *shared_helpers.ApiContainerModeArgs,
+	engineServerModeArgs *shared_helpers.EngineServerModeArgs,
+	kubernetesManager *kubernetes_manager.KubernetesManager,
+) (chan string, chan *exec_result.ExecResult, error) {
```
gbouv, line 23: nit: you can orient the channels here to be more explicit. I think `<-chan` would work here

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
```diff
@@ -1482,6 +1484,105 @@ func (manager *KubernetesManager) RunExecCommand(
...
+			Stdin:     false,
+			Stdout:    true,
+			Stderr:    true,
+			TTY:       true,
+			Container: containerName,
+			Command:   command,
+		}
+
+		//Create a RESTful command request.
+		request := manager.kubernetesClientSet.CoreV1().RESTClient().
```
gbouv: Not super familiar with Kubernetes API, I trust you that's the best way to run the exec command. Though it seems quite painful

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +957,106 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+		}
+		unsizedExitCode := inspectResponse.ExitCode
+		if unsizedExitCode > math.MaxInt32 || unsizedExitCode < math.MinInt32 {
+			sendErrorAndFail(execOutputChan, stacktrace.NewError("Could not cast unsized int '%v' to int32 because it does not fit", unsizedExitCode), "An error occurred while inspecting the exec command")
+			return
+		}
+		int32ExitCode := int32(unsizedExitCode)
+
+		// Send final result over channel
+		finalExecResultChan <- exec_result.NewExecResult(int32ExitCode, finalOutputString)
```
gbouv: Interesting, why do you need to send the entire result here too?

container-engine-lib/lib/stream_writer/stream_writer.go
```diff
@@ -0,0 +1,35 @@
...
+func (writer *StreamWriter) Write(p []byte) (n int, err error) {
+	writer.mutex.Lock()
+	defer writer.mutex.Unlock()
+
+	writer.outputChannel <- string(p)
+
+	// TODO: remove this time.Sleep
+	// This sleep is a hack to make the streamed output UI deliver steadily as its
+	// should be removed once UX of streaming is improved
+	time.Sleep(500 * time.Millisecond)
```
gbouv: Can't we remove those sleeps?

container-engine-lib/lib/stream_writer/stream_writer.go
```diff
@@ -0,0 +1,35 @@
...
+
+func NewStreamWriter(underlying io.Writer, outputChannel chan string) *StreamWriter {
+	return &StreamWriter{
+		underlying:    underlying,
+		mutex:         &sync.Mutex{},
+		outputChannel: outputChannel,
+	}
+}
+
+func (writer *StreamWriter) Write(p []byte) (n int, err error) {
```
gbouv: O.o so, if I understand correctly, what `StreamWriter ` is that it implements the Writer interface and it writes both to the underlying writer and the channel, is that correct?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
```diff
@@ -1482,6 +1484,105 @@ func (manager *KubernetesManager) RunExecCommand(
...
+			if err != nil {
+				sendErrorAndFail(
+					execOutputChan,
+					stacktrace.Propagate(err, "There was an error trying to parse the message '%s' to an exit code.", statusError),
+					"An error occurred while streaming from kubernetes with following exit code '%d'",
+					1)
+				return
+			}
+
+			finalExecResultChan <- exec_result.NewExecResult(exitCode, outputBuffer.String())
```
gbouv: Still wondering why you need to send the entire output once again here. If you can avoid it, then you basically don't need your StreamWriter

container-engine-lib/lib/stream_writer/stream_writer.go
```diff
@@ -0,0 +1,35 @@
...
+	writer.mutex.Lock()
+	defer writer.mutex.Unlock()
```
gbouv: Wow, this one frightens me as hell. I'm guessing if you added a lock, you experienced a weird behaviour without one

gbouv (commented):
Globally it looks good! Though the StreamWriter frightens me, wondering if we really need it

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
tedim52: Correct!

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: I think this is a much better idea! Will do.

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: removed this code actually, will remove comment as well.

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: You're absolutely right, this sleep can be removed. It was only initially added for testing so stream would print more steadily -> this might be desirable for UX, but if so can be done in apic.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
tedim52: very painful lol. I followed how it was done for the initial `kubernetesManager.RunExecCommand`, but all examples I found online and api point to doing it this way :/

container-engine-lib/lib/stream_writer/stream_writer.go
tedim52: That's correct.

container-engine-lib/lib/stream_writer/stream_writer.go
tedim52: yes, we can!

container-engine-lib/lib/stream_writer/stream_writer.go
tedim52: So I was actually following the ex. of[ `ConcurrentWriter` ](https://github.com/kurtosis-tech/kurtosis/blob/44d3b16528a8523f3c20842e373b3b51458fb267/container-engine-lib/lib/concurrent_writer/concurrent_writer.go#L8)here - `bytes.Buffer` is not [thread safe](https://stackoverflow.com/questions/19646717/is-the-go-bytes-buffer-thread-safe) - which is why the lock exists. This only applies if `underlying` is a `bytes.Buffer` tho so maybe I should change the struct to take in a `bytes.Buffer` as opposed to `io.Writer`. wdyt?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: You're right. Wonder if I should modify `ExecResult` to only contain an error code then. If you agree, I can adjust `ExecResult` when deprecating the existing exec logic, for now I'll just return empty string.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/run_user_service_exec_command_streamed.go
tedim52: I opted not to do this because I found I might need to send errors through the channel upstream in apic.

container-engine-lib/lib/stream_writer/stream_writer.go
tedim52: Is there anyway I can make its behavior more explicit - maybe a better name than `StreamWriter`?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
tedim52: don't need send output to but `StreamWriter` is needed to send output through channel

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: adjusted to `sendErrorThroughChannel`

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: That comment is a remnant of the old `dockerManager.RunExecCommand` code I lifted, but it doesn't seem relevant anymore - the reader returned by `ContainerExecAttach` is connected directly to the exec process so I don't think it matters whether the command runs in detached mode or not so I'll remove the comment.

gbouv: Cool, yeah let's remove the comment. Would be good also to understand this `detach` flag as well, just to make sure we're not missing anything. It seems here we should be running in detached mode, dunno 🤷‍♂️

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
gbouv: Yeah I agree, let's not add UX consideration into the backend implementation. If we do need to sleep between each print, this should be added wherever we print this stuff (in this case it's in the CLI). But let's not overthink this too much, we can see how it operates in practice and if we feel a strong need to add the sleep, we can add it later

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
gbouv: 👍  Perfect

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/run_user_service_exec_command_streamed.go
gbouv: Really?? In any case the channels that are getting returned by this function will be "read only" channels right? Or are you planning to have other components pushing values to those channels as well? For which purpose?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
gbouv: I think it would be a lot cleaner to only contain the exit code yes. The alternative here would be to have a single channel of composite objects like
```
string output_line
optional int32 exit_code
```
While the execution is in progress you keep exit code nil and sends the output lines, and when the excution finished you send the last line with the exit code (or the exit code alone) and the consumer knows that the execution finished.
Both works, the alternative solution saves you one channel but adds an "object type". No strong opinion here, apart from not sending the entire output once again at the end

container-engine-lib/lib/stream_writer/stream_writer.go
gbouv: 👍 perfect!

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
```diff
@@ -1482,6 +1484,90 @@ func (manager *KubernetesManager) RunExecCommand(
...
+	finalExecResultChan := make(chan *exec_result.ExecResult)
+	go func() {
+		defer func() {
+			close(execOutputChan)
+			close(finalExecResultChan)
+		}()
+
+		// Stream output from k8s to output channel
+		outputBuffer := &bytes.Buffer{}
+		streamWriter := stream_writer.NewStreamWriter(outputBuffer, execOutputChan)
```
gbouv: Coolios so here you don't need the outputBuffer anymore, right? You can implement a simple ChannelWriter, something like:
```
type ChannelWriter struct {
	channel chan string
}

func (writer *ChannelWriter) Write(input []byte) (int, error) {
	writer.channel <- string(input)
	return len(input), nil
}
```
which is a bit simpler than the StreamWriter (no lock)

gbouv (commented):
I think now you can simplify the StreamWriter since you don't need the bytes buffer, but the rest looks good 👍 
One thing I was thinking while doing the second review is that we don't have to use a "Read entire line" mechanism, we could just use a regular/fixed-size bytes buffer and send the buffer through the channel directly. Would that work?
The reason I'm raising this is b/c if the output line you're reading is more than 4Mb, you'll hit the GRPC limit again, which is quite annoying. Using a fixed-size buffer would be quite clean and also alleviate this GRPC limitation. Wdyt?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: deal, and ill test setting detach to false to verify

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/run_user_service_exec_command_streamed.go
tedim52: if an error in processing exec output upstream (eg. exit code or setting runtime values) then currently I send the error through `execOutputChan`, might not great tho so we can continue the discussion in the apic PR and change there if needed.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
tedim52: yes this is much better, will change

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +956,100 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+		return nil, nil, stacktrace.Propagate(err, "An error occurred creating the exec process")
+	}
+
+	execId := createResp.ID
+	if execId == "" {
+		return nil, nil, stacktrace.NewError("Got back an empty exec ID when running '%v' on container '%v'", command, containerId)
+	}
+
+	execStartConfig := types.ExecStartCheck{
+		// Because detach is false, we'll block until the command comes back
```
gbouv: Still not sure what we get from this command... Either make it more explicit or remove it, but as it is it's more confusing that anything

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +956,100 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+		}
+		unsizedExitCode := inspectResponse.ExitCode
+		if unsizedExitCode > math.MaxInt32 || unsizedExitCode < math.MinInt32 {
+			sendErrorThroughChannel(execOutputChan, stacktrace.NewError("Could not cast unsized int '%v' to int32 because it does not fit", unsizedExitCode))
+			return
+		}
+		int32ExitCode := int32(unsizedExitCode)
+
+		// Don't send output in final result because it was already streamed
+		finalExecResultChan <- exec_result.NewExecResult(int32ExitCode, "")
```
gbouv, line 1047: Are we still using ExecResult for something else? Or should we just remove the string attribute from it entirely?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -954,6 +956,100 @@ func (manager *DockerManager) RunExecCommand(context context.Context, containerI
...
+	}()
+	return execOutputChan, finalExecResultChan, nil
+}
+
+func sendErrorThroughChannel(destChan chan<- string, err error) {
+	destChan <- err.Error()
```
gbouv: nit: wondering if it's really worth hive this one-line function. Can't we inline it above?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
```diff
@@ -1482,6 +1484,89 @@ func (manager *KubernetesManager) RunExecCommand(
...
+
+			// Don't send output in final result because it was already streamed
+			finalExecResultChan <- exec_result.NewExecResult(exitCode, "")
+		}
+	}()
+	return execOutputChan, finalExecResultChan, nil
+}
+
+func sendErrorThroughChannel(destChan chan<- string, err error) {
+	destChan <- err.Error()
```
gbouv: nit: same remark here, maybe just inline this above

container-engine-lib/lib/channel_writer/channel_writer.go
```diff
@@ -0,0 +1,17 @@
+package channel_writer
+
+// Enables extracting output that is being written to an io.Writer via a channel
+type ChannelWriter struct {
+	channel chan string
+}
+
+func NewChannelWriter(channel chan string) *ChannelWriter {
```
gbouv, line 8: I'm surprised there isn't anything for this in go's std lib... But I couldn't find anything

gbouv (approved):
Coolios, so I guess this works but are we still safeguarding against lines that would be bigger than 4MB? I just want to make sure we're not re-introducing the bug Gyani fixed a few weeks ago.
I think there's a clean way to fix that bug here by streaming fixed-size byte chunks, instead of non-fixed-size output lines, but I'm fine keeping that for a future improvement

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: verified detach must be false, not exactly sure why, but i made comment more explicit

container-engine-lib/lib/channel_writer/channel_writer.go
tedim52: rightttt, its so simple and would seem useful?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: done

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
tedim52: done

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: It's used in the old `RunExecCommand` which I'm waiting to remove until the second PR for stream exec is up. But I'll note to remove it in the second PR!

PR #1052, opened 2023-08-03
docs: Add documentation for persistent service directories

gbouv:
## Description:
Add documentation on what a persistent directory is, and also how to migrate from old files artifact syntax in `ServiceConfig` objects to the new one using `Directory` objects

NO


docs/docs/starlark-reference/directory.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Directory
+sidebar_label: Directory
+---
+
+The `Directory` constructor creates a `Directory` object that represents a directory inside an existing service (see
+the [ServiceConfig][service-config] object).
+Directory object can be either a files artifact or a persistent directory, depending on the arguments passed to the
```
mieubrisse, line 8: Can't it actually be both - populated by a files artifact, AND a persistent directory?

docs/docs/starlark-reference/directory.md
```diff
@@ -0,0 +1,36 @@
...
+file_artifact_directory = Directory(
+    artifact_name=files_artifact_1,
+)
+# Or:
+persistent_directory = Directory(
+    persistent_key="data-directory"
+)
+```
+
+A directory composed of a files artifact will be automatically provisioned with the given files artifact content and
```
mieubrisse: I thought it wasn't either-or (which is why we have them both as possible flags)?

docs/docs/starlark-reference/directory.md
```diff
@@ -0,0 +1,36 @@
...
+    artifact_name=files_artifact_1,
+)
+# Or:
+persistent_directory = Directory(
+    persistent_key="data-directory"
+)
+```
+
+A directory composed of a files artifact will be automatically provisioned with the given files artifact content and 
+will be immutable. The service on which the directory is mounted will not be able to modify the
```
mieubrisse: Isn't this also untrue? Once the files artifact's contents are copied in, the service can modify freely

docs/docs/starlark-reference/directory.md
```diff
@@ -0,0 +1,36 @@
...
+content of the files artifact. In the above example, `files_artifact_1` is a files artifact name. (see 
+[upload_files][upload-files-reference], [render_templates][render-templates-reference] and 
+[store_service_files][store-service-reference] to learn more about on how to create file artifacts).
+
+A persistent directory, as its name indicates, persists over service updates and restarts. It is uniquely identified 
+by its `persistent_key` and the service ID on which it is being used (a persistent directory cannot be shared across
+multiple services). When it is first created, it will be empty. The service can write anything in it. When the service 
+gets updated, the data in it persists. It is particularly useful for services data directory, logs directory, etc.
```
mieubrisse: `for services` -> `for a service's`

mieubrisse (commented):
Questions around implementation

docs/docs/starlark-reference/directory.md
gbouv: Nope populating both will throw an interpreation error right now. 
What would you expect to happen in this case? For the files artifact to be mounted to the persistent directory? B/c files artifact are immutable right now, this would basically be equivalent to mounting a regular files artifact.

docs/docs/starlark-reference/directory.md
gbouv: It is a XOR right now. I think you need to clarify what the workflow would be when a user specify both a files artifact name and a persistent directory. It's a bit unclear to me

docs/docs/starlark-reference/directory.md
gbouv: Really? I didn't know that. Isn't this an anti-pattern though? Why would a service modify a files artifact? And also, the modification would not be propagated to the artifact stored in the APIC, so it's pretty confusing. I thought we blocked this, and if not shouldn't we block it?

mieubrisse: I'm not actually sure how we'd block it - on k8s we could probably do it by mounting the volume as ReadOnly, but on Docker I don't think we can.

mieubrisse (approved):
Approved pending removing the "immutable" aspect

docs/docs/starlark-reference/directory.md
gbouv: I'm pretty sure we can mount a volume in read only mode in Docker.
I just tried
```
docker volume create test
docker run --name test-volume --rm -v test:/data:ro docker/getting-started
```
(notice the `:ro` at the end of the volume)

and that volume is read only from the container perspective:
```
/ # cd /data/
/data # echo 'Bonjour' > test.txt 
/bin/sh: can't create test.txt: Read-only file system
/data # 
```

gbouv: I'll resolve this comment to be able to merge, but let me know if you'd like to keep track of this, think more about it, of just action it. IDT it would be too complex to put in place, but that's maybe not something we want to do, deferring to you on this. ping me directly if you want to talk more about it

PR #1053, opened 2023-08-03
fix: Fix comment about sidecar

gbouv:
## Description:
Just changing the wording of an error message to reflect what Kurtosis is actually doing.

NO

### DIFF
--- a/core/server/api_container/server/service_network/default_service_network.go
+++ b/core/server/api_container/server/service_network/default_service_network.go
@@ -1446,7 +1446,7 @@ func (network *DefaultServiceNetwork) destroyService(ctx context.Context, servic
 	}
 	successfullyDestroyedUuids, failedToDestroyUuids, err := network.kurtosisBackend.DestroyUserServices(context.Background(), network.enclaveUuid, userServiceFilters)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1446,15 +1446,15 @@ func (network *DefaultServiceNetwork) destroyService(ctx context.Context, servic
 	}
 	successfullyDestroyedUuids, failedToDestroyUuids, err := network.kurtosisBackend.DestroyUserServices(context.Background(), network.enclaveUuid, userServiceFilters)
 	if err != nil {
-		return stacktrace.Propagate(err, "Attempted to destroy the services with UUID '%v' but had no success. You must manually destroy the services. Kurtosis will now try to remove its sidecar if it exists but might it fail as well.", serviceUuid)
+		return stacktrace.Propagate(err, "Attempted to destroy the service with UUID '%v' but had no success. You must manually destroy the service as well as its sidecar.", serviceUuid)
```
h4ck3rk3y: are we sure that the service has a sidecar? (here and there)

if not perhaps

s/as its sidecar/as its sidecar if any/g

h4ck3rk3y (approved):

PR #1055, opened 2023-08-03
fix: cluster set doesnt get into a weird state of no cluster being set

h4ck3rk3y:
## Description:
Don't restart the engine in cluster set


cli/cli/commands/cluster/set/set.go
```diff
@@ -60,51 +57,10 @@ func run(ctx context.Context, flags *flags.ParsedFlags, args *args.ParsedArgs) e
 		return nil
 	}
 
-	engineManagerOldCluster, err := engine_manager.NewEngineManager(ctx)
```
h4ck3rk3y, line 63: note this creates a failure condition in case the local engine isn't stopped; where there would be portclash starting gateway

leoporoli (approved):

h4ck3rk3y:
Just ran into this! @laurentluce there was a slack discussion around this I believe; do you think we still need this?

laurentluce:
Here is what Kevin had to say about that: https://kurtosistech.slack.com/archives/C04HLLL98J2/p1690984606826499?thread_ts=1690933988.153719&cid=C04HLLL98J2.

We made an exception to that list for when we switch from a local context to a remote context and we automatically stop the engine running locally so it does not conflict with the forwarded remote engine port.

cli/cli/commands/cluster/set/set.go
```diff
@@ -60,51 +57,10 @@ func run(ctx context.Context, flags *flags.ParsedFlags, args *args.ParsedArgs) e
...
-				command_str_consts.ClusterSetCmdStr, clusterPriorToUpdate)
-		}
-	}()
-	logrus.Infof("Cluster set to '%s', Kurtosis engine will now be started", clusterName)
-	engineManagerNewCluster, err := engine_manager.NewEngineManager(ctx)
-	if err != nil {
-		return stacktrace.Propagate(err, "An error occurred creating an engine manager.")
-	}
-	// Start the engine inside the new cluster
-	_, engineClientCloseFunc, err := engineManagerNewCluster.StartEngineIdempotentlyWithDefaultVersion(ctx, defaults.DefaultEngineLogLevel, defaults.DefaultEngineEnclavePoolSize)
```
h4ck3rk3y, line 94: add back this so that we have an engine in case none is running

cli/cli/commands/cluster/set/set.go
```diff
@@ -60,51 +62,62 @@ func run(ctx context.Context, flags *flags.ParsedFlags, args *args.ParsedArgs) e
 		return nil
 	}
 
-	engineManagerOldCluster, err := engine_manager.NewEngineManager(ctx)
-	if err != nil {
-		return stacktrace.Propagate(err, "An error occurred creating an engine manager.")
+	// at the moment docker clusters are local only, we stop engine in them in favor of switching to any other cluster
```
laurentluce, line 75: Should we fail if the current context is a remote one so we don't end up stopping the engine on the remote docker instance?  The user would have to first switch to the local context using `kurtosis context switch default` before setting the cluster to kubernetes.

h4ck3rk3y: in my mind i had isolated the context & the cluster users, I will do a check for current context type and pop an error with the recommended step if remote

PR #1066, opened 2023-08-04
feat: Update files if already present in enclave

gbouv:
## Description:
Right now, when a Starlark package creates a file artifact (via `upload_files`, `render_template`, or `store_service_file`), if a file artifact already exist with that name inside the enclave, it fails. This is a big pain for idempotency.
This PR introduces files artifact updates. From now on, when a files artifact is "created" via starlark and already exists in the enclave (based on its artifact name), Kurtosis will check the equality of the files based on their hash, and if they are different the files artifact will be replaced by the new one (and of course it no-ops if they are the same).
No change requires in the Starlark packages

YES


api/golang/core/lib/shared_utils/shared_utils.go
```diff
@@ -21,17 +21,33 @@ const (
...
+	if err != nil {
+		return nil, 0, nil, stacktrace.Propagate(err,
+			"Failed to open the archive file at '%s' during files upload for '%s'.", compressedFilePath, pathToCompress)
+	}
+	return compressedFile, compressedFileSize, compressedFileContentMd5, nil
+}
+
+// CompressPathToFile compressed the entire content of the file or directory at pathToCompress and returns the path
+// to the TGZ archive created, alongside the size (in bytes) of the archive
+// The consumer should take care of closing the io.ReadClose returned
```
leoporoli: isn't returning a read/closer.

leoporoli: it seems it's missing a comment about the third return value

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
```diff
@@ -1232,6 +1232,13 @@ func (manager *KubernetesManager) CreatePod(
 		logrus.Debugf("Going to start pod using the following JSON: %v", string(podDefinitionBytes))
 	}
 
+	// In case of a service update, it's possible the pod has not been fully deleted yet. It's still "scheduled for
+	// deletion", and so we wait for it to be deleted before creating it again. There's probably a way to optimize this
+	// a bit more using native k8s pod update operation
+	if err := manager.waitForPodDeletion(ctx, namespaceName, podName); err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred waiting for pod '%v' to become available", podName)
```
leoporoli: it seems the error msg is not accurate.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
```diff
@@ -1877,6 +1884,36 @@ func (manager *KubernetesManager) waitForPodAvailability(ctx context.Context, na
 	)
 }
 
+// waitForPodDeletion waits for the pod to be fully deleted if it has been marked for deletion
+func (manager *KubernetesManager) waitForPodDeletion(ctx context.Context, namespaceName string, podName string) error {
+	// Wait for the pod to start running
+	deadline := time.Now().Add(podWaitForAvailabilityTimeout)
```
leoporoli: if is necessary to reuse the constant I think that it could be renamed to support both cases

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1551,20 +1580,35 @@ func (network *DefaultServiceNetwork) copyFilesFromServiceUnlocked(ctx context.C
 		return "", stacktrace.Propagate(err, "An error occurred getting the files artifact store")
 	}
 
+	existingFilesArtifactUuid, _, _, filesArtifactAlreadyExists, err := store.GetFile(artifactName)
+	if err != nil {
+		return "", stacktrace.Propagate(err, "An unexpected error occurred checking for file artifact '%s' existence in the store", artifactName)
+	}
+
 	pipeReader, pipeWriter := io.Pipe()
 	defer pipeReader.Close()
```
leoporoli: out of the scope, but just to check, I'm seeing that this reader is closed in the main routine and in the subroutine, is this ok?

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -1725,8 +1769,22 @@ func (network *DefaultServiceNetwork) renderTemplatesUnlocked(templatesAndDataBy
...
+	existingFilesArtifactUuid, _, existingFilesArtifactMd5, found, err := store.GetFile(artifactName)
 	if err != nil {
+		return "", stacktrace.Propagate(err, "An unexpected error occurred checking for file artifact '%s' existence in the store", artifactName)
+	}
+
+	var storeOrUpdateErr error
+	var filesArtifactUuid enclave_data_directory.FilesArtifactUUID
+	if found {
+		filesArtifactUuid = existingFilesArtifactUuid
+		if len(existingFilesArtifactMd5) == 0 || len(compressedFileMd5) == 0 || !bytes.Equal(existingFilesArtifactMd5, compressedFileMd5) {
```
leoporoli, line 1160: could you add a comment above these conditions or extract them in a helper function with a descriptive name?

leoporoli: I can image that you are comparing files content, but I think that a comment could help or a helper method

core/server/api_container/server/startosis_engine/kurtosis_instruction/upload_files/upload_files.go
```diff
@@ -104,36 +109,87 @@ func (builtin *UploadFilesCapabilities) Interpret(locatorOfModuleInWhichThisBuil
...
-		return "", stacktrace.Propagate(err, "An error occurred while uploading the compressed contents\n'%v'", compressedData)
+		return "", stacktrace.Propagate(err, "An error occurred checking if the files artifact '%s' is already stored in this enclave", builtin.artifactName)
+	}
+	var instructionResult string
+	if found && (len(builtin.filesArtifactMd5) > 0 && bytes.Equal(currentlyStoredFileContentMd5, builtin.filesArtifactMd5)) {
+		instructionResult = fmt.Sprintf("Files with artifact name '%s' resolved with artifact UUID '%s' as content were matching", builtin.artifactName, currentlyStoredFileArtifactUuid)
+	} else if found {
+		err = builtin.serviceNetwork.UpdateFilesArtifact(currentlyStoredFileArtifactUuid, filesArtifactContentReader, builtin.filesArtifactMd5)
+		if err != nil {
+			return "", stacktrace.Propagate(err, "An error occurred while uploading the compressed contents with md5 '%s' to artifact '%s' (UUID: '%s')",
```
leoporoli: the error msg is not accurate

leoporoli: it should be ...while updating...

leoporoli (approved):
LGTM

core/server/api_container/server/service_network/default_service_network.go
gbouv: Good call, I think we want to close it only in the subroutine, the other one is probably a mistake, I'll remove it.
I think it doesn't make a real difference here b/c the code will be blocked on `gzipAndPushTarredFileBytesToOutput` until the subroutine is finished. But I guess it also doesn't hurt to close the channel only once :D

core/server/api_container/server/service_network/default_service_network.go
gbouv: Will do!

core/server/api_container/server/startosis_engine/kurtosis_instruction/upload_files/upload_files.go
gbouv: ok, this is the error that will be visible so I thought having the same error message for both update and upload would be appropriate to hide the inner mechanism of this instruction. But I'm also fine updating the message, will do!

PR #1069, opened 2023-08-04
refactor: service identifiers repository implementation (inside APIC)

leoporoli:
## Description:
service identifiers repository implementation (inside APIC) for persistent historical service identifiers 

NO

This is part of the [APIC restart project](https://www.notion.so/kurtosistech/API-Container-Restart-567fe61fb72c422ba7483afb12a3cddf)


core/server/api_container/server/service_network/service_identifiers/service_identifier.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (serviceIdentifier *serviceIdentifier) GetShortenedUUIDStr() string {
+	return serviceIdentifier.shortenedUuidStr
+}
+
+func (serviceIdentifier *serviceIdentifier) GetName() service.ServiceName {
+	return serviceIdentifier.name
+}
+
+func (serviceIdentifier *serviceIdentifier) MarshalJSON() ([]byte, error) {
```
victorcolombo, line 41: Given that `serviceIdentifier` is a private struct, can we drop this in favor of the default implementation of `json.Marshal`?

victorcolombo:
@leoporoli  Could we add an E2E test that creates an enclave, adds a service, restarts APIC then get services and see if its there?

core/server/api_container/server/service_network/service_identifiers/service_identifier.go
leoporoli: if we do that we are going to break the encapsulation.
serviceNane

I could implement the `privateInternalStruct` pattern that you show me today instead, but in this case, we will need a custom Marshal and Unmarshal method also. WDYT?

victorcolombo: Fine to leave as is then!

victorcolombo (approved):

PR #1070, opened 2023-08-04
feat: Enclave inspect relying on the API container service only

laurentluce:
## Description:
The current `enclave inspect` command gets the services info from the backend and from the APIC.  This is done to be able to display the enclave details even when the enclave (APIC) is stopped.  In practice, inspecting a stopped enclave hasn't proven useful so we are migrating this command to only use the APIC.  This change is also required by the bastion/remote host design because the CLI only has access to the remote backend APIC.

The APIC provides all the information required except the service status so we are adding it to `ServiceInfo`.

YES


api/protobuf/core/api_container_service.proto
```diff
@@ -68,6 +68,11 @@ message Port {
   string maybe_wait_timeout = 4;
 }
 
+enum ServiceStatus {
+  STOPPED = 0;
+  RUNNING = 1;
```
gbouv, line 73: Do we need `REMOVED` as well here? My memory is a bit fuzzy here, but since we're not removing the container but just stopping it, do we keep the service in memory at the APIC level and should we return it as `REMOVED` here? Probably not, but just double checking

core/server/api_container/server/api_container_service.go
```diff
@@ -774,3 +778,12 @@ func getTextRepresentation(reader io.Reader, lineCount int) (*string, error) {
 	text := textRepresentation.String()
 	return &text, nil
 }
+
+func convertServiceStatusToServiceInfoStatus(serviceStatus service.ServiceStatus) kurtosis_core_rpc_api_bindings.ServiceStatus {
+	switch serviceStatus {
+	case service.ServiceStatus_Started:
+		return kurtosis_core_rpc_api_bindings.ServiceStatus_RUNNING
+	default:
```
gbouv, line 791: This is an errorprone pattern. If we add more values to the enum in the future we will silently default to STOPPED.
We should rather `case` on the two values and throw an error for anything else, this way we will fail loudly. Ideally we would want to fail at compile time, but go doesn't make that super easy. Less ideally but still better than throwing an error would be to have a good linter for this, but not sure if that exists. Short term throwing an error is good enough

gbouv (approved):

api/protobuf/core/api_container_service.proto
laurentluce: We don't keep removed services at that level.

PR #1086, opened 2023-08-08
feat: Portal automatic start and stop on context change

### LINKED ISSUE #970: The portal daemon should start locally automatically when a remote context is used
### Background & motivation

The portal daemon on the user machine needs to be started manually when a remote context is used.  It results in CLI error messages before the user remembers to start the daemon.

### Desired behaviour

The portal daemon should start locally automatically when a remote context is used

laurentluce:
## Description:
The user currently needs to make sure that the portal is running before switching to a remote context.  This change does that automatically for the user.  This change also stops the portal when the context switches back to local.

Add the portal release tag to install so we can pin a portal version instead of using the latest.  The portal release tag can also be set to "latest".  This PR pins the portal version to 0.0.4

Upcoming changes:
- Release portal 0.0.5 containing the support for remote endpoints.
- Update the mono-repo to pass the remote endpoint type to the portal client forward port calls.  Set the portal version to latest.

YES

Closes #970


cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -120,11 +123,11 @@ func SwitchContext(
 			}
 		}
 	} else {
-		if store.IsRemote(currentContext) {
-			return stacktrace.NewError("New context is remote but Kurtosis Portal is not reachable locally. " +
-				"Make sure Kurtosis Portal is running before switching to a remote context again.")
+		if err := portalManager.StopExisting(ctx); err != nil {
```
gbouv, line 131: Probably fine for now, I'm wondering if we shouldn't live it running at some point. It's pretty lightweight anyway. maybe add a comment mentioning this possibility for future readers of this code

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -54,48 +56,60 @@ func DownloadLatestKurtosisPortalBinary(ctx context.Context) (bool, error) {
 	}
 
 	ghClient := github.NewClient(http.DefaultClient)
-	latestRelease, err := getLatestRelease(ctx, ghClient)
+	var release *github.RepositoryRelease
+	if portalTag == portalTagLatest {
```
gbouv: Can't we just call `getReleaseByTag` and if the tag is `latest` we just pull latest? Seems overcomplicated to have 2 different functions here. Maybe there's some difficulty I'm missing

cli/cli/helpers/portal_manager/binary_artifact_getter.go
```diff
@@ -54,48 +56,60 @@ func DownloadLatestKurtosisPortalBinary(ctx context.Context) (bool, error) {
...
 	if err != nil {
-		return nil, stacktrace.Propagate(err, "Unable to retrieve latest version of Kurtosis Portal form GitHub")
+		return nil, stacktrace.Propagate(err, "Unable to retrieve latest version of Kurtosis Portal from GitHub")
 	}
 	logrus.Debugf("Latest release is %s", latestRelease.GetName())
 	return latestRelease, nil
 }
 
+func getReleaseByTag(ctx context.Context, ghClient *github.Client, tag string) (*github.RepositoryRelease, error) {
+	release, _, err := ghClient.Repositories.GetReleaseByTag(ctx, kurtosisTechGithubOrg, kurtosisPortalGithubRepoName, tag)
```
gbouv: Yes I'd basically move your `if == "latest"` in here either call `GetLatestRelease` or `GetReleaseByTag`. This way you have only one function, and you move a bit of the complexity inside this simple function, rather than adding some to the already "fairly complex" upstream one.

cli/cli/helpers/portal_manager/portal_manager.go
```diff
@@ -213,6 +213,35 @@ func (portalManager *PortalManager) MapPorts(ctx context.Context, localPortToRem
...
 
+// DownloadAndStart downloads the required version and starts the portal if the required version
+// is not already running.
+func (portalManager *PortalManager) DownloadAndStart(ctx context.Context) error {
+	// Checking if new version is available and potentially downloading it
+	currentPortalPid, process, isPortalReachable, err := portalManager.CurrentStatus(ctx)
+	if err != nil {
+		return stacktrace.Propagate(err, "Unable to determine current state of Kurtosis Portal process")
+	}
+	if isPortalReachable {
```
gbouv: I might be missing something here, but I'm under the impression that if the portal is already running, we exit early and we don't even check whether it's running the required version. Whereas in the comment on top of this function you say "if the required version is not already running". 
It seems that we should validate the currently running process is the required version, and not an older one, right?

gbouv (commented):
I think there's a mismatch between what you say in your comments and what the code is actually doing. Worth for you to take a second look. Maybe I'm missing something also. Let me know if you want to chat

cli/cli/commands/portal/start/start.go
```diff
@@ -22,32 +23,22 @@ var PortalStartCmd = &lowlevel.LowlevelKurtosisCommand{
...
-	}
-
-	currentPortalPid, process, isPortalReachable, err := portalManager.CurrentStatus(ctx)
+	currentPortalPid, _, isPortalReachable, err := portalManager.CurrentStatus(ctx)
 	if err != nil {
 		return stacktrace.Propagate(err, "Unable to determine current state of Kurtosis Portal process")
 	}
+
 	if isPortalReachable {
 		logrus.Infof("Portal is currently running on PID '%d' and healthy.", currentPortalPid)
```
gbouv, line 36: Maybe add a comment here explaining why we're not trying to upgrade it (we don't want to break the currently running port forward, as the portal is stateless)

gbouv (approved):

Issue #934, opened 2023-07-18
Use index instead or enclave name of hash for enclave and engine names


barnabasbusa:
### Background & motivation

Would it be possible to migrate from `kurtosis-enclave--af4a70624dae4a91a91cab4cfdc37f75` to `kurtosis-enclave--0`
and possibly `kurtosis-enclave--1,2,3` in case of multiple enclaves?

Same for engine; going from `kurtosis-engine--856cb53dd06c425c988612a366fec76b` to `kurtosis-engine--0`

It can be somewhat confusing when you have multiple enclaves running at the same time, and have to switch among them. 

### Desired behaviour

Mentioned above. 

### How important is this to you?

Nice to have; this feature would make using Kurtosis more enjoyable.

mieubrisse:
Hey @barnabasbusa , could explain more how you're coming across the enclave UUID and how you're doing the switch between them? We try to make the name the central feature by which enclaves are manipulated, e.g.:
```
$ kurtosis enclave ls
UUID           Name               Status     Creation Time
79813f61cdf7   gloomy-steppe      RUNNING    Mon, 17 Jul 2023 13:27:02 -03
0222af561e1f   calm-orchard       RUNNING    Mon, 17 Jul 2023 14:25:55 -03
93f3016a4d09   dull-marsh         RUNNING    Mon, 17 Jul 2023 14:26:38 -03
11e998fee997   lazy-sea           RUNNING    Mon, 17 Jul 2023 14:26:49 -03
3e4466c0210f   stunning-atoll     RUNNING    Mon, 17 Jul 2023 14:33:53 -03
c5d35f9492de   aged-reef          RUNNING    Mon, 17 Jul 2023 14:38:06 -03
6aa3c8c99351   empty-glade        RUNNING    Mon, 17 Jul 2023 16:23:35 -03
6fcb12f513dc   sacred-plateau     RUNNING    Tue, 18 Jul 2023 10:19:42 -03
```

barnabasbusa:
My primary interaction with the content of an enclave is kubectl, and namespaces, not kurtotis itself (yet).

mieubrisse:
Ah, got it. I'd be keen to learn more about what workflows are missing in the `kurtosis` CLI for that to be your primary interaction with Kurtosis? We try to make the CLI pleasant to use with lots of quality-of-life stuff (e.g. tab-complete for all fields, descriptive error messages, etc.)

barnabasbusa:
[...]

PR #1100, opened 2023-08-11
feat: make enclave namespace and network naming deterministic

tedim52:
## Description:
- Limit enclave names to 60 chars
- Make Kubernetes namespace and Docker network names be `kt-ENCLAVENAME`

YES


api/golang/engine/lib/enclave/consts.go
```diff
@@ -1,3 +1,3 @@
 package enclave
 
-const AllowedEnclaveNameCharsRegexStr = `^[-A-Za-z0-9._]{1,63}$`
+const AllowedEnclaveNameCharsRegexStr = `^[-A-Za-z0-9._]{1,59}$`
```
mieubrisse: isn't this 59 chars (not 60)? either is fine with me, but we should be clear

mieubrisse: maybe should also add a comment explaining _why_ we're making this particular length restriction

mieubrisse (approved):

mieubrisse:
Fixes #934

api/golang/engine/lib/enclave/consts.go
tedim52: lol facepalm tedi

PR #1123, opened 2023-08-18
feat: Add connect go to APICs

Peeeekay:
YES/NO


core/server/api_container/server/api_container_connect_service.go
```diff
@@ -0,0 +1,95 @@
+package server
```
Peeeekay: will add implementation in next; combined with ui integration so we should have everything very soon!

api/protobuf/engine/engine_service.proto
```diff
@@ -102,6 +104,8 @@ message EnclaveAPIContainerHostMachineInfo {
   string ip_on_host_machine = 4;
   // The grpc port on the container engine host machine where the API container can be reached
   uint32 grpc_port_on_host_machine = 5;
+  // the connect grpc port on the container engine host machine where API container can be reached temporarily
+  uint32 connect_port_on_host_machine = 6;
```
adschwartz: @Peeeekay could you help me understand what this is? Is it the port you use to connect to the APIC when you are on the bridge network?
is `connect` in this relation refer to `connect-go`? I'm thinking we remove the word connect.
`temporarily`: in what way is this temporary? Is the port going to change so you have to call back periodically?

api/typescript/src/engine/kurtosis_engine_rpc_api_bindings/connect/engine_service_pb.d.ts
```diff
@@ -258,6 +265,13 @@ export declare class EnclaveAPIContainerHostMachineInfo extends Message<EnclaveA
    */
   grpcPortOnHostMachine: number;
 
+  /**
+   * the connect grpc port on the container engine host machine where API container can be reached temporarily
```
adschwartz: Same question here

api/protobuf/engine/engine_service.proto
Peeeekay: Yes it is the port they will allow
me to use to connect to connect-go; I think I missed propagating connect internal port let me do that. It would be internal port where connect is running that needs to accessed by EM. The reason I said temp was cause we will eventually have one port as gRPC will be removed. @adschwartz

PR #1139, opened 2023-08-21
feat: added streaming to enclave manager

Peeeekay:
YES/NO


enclave-manager/server/server.go
```diff
@@ -118,15 +135,39 @@ func (c *WebServer) ListFilesArtifactNamesAndUuids(ctx context.Context, req *con
...
 	serviceRequest := &connect.Request[kurtosis_core_rpc_api_bindings.RunStarlarkPackageArgs]{
 		Msg: req.Msg.RunStarlarkPackageArgs,
 	}
+
 	apicStream, err := (*apiContainerServiceClient).RunStarlarkPackage(ctx, serviceRequest)
-	res := apicStream.Msg()
-	return str.Send(res)
+	ctxWithCancel, cancel := context.WithCancel(ctx)
+
+	logs := getRuntimeLogsWhenCreatingEnclave(cancel, apicStream)
```
Peeeekay, line 197: Meta point:
1. will propagate errors in the next pr
2. there may be cases that this may bug out; will tackle them whenever they appear. 

Streaming currently works for both service logs and create enclave; they are different in nature --
1. Service logs - they are streaming continuously until client sends a stop signal
2. RunStarlarkPackage - stop the stream whenever io.EOF is encountered 

that's why there's slight different in the logic.

engine/frontend/src/component/CreateEnclaveView.js
```diff
@@ -26,49 +26,68 @@ export const CreateEnclaveView = ({packageId, enclave, args}) => {
...
-                }
-            }
-          });
-
-          stream.on("end", () => {
-            setLoading(false)
-          });
+              if (result.case === "instructionResult" && result.value.serializedInstructionResult) {
+                  if (result.value.serializedInstructionResult.includes(SERVICE_IS_ADDED)) {
+                      //getServices(enclave.apiClient)
```
Peeeekay, line 47: will add this back in. next pr.

enclave-manager/server/server.go
```diff
@@ -89,12 +89,29 @@ func (c *WebServer) GetServiceLogs(
...
-	res := result.Msg()
-	return str.Send(res)
+
+	logs := getServiceLogsFromEngine(result)
+	for {
+		select {
+		case <-ctx.Done():
+			err := result.Close()
+			if err != nil {
+				logrus.Errorf("Error ocurred: %+v", err)
```
adschwartz, line 149: @Peeeekay minor thing but maybe we `stack.propagate` this error rather than logging here. That way the caller of this function can figure out what to do with it

enclave-manager/server/server.go
adschwartz: Great work PK.

enclave-manager/server/server.go
adschwartz: Saw your comment further down. Would you mind making the change in this PR?

adschwartz (approved):

adschwartz:
I merged this in manually to better handle the merge conflicts. Will close this out.

PR #1141, opened 2023-08-22
refactor: Update cloud API to return more AWS info

gbouv:
## Description:
It's expected that the cloud BE will also return the bucket region, name and user folder created for this user.

NO


Peeeekay (approved):
Looks good

PR #1143, opened 2023-08-23
fix: lower calls to backend by doing get all services more efficiently

### LINKED ISSUE #1074: Retrieving enclave services fails with a large number of services
### What's your CLI version?

0.81.4

Commands or instructions retrieving a large number of enclave services fail.  This can also be seen when the Gateway starts.  The service network which sits in front of the backend only supports retrieving a single service which results in a large number of backend calls.  The K8S client rate limiter gets triggered.

### Desired behavior

Retrieving a large number of enclave services from the APIC should work with the help of a single call to the backend.

h4ck3rk3y:
## Description:
We were imposed with Kubernetes rate limits due to how we were doing GetServcies by fetching all individual services for plural runs as well, this fetches once from the underlying backend

YES - user will get speed improvement but doesn't need docs

Closes #1074


core/server/api_container/server/api_container_service.go
```diff
@@ -810,6 +827,8 @@ func convertServiceStatusToServiceInfoStatus(serviceStatus service.ServiceStatus
 		return kurtosis_core_rpc_api_bindings.ServiceStatus_RUNNING, nil
 	case service.ServiceStatus_Stopped:
 		return kurtosis_core_rpc_api_bindings.ServiceStatus_STOPPED, nil
+	case service.ServiceStatus_Registered:
```
h4ck3rk3y, line 830: @laurentluce i had to add support for this as perhaps making the call an individual call is just too fast for the APIC

laurentluce (approved):
Thanks @h4ck3rk3y !

PR #1147, opened 2023-08-23
feat: added package to loader UI

Peeeekay:
User can browse through all the available packages and run them.


engine/frontend/src/App.js
```diff
@@ -1,13 +1,15 @@
 import './App.css';
 import Home from './component/Home';
+import { ChakraProvider } from '@chakra-ui/react'
 
 const App = () => {
-    console.log("Enclave Manager version: 2023-08-28-2")
```
adschwartz, line 5: @Peeeekay leave this in here so we can track the versions to make sure we know what version people are running in the wild.

engine/frontend/src/component/PackageCatalogOption.js
```diff
@@ -0,0 +1,47 @@
...
+const PackageCatalogOption = () => {
+    const navigate = useNavigate();
+    const location = useLocation();
+    const handleCatalogForm = () => {
+        if (!(location.pathname.includes("catalog"))) {
+            navigate("/catalog")
+        }
+    }
+
+    const handleManualCatalogBtn = () => {
```
adschwartz: Should this do anything?

adschwartz (approved):

PR #1156, opened 2023-08-24
fix: get service logs tail

Peeeekay:
YES: This PR fixes the issue where `kurtosis service logs -f` should be able to tail the logs from services.


engine/server/engine/centralized_logs/client_implementations/kurtosis_backend/kurtosis_backend_logs_database_client_test.go
```diff
@@ -169,39 +169,39 @@ func TestStreamUserServiceLogs_FilteredAllLogLines(t *testing.T) {
 
 }
 
-func TestStreamUserServiceLogs_FilteredAllLogLinesWithFollowLogs(t *testing.T) {
-	expectedServiceAmountLogLinesByServiceUuid := map[service.ServiceUUID]int{
```
Peeeekay, line 173: The follow logs do not work at the moment because they will get timeout after 2 seconds; need to add a kill signal.. which i can we can in next pr.. @tedim52 .. this should make kurtosis service logs -f usable until then.

engine/server/engine/centralized_logs/client_implementations/persistent_volume/persistent_volume_logs_database_client_test.go
```diff
@@ -91,131 +91,6 @@ func TestStreamUserServiceLogs_WithFilters(t *testing.T) {
 	require.NoError(t, testEvaluationErr)
 }
 
-func TestStreamUserServiceLogs_WithFiltersAndWithFollowLogs(t *testing.T) {
```
Peeeekay, line 94: the test won;t work as they will timeout after 2 seconds. would need to a add a kill singal. to make kurtosis service logs -f usable again; lets get this merged in and refactor the tests to support `tail` case. @tedim52

engine/server/engine/centralized_logs/client_implementations/persistent_volume/persistent_volume_logs_database_client.go
```diff
@@ -162,11 +162,15 @@ func streamServiceLogLines(
 			logrus.Debugf("Context was canceled, stopping streaming service logs for service '%v' in enclave '%v", serviceUuid, enclaveUuid)
 			return
 		default:
-			jsonLogStr, err := logsReader.ReadString(newlineRune)
+			jsonLogStr, _, err := logsReader.ReadLine()
 			if err != nil && errors.Is(err, io.EOF) {
 				// exiting stream
-				logrus.Debugf("EOF error returned when reading logs for service '%v' in enclave '%v'", serviceUuid, enclaveUuid)
-				return
+				if shouldFollowLogs {
```
Peeeekay, line 166: keep looping if `tail` is passed; otherwise stop when EOF is found.

leoporoli (approved):

PR #1158, opened 2023-08-24
feat: use proxy url

adschwartz:
## Description:
need to use the proxy url when hitting the remote instance

NO


engine/frontend/src/component/Home.js
```diff
@@ -34,17 +34,11 @@ const queryParamToBool = (value) => {
...
-    }
-    return undefined;
-}
-
 const createApiUrl = (apiHost, requireHttps) => {
-    const protocol = requireHttps ? "https" : "http"
-    const address = !apiHost ? DefaultApiHost : parseApiHostIntoIp(apiHost);
-    return `${protocol}://${address}:${DefaultApiPort}`;
+    if(requireHttps){
+        return `https://cloud.kurtosis.com/gateway/ips/${apiHost}/ports/${DefaultApiPort}`;
```
Peeeekay, line 39: this could be a global variable.

`https://cloud.kurtosis.com/gateway/ips/`

Peeeekay (approved):

engine/frontend/src/component/Home.js
adschwartz: yeah I'll make this change

PR #1162, opened 2023-08-25
fix: paths

adschwartz:
NO


engine/frontend/src/component/EnclaveInfo.js
```diff
@@ -87,18 +87,18 @@ const EncalveInfo = ({enclaves}) => {
     }, [name, enclaves])
 
     const handleServiceClick = (service, services) => {
-        navigate(`/enclaves/${name}/services/${service.uuid}`, {state: {services, selected: service}})
+        navigate(`services/${service.uuid}`, {state: {services, selected: service}})
     }
 
     const handleLeftPanelClick = (enclaveName) => {
-        navigate(`/enclaves/${enclaveName}`, {replace:true})
+        navigate(`./../${enclaveName}`, {replace:true})
```
Peeeekay: Not a big fan of relative paths in navigate; is it possible to use constructpath method to make this absolute ?

Peeeekay (approved):

PR #1171, opened 2023-08-28
feat: Production mode enclave

victorcolombo:
## Description:

This PR introduces the concept of "production mode" enclave. For now this only means that services will restart upon failure. Additionally, both non-production and production enclaves now have their APIC restart policy as "on failure".

For create a "production enclave", you can use the new SDK call `createProductionEnclave`, or via CLI using the new flag `kurtosis enclave add --production` or `kurtosis run --production`.

YES


cli/cli/commands/enclave/add/add.go
```diff
@@ -68,6 +68,13 @@ var EnclaveAddCmd = &engine_consuming_kurtosis_command.EngineConsumingKurtosisCo
 			),
 			Type: flags.FlagType_String,
 		},
+		{
+			Key:       "production",
+			Usage:     "Deletes all enclaves, regardless of whether they're already stopped",
```
leoporoli: is this text ok?

container-engine-lib/lib/backend_impls/docker/docker_manager/create_and_start_container_args.go
```diff
@@ -218,6 +222,12 @@ func (builder *CreateAndStartContainerArgsBuilder) WithLoggingDriver(loggingDriv
 	return builder
 }
 
+// Will configure the container to use and specific logging driver which can be configured using the different implementations
```
leoporoli: the comment is not accurated

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/user_services_functions/start_user_services.go
```diff
@@ -676,7 +676,7 @@ func createStartServiceOperation(
 			volumeMounts,
 		).WithLoggingDriver(
 			fluentdLoggingDriverCnfg,
-		)
+		).WithRestartPolicy(docker_manager.RestartPolicy(""))
```
leoporoli: I think you can use the "NoRestart" constant that you created here

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/label_key_consts/label_key_consts.go
```diff
@@ -66,3 +67,4 @@ var PrivateIPDockerLabelKey = docker_label_key.MustCreateNewDockerLabelKey(priva
 var UserServiceGUIDDockerLabelKey = docker_label_key.MustCreateNewDockerLabelKey(userServiceGuidDockerLabelKeyStr)
 var LogsEnclaveUUIDDockerLabelKey = docker_label_key.MustCreateNewDockerLabelKey(logsEnclaveUuidLabelKeyStr)
 var LogsServiceUUIDDockerLabelKey = docker_label_key.MustCreateNewDockerLabelKey(logsServiceUuidDockerLabelKey)
+var EnclaveModeLabelKey = docker_label_key.MustCreateNewDockerLabelKey(enclaveTypeLabelKeyStr)
```
leoporoli: nit: is a bit confusing for me that the var name says `EnclaveMode` but the key says `enclave-type`, I think that these could be aligned WDYT?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -133,6 +133,13 @@ const (
 	dontStreamStats        = false
 )
 
+type RestartPolicy string
```
leoporoli, line 136: nice!

engine/server/engine/enclave_manager/enclave_creator.go
```diff
@@ -176,6 +180,7 @@ func (creator *EnclaveCreator) launchApiContainer(
 		grpcListenPort,
 		creator.apiContainerKurtosisBackendConfigSupplier,
 		enclaveEnvVars,
+		true,
```
leoporoli: nit: could use a constant here

engine/server/engine/enclave_manager/enclave_manager.go
```diff
@@ -139,7 +140,7 @@ func (manager *EnclaveManager) CreateEnclave(
 		return nil, stacktrace.Propagate(err, "An error occurred validating enclave name '%v'", enclaveName)
 	}
 
-	if manager.enclavePool != nil {
+	if !isProduction && manager.enclavePool != nil {
```
leoporoli, line 144: I think you could pass the `isProduction` value to the `enclavePool.GetEnclave` method and the value could be compared in the `areRequestedEnclaveParamsEqualToEnclaveInThePoolParams` helper to determine if it can be pulled from the pool or not. I mean in order to follow the same pattern that we did for the other enclave arguments, like `apiContainerLogLevel` for instance.

leoporoli (approved):
LGTM, the only missing part that I found is the docs upgrade explaining `CreateProductionEnclave` and the upgrade on `CreateEnclave` (which now creates an enclave in test mode) are you planning to add these in another PR?

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/label_key_consts/label_key_consts.go
victorcolombo: Good catch, this is unused

engine/server/engine/enclave_manager/enclave_manager.go
victorcolombo: I added a todo. But in general I think production enclaves are less sensitive to startup time since they will be long running

PR #1190, opened 2023-08-29
feat: add creation dialog

adschwartz:
## Description:
When user submits the package to run we disable the button and provide a spinner to indicate it's being worked on.

NO


engine/frontend/src/component/PackageCatalogForm.js
```diff
@@ -108,25 +114,26 @@ const PackageCatalogForm = ({handleCreateNewEnclave}) => {
     const location = useLocation()
     const {state} = location;
     const {kurtosisPackage} = state
+    const [runningPackage, setRunningPackage] = useState(false)
```
Peeeekay, line 117: I think you need similar logic on `CreateEnclaveModel` -- it's for manual flow.

adschwartz: oh good catch

PR #1191, opened 2023-08-29
fix: Point to the engine restart command as part of the context switch failure remediation to not conflict with lower level commands

laurentluce:
## Description:
The engine remediation command in the context switch logic sometimes conflicts with lower level engine remediation commands.  One example is when the running engine version does not match the CLI version, the user is presented with two remediation commands: `engine start` and `engine restart`.  Only the second one remediates the issue.  This change updates the engine remediation command in the context switch logic to be `engine restart` which also works if the engine is not running.

Note that we don't want to restart the engine automatically when there is a version mismatch because different users could be using different CLI versions and we don't want to end up in a tug of war match.

YES


cli/cli/commands/kurtosis_context/context_switch/switch.go
```diff
@@ -71,7 +71,7 @@ func SwitchContext(
 				"needs to be stopped before the context can be switched. The engine status can be obtained running"+
 				"kurtosis %s %s and it can be stopped manually by running kurtosis %s %s.",
 				command_str_consts.EngineCmdStr, command_str_consts.EngineStatusCmdStr,
-				command_str_consts.EngineCmdStr, command_str_consts.EngineStartCmdStr)
+				command_str_consts.EngineCmdStr, command_str_consts.EngineStopCmdStr)
```
laurentluce, line 74: Typo

cli/cli/helpers/engine_manager/engine_existence_guarantor.go
```diff
@@ -187,13 +187,10 @@ func (guarantor *engineExistenceGuarantor) VisitStopped() error {
 // unusual and very bad, so we'd rather fail loudly
 func (guarantor *engineExistenceGuarantor) VisitContainerRunningButServerNotResponding() error {
 	remediationCmd := fmt.Sprintf(
-		"%v %v %v && %v %v %v",
```
laurentluce, line 190: Replace `kurtosis engine stop && kurtosis engine start` with `kurtosis engine restart` since it does the same.

adschwartz (approved):
This is a great catch @laurentluce. It helps to tell the user to use `engine restart`, but it's still odd to me that `engine restart != engine stop && engine start`. I feel like we should fix that bit at some point.

laurentluce:
> This is a great catch @laurentluce. It helps to tell the user to use `engine restart`, but it's still odd to me that `engine restart != engine stop && engine start`. I feel like we should fix that bit at some point.

This change is unrelated to the main change.  I did this one as a cleanup task.

PR #1200, opened 2023-08-30
fix:  made forwarding efficient by reducing calls to Kubernetes

### LINKED ISSUE #1075: Service ports forward connection initialization is slow with a large number of services
### What's your CLI version?

0.81.4

The Gateway takes a long time to initialize the service port forward connections when there are many services.  Two backend calls are made to build each connection.

### Desired behavior

The Gateway should initialize the service port forward connections with no calls to the backend. The port forward URL has a well-known format and it can be built directly since we know how the namespace and pod are derived from the enclave and service info.

h4ck3rk3y:
This PR needs some work. This has been created for a visual diff


cli/cli/kurtosis_gateway/connection/provider.go
```diff
@@ -211,7 +182,14 @@ func (provider *GatewayConnectionProvider) getRunningPodNamesByLabels(namespace
 	return runningPodNames, nil
 }
 
+// TODO - this function shouldn't exist when kt- + enclave name is the namespace inside kurtosis
+// https://github.com/kurtosis-tech/kurtosis/issues/1203 - till then we cache it
 func (provider *GatewayConnectionProvider) getEnclaveNamespaceNameForEnclaveId(enclaveId string) (string, error) {
+	enclaveNamespaceName, found := provider.enclaveIdToEnclaveNamespaceName[enclaveId]
```
laurentluce, line 188: Great workaround until the namespace names are deterministic.

laurentluce (approved):
Thanks @h4ck3rk3y !

cli/cli/kurtosis_gateway/connection/provider.go
h4ck3rk3y: Thank you! :-)

PR #1207, opened 2023-08-30
fix: Remove 'kurtosis version' depending on the engine

### LINKED ISSUE #1206: `kurtosis version` hangs for 7 seconds, consistently
### What's your CLI version?

0.82.14

When I run `kurtosis version`, my CLI hangs for 7 seconds before returning. This is very surprising and unexpected to me, and I suspect it's related to some sort of Portal bug.

### Desired behavior

`kurtosis version` should return immediately

mieubrisse:
## Description:
I was surprised that `kurtosis version` was taking 7 seconds without any extra output, and so I checked out the code to discover that the  `kurtosis version` command was depending on the engine. This felt wrong to me, given that we already have `engine status` to display the version (and it's more detailed besides - e.g. if the engine is unreachable, you get a message saying why).  If we want one command to display the status of both CLI and engine(which would also be fine), we should consolidate the two so that we're not duplicating partial functionality of `engine status` in `version`.

All that said, I'd also be fine if we went the other way and built out `version` heavily to properly communicate what's happening with the engine, and why it's hanging for 7s. This seemed like it would be a bigger change though and my Goland is already yelling at me for Go workspaces stuff that I don't know how to fix, so I went with the quickest change I could.

YES

Fixes #1206


h4ck3rk3y (approved):

leeederek:
Added docs here: https://github.com/kurtosis-tech/kurtosis/pull/1240, @mieubrisse can I get a review 🙏🏻

PR #1227, opened 2023-09-04
docs: Mention that cloud S3 data is now public

gbouv:
NO


docs/docs/guides/running-in-kurtosis-cloud.md
```diff
@@ -21,6 +21,11 @@ When provisioning a cloud instance, Kurtosis will create a specific AWS user acc
 
 Services running inside a Kurtosis enclave in Kurtosis Cloud can freely read and write objects from/to this S3 storage.
 
+:::warning All data in S3 is publicly exposed
+Only the user can write to its own S3 space. However, the data is publicly available to whoever knows the key of
+the stored object. No sensitive data should be stored.
+:::
```
adschwartz, line 27: @gbouv Proposal:
:::warning S3 storage is publicly accessible
While only the owner of the S3 storage is permitted to _write_ to the S3 storage, the data is _publicly accessible_ to anyone that knows the object key. For this reason, we don't recommend storing sensitive data in the S3 storage.
:::

adschwartz: @gbouv on line 20 (four lines up) we mention that:
> Kurtosis will create a specific AWS user account and a private storage space in S3.
Maybe we remove the "private" part or clarify what we mean with that.

adschwartz: @gbouv I have lost track of the technical complexity on making this private to the user, but is this something we can/should do soon?

gbouv: Thanks Anders I'll update with your proposal, which is a lot better than mine (and I'll remove the "private")

On your last question, well unfortunately this is required for Hyperlane to be able to run their package b/c they use unauthenticated AWS client to read from s3...
So, I even though I'm not a big fan of having all the data public, we don't have a good workaround. I think Kevin is okay with it until we provide a first class way for the client to provision whatever they want in their own AWS account. but this requires creating a distinct account per cloud user inside AWS, and then exposing the capability of running terraform or whatever to setup their AWS account with whatever they need (S3 bucket, KMS, etc.). So this requires a bit of work to get there

adschwartz: Got it! Thanks!

adschwartz (approved):

PR #1230, opened 2023-09-04
fix: disable Starlark time module

h4ck3rk3y:
## Description:
Disabled Starlark time module

YES


h4ck3rk3y:
@mieubrisse wants just now() to be disabled and it seems like another ticket is tracking it; closing this

PR #1232, opened 2023-09-04
fix: Do not delete runtime value during idempotent runs

### LINKED ISSUE #1225: missing future references in idempotent run
### What's your CLI version?

0.82.18

```
# run this to completion
$ kurtosis run github.com/kurtosis-tech/eth-network-package@gyani/deterministic-time --enclave spooky-meadow
# run it again
$ kurtosis run github.com/kurtosis-tech/eth-network-package@gyani/deterministic-time --enclave spooky-meadow
```

image

### Desired behavior

All instructions should be SKIPPED future reference shouldn't be lost

victorcolombo:
NO

Closes https://github.com/kurtosis-tech/kurtosis/issues/1225


victorcolombo:
@gbouv This is what I believe to be the fix. I imagine we delete the value and keep the same UUID in case a service restarts and we get new information about the service. That being said, in the case where nothing changes but a print of the same runtime value of the existing service, what is the expected behavior?

gbouv (approved):
Yup I think this is correct, it was a mistake to delete the value. It's working when the instruction is an update, but it will fail if the instruction is effectively skipped at runtime.
Ideally we should differentiate between those 2 cases, and remove the value if the instruction is an update, but I guess it's fine to keep old unused values for now. Good catch!

PR #1234, opened 2023-09-04
feat: implements enclave plan repository for persisting Startosis idempotent run

leoporoli:
## Description:
implements enclave plan repository for persisting Startosis idempotent run)

NO

This is part of the APIC restart project


core/server/api_container/server/startosis_engine/kurtosis_instruction/add_service/add_services.go
```diff
@@ -81,6 +82,19 @@ type AddServicesCapabilities struct {
...
+func (builtin *AddServicesCapabilities) GetEnclavePlanCapabilities() *enclave_plan_capabilities.EnclavePlanCapabilities {
+	enclavePlanCapabilitiesBuilder := enclave_plan_capabilities.NewEnclavePlanCapabilitiesBuilder(AddServicesBuiltinName)
+
+	serviceNames := []service.ServiceName{}
+
+	for serviceName := range builtin.serviceConfigs {
+		serviceNames = append(serviceNames, serviceName)
+	}
+
+	enclavePlanCapabilitiesBuilder.WithServiceNames(serviceNames)
```
gbouv, line 94: nit: wondering if we could call `WithServiceName(serviceName)` multiple time directly inside the loop above to shorten a bit this function
If we do that, maybe `AddServiceName` would be better (rather than `WithServiceName`)

core/server/api_container/server/startosis_engine/kurtosis_instruction/add_service/add_service.go
```diff
@@ -86,6 +87,12 @@ type AddServiceCapabilities struct {
 	resultUuid string
 }
 
+func (builtin *AddServiceCapabilities) GetEnclavePlanCapabilities() *enclave_plan_capabilities.EnclavePlanCapabilities {
+	enclavePlanCapabilitiesBuilder := enclave_plan_capabilities.NewEnclavePlanCapabilitiesBuilder(AddServiceBuiltinName)
+	enclavePlanCapabilitiesBuilder.WithServiceName(builtin.serviceName)
+	return enclavePlanCapabilitiesBuilder.Build()
```
gbouv, line 93: nit: I like the builder pattern here, would be even nicer if we could so something like:
```
return enclave_plan_capabilities.NewEnclavePlanCapabilitiesBuilder(AddServiceBuiltinName)
    .WithServiceName(builtin.serviceName)
    .Build()
```
and same in all the other instructions

core/server/api_container/server/startosis_engine/kurtosis_instruction/enclave_plan_instruction.go
```diff
@@ -0,0 +1,10 @@
+package kurtosis_instruction
+
+import (
+	"github.com/kurtosis-tech/kurtosis/core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_capabilities"
+)
+
+type EnclavePlanInstruction interface {
+	GetKurtosisInstructionStr() string
```
gbouv, line 8: nit: I'd make this more explicit. I guess we're returning the starlark code here right? So maybe `GetStarlarkCode` (given that it's applied on an instruction object)

core/server/api_container/server/startosis_engine/kurtosis_instruction/kurtosis_instruction.go
```diff
@@ -34,5 +35,11 @@ type KurtosisInstruction interface {
 	ValidateAndUpdateEnvironment(environment *startosis_validator.ValidatorEnvironment) error
 
 	// TryResolveWith assesses whether the instruction can be resolved with the one passed as an argument.
-	TryResolveWith(other KurtosisInstruction, enclaveComponents *enclave_structure.EnclaveComponents) enclave_structure.InstructionResolutionStatus
+	TryResolveWith(other EnclavePlanInstruction, enclaveComponents *enclave_structure.EnclaveComponents) enclave_structure.InstructionResolutionStatus
+
+	GetCapabilites() EnclavePlanCapabilitiesProvider
```
gbouv, line 40: Can we be more explicit here? given that there's already `*Capablities` object for each instruction, this broad name is verrrry confusing

core/server/api_container/server/startosis_engine/kurtosis_instruction/kurtosis_instruction.go
```diff
@@ -34,5 +35,11 @@ type KurtosisInstruction interface {
...
+}
+
+type EnclavePlanCapabilitiesProvider interface {
```
gbouv, line 43: nit: one interface per file maybe?

gbouv: What's this purpose of this provider? Can't the above `GetCapabilities` return the capability straight, whithout this provider wrapper?

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_capabilities/enclave_plan_capabilites_builder.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/service"
+)
+
+type EnclavePlanCapabilitiesBuilder struct {
+	instructionTypeStr string
+	serviceName        service.ServiceName
+	serviceNames       []service.ServiceName
+	artifactName       string
+	filesArtifactMD5   []byte
```
gbouv, line 12: I think it would be less confusing to keep the mapping file artifact -> file MD5. Here we're losing the relationship info, which right now it surprisingly not a blocker, but might become on. And generally speaking I prefer not creating too much entropy

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_capabilities/enclave_plan_capabilites_builder.go
```diff
@@ -0,0 +1,59 @@
+package enclave_plan_capabilities
+
+import (
+	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/service"
+)
+
+type EnclavePlanCapabilitiesBuilder struct {
+	instructionTypeStr string
+	serviceName        service.ServiceName
+	serviceNames       []service.ServiceName
```
gbouv, line 10: I don't understand those 2 fields... Can't we have only one `serviceNames` array, and eventually have a single element in it?

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_capabilities/enclave_plan_capabilites_builder.go
```diff
@@ -0,0 +1,59 @@
+package enclave_plan_capabilities
+
+import (
+	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/service"
+)
+
+type EnclavePlanCapabilitiesBuilder struct {
```
gbouv, line 7: See my other comments, at first glance this seems like a big bag of everything. We should make it representative of what an instruction can affect IMO (i.e. one or multiple services, one or multiple files)

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_capabilities/enclave_plan_capabilities.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"encoding/json"
+	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/service"
+	"github.com/kurtosis-tech/stacktrace"
+)
+
+// Use NewEnclavePlanCapabilitiesBuilder to construct this object
+type EnclavePlanCapabilities struct {
+	// we do this way in order to have exported fields which can be marshalled
+	// and an unexported type for encapsulation
```
gbouv, line 12: This isn't clear to me at all to be honest. I trust you this is needed, but I need more info to really understand why we need such a complex structure here...
If this is just to be able to have public attributes, honestly I think we should just put the attribute in this object even though they are public. This nested private struct is a lot more difficult to understand IMO rather than bearing the cost of having public attributes

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_capabilities/enclave_plan_capabilities.go
```diff
@@ -0,0 +1,60 @@
+package enclave_plan_capabilities
+
+import (
+	"encoding/json"
+	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/service"
+	"github.com/kurtosis-tech/stacktrace"
+)
+
+// Use NewEnclavePlanCapabilitiesBuilder to construct this object
```
gbouv, line 9: outdated comment here?

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_capabilities/enclave_plan_capabilities_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+func TestEnclavePlanCapabilitiesMarshallers(t *testing.T) {
+	originalEnclavePlanCapabilities := getEnclavePlanCapabilitiesForTest()
+
+	marshaledEnclavePlanCapabilities, err := json.Marshal(originalEnclavePlanCapabilities)
+	require.NoError(t, err)
+	require.NotNil(t, marshaledEnclavePlanCapabilities)
+
+	// Suppressing exhaustruct requirement because we want an object with zero values
+	// nolint: exhaustruct
```
gbouv, line 22: why not doing `newEnclavePlanCapabilities  = new(EnclavePlanCapabilities)` to avoid the // nolint ?

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_capabilities/enclave_plan_capabilities_test.go
```diff
@@ -0,0 +1,68 @@
...
+	newEnclavePlanCapabilities := &EnclavePlanCapabilities{}
+
+	err = json.Unmarshal(marshaledEnclavePlanCapabilities, newEnclavePlanCapabilities)
+	require.NoError(t, err)
+
+	require.Equal(t, originalEnclavePlanCapabilities, newEnclavePlanCapabilities)
+	require.Equal(t, filesArtifactContentStr, string(newEnclavePlanCapabilities.GetFilesArtifactMD5()))
```
gbouv, line 29: I'd personally prefer to have more simple and explicit test where you just manually write the JSON.
Those 2-ways test are nice but they are not `unit` per-se. So basically if both the marshaller and the unmarshaller forget about a field, the test would still pass but you would not persist this field

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_instruction/enclave_plan_instruction_impl.go
```diff
@@ -0,0 +1,57 @@
+package enclave_plan_instruction
+
+import (
+	"encoding/json"
+	"github.com/kurtosis-tech/kurtosis/core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_capabilities"
+	"github.com/kurtosis-tech/stacktrace"
+)
+
+type EnclavePlanInstructionImpl struct {
```
gbouv, line 9: nit: we usually don't like the `*Impl` name for struct. Why do you use it here? Is it because the `EnclavePlanInstruction` interface has only one implementation? If yes, then why do we even have an interface?

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_instruction/enclave_plan_instruction_impl.go
```diff
@@ -0,0 +1,57 @@
...
+	// we do this way in order to have exported fields which can be marshalled
+	// and an unexported type for encapsulation
+	privateEnclavePlanInstruction *privateEnclavePlanInstruction
+}
+
+type privateEnclavePlanInstruction struct {
+	KurtosisInstructionStr string
+	Capabilities           *enclave_plan_capabilities.EnclavePlanCapabilities
```
gbouv, line 17: Any pros for splitting the instruction and the capability object here? We already have a ton of objects for pretty much anything, I don't feel too nervous about flattening the structure a bit here and having all the capabilities field at this level

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_instruction/enclave_plan_instruction_impl_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func TestEnclavePlanInstructionMarshallers(t *testing.T) {
+	originalEnclavePlanInstruction := getEnclavePlanInstructionForTest(1)[0]
+
+	marshaledEnclavePlanInstruciton, err := json.Marshal(originalEnclavePlanInstruction)
+	require.NoError(t, err)
+	require.NotNil(t, marshaledEnclavePlanInstruciton)
+
+	// Suppressing exhaustruct requirement because we want an object with zero values
+	// nolint: exhaustruct
```
gbouv, line 32: Same remarks here

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_instruction/enclave_plan_instruction_repository.go
```diff
@@ -0,0 +1,103 @@
...
+	enclavePlanInstructionBucketName = []byte("enclave-plan-instruction-repository")
+)
+
+type EnclavePlanInstructionRepository struct {
+	enclaveDb *enclave_db.EnclaveDB
+}
+
+func GetOrCreateNewEnclavePlanInstructionRepository(enclaveDb *enclave_db.EnclaveDB) (*EnclavePlanInstructionRepository, error) {
+	if err := enclaveDb.Update(func(tx *bolt.Tx) error {
+		bucket, err := tx.CreateBucketIfNotExists(enclavePlanInstructionBucketName)
```
gbouv, line 27: nit: you can move this createBucketIfNotExists in the `Save` function, this will simplify the interface of this function by not returning an error

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_instruction/enclave_plan_instruction_repository.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			return stacktrace.Propagate(err, "An error occurred while creating the enclave plan instruction database bucket")
+		}
+		logrus.Debugf("Enclave plan instruction bucket: '%+v'", bucket)
```
gbouv, line 31: wondering what printing a bucket object looks like in the logs... Is that informative?

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_instruction/enclave_plan_instruction_repository.go
```diff
@@ -0,0 +1,103 @@
...
+	return repository, nil
+}
+
+func (repository *EnclavePlanInstructionRepository) Save(
+	uuid instructions_plan.ScheduledInstructionUuid,
+	instruction *EnclavePlanInstructionImpl,
+) error {
+
+	if err := repository.enclaveDb.Update(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(enclavePlanInstructionBucketName)
```
gbouv, line 51: Yeah here and below in Get just replace this with `GetOrCreateBucket`

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/enclave_plan_instruction/enclave_plan_instruction_repository.go
```diff
@@ -0,0 +1,103 @@
...
+	uuid instructions_plan.ScheduledInstructionUuid,
+) (*EnclavePlanInstructionImpl, error) {
+	// Suppressing exhaustruct requirement because we want an object with zero values
+	// nolint: exhaustruct
+	instruction := &EnclavePlanInstructionImpl{}
+
+	if err := repository.enclaveDb.View(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket(enclavePlanInstructionBucketName)
+
+		bucket.Sequence()
```
gbouv: Interesting, what does this do?

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/kurtosis_plan_instruction/kurtosis_plan_instruction.go
```diff
@@ -59,7 +69,14 @@ func (builtin *KurtosisPlanInstructionWrapper) CreateBuiltin() func(thread *star
 		if builtin.instructionPlanMask.HasNext() {
 			_, scheduledInstructionPulledFromMaskMaybe = builtin.instructionPlanMask.Next()
 			if scheduledInstructionPulledFromMaskMaybe != nil {
-				instructionResolutionStatus = instructionWrapper.TryResolveWith(scheduledInstructionPulledFromMaskMaybe.GetInstruction(), builtin.enclaveComponents)
+
+				scheduledInstructionUuid := scheduledInstructionPulledFromMaskMaybe.GetUuid()
```
gbouv, line 73: Interesting, what guarantees here that the UUID will be the same? Were you able to test that idempotent runs still work? Reading this, I got worried that we might loose the UUID of the scheduled instructions, but maybe I'm wrong. Quickly checking runs are still idempotent with an APIC restart in the middle would clear that doubt

gbouv: After thoughts, I think it's fine actually. But still, curious to know whether you got it to work

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/kurtosis_plan_instruction/kurtosis_plan_instruction.go
```diff
@@ -92,13 +110,20 @@ func (builtin *KurtosisPlanInstructionWrapper) CreateBuiltin() func(thread *star
 				logrus.Debugf("Marking the plan as invalid as instruction '%s' differs from '%s'",
 					instructionWrapper.String(), scheduledInstructionPulledFromMaskMaybe.GetInstruction().String())
 			}
+			kurtosisInstructionStr := scheduledInstruction.GetInstruction().String()
+			enclaveCapabilities := scheduledInstruction.GetInstruction().GetCapabilites().GetEnclavePlanCapabilities()
```
gbouv, line 118: This chain `.GetInstruction().GetCapabilites().GetEnclavePlanCapabilities()` is blowing my mind 🤯

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/kurtosis_plan_instruction/kurtosis_plan_instruction.go
```diff
@@ -92,13 +110,20 @@ func (builtin *KurtosisPlanInstructionWrapper) CreateBuiltin() func(thread *star
...
+			enclavePlanInstruction := enclave_plan_instruction.NewEnclavePlanInstructionImpl(kurtosisInstructionStr, enclaveCapabilities)
+			if err := builtin.enclavePlanInstructionRepository.Save(scheduledInstruction.GetUuid(), enclavePlanInstruction); err != nil {
```
gbouv, line 120: So, it's interesting here that we now need to do both `AddInstruction` and `Save`. I'm wondering if there could be a better hierarchy here such that we calling `AddInstruction` could do both. I don't think we'd ever want to `AddInstruction` withou saving it, right?

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/kurtosis_plan_instruction/kurtosis_plan_instruction.go
```diff
@@ -92,13 +110,20 @@ func (builtin *KurtosisPlanInstructionWrapper) CreateBuiltin() func(thread *star
...
+			}
+
 			return returnedFutureValue, nil
 		case enclave_structure.InstructionIsNotResolvableAbort:
 			// if the instructions differs, then the mask is invalid
 			builtin.instructionPlanMask.MarkAsInvalid()
 			logrus.Debugf("Marking the plan as invalid as instruction '%s' had the following resolution status: '%s'",
 				instructionWrapper.String(), instructionResolutionStatus)
-			if err := builtin.instructionsPlan.AddInstruction(instructionWrapper, returnedFutureValue); err != nil {
+			if _, err := builtin.instructionsPlan.AddInstruction(instructionWrapper, returnedFutureValue); err != nil {
```
gbouv, line 130: Ah, well, mayyybe? Why aren't we saving the instruction in this case?

leoporoli:
it was finally implemented here: https://github.com/kurtosis-tech/kurtosis/pull/1280

PR #1235, opened 2023-09-05
feat: retain logs for x weeks

tedim52:
## Description:
This PR is the final PR in centralized logging v0 over Docker. Specifically,

It implements basic log retention by retaining logs for 5 weeks and removing logs older than that period. (To support pulling logs from enclaves created prior to this feature until all users are migrated, pull strategy is based on the enclaves creation time. Although, logs created prior to this feature will not be removed via log retention.)

Overall, this allows users to retrieve logs from services in long running enclaves up to 4 weeks prior.

YES


engine/server/engine/server/engine_connect_server_service.go
```diff
@@ -20,29 +25,37 @@ type EngineConnectServerService struct {
...
 
-	//The client for consuming container logs from the logs' database server
-	logsDatabaseClient centralized_logs.LogsDatabaseClient
+	// The clients for consuming container logs from the logs' database server
+
+	// per week pulls logs from enclaves created post log retention feature
+	perWeekLogsDatabaseClient centralized_logs.LogsDatabaseClient
+
+	// per file pulls logs from enclaves created pre log retention feature
+	// TODO: remove once users are fully migrated to log retention/new log schema
```
tedim52, line 40: If you think that there are no users running long running enclaves (enclaves prior to next release) as of this PR, I can remove this and this PR would be simplified quite a bit

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -27,6 +29,7 @@ const (
 	maxWaitForEngineAvailabilityRetries         = 10
 	timeBetweenWaitForEngineAvailabilityRetries = 1 * time.Second
 	logsStorageDirpath                          = "/var/log/kurtosis/"
+	everyWeekCron                               = "0 0 * * 0"
```
gbouv: Wait, is that once per week on Sunday? How will that work? let's say I manually stop my engine every Friday when I log off and restart it every Monday morning when I arrive at the office... When will the logs get removed in this case?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -255,3 +264,13 @@ func CreateEngine(
 	shouldKillEngineContainer = false
 	return result, nil
 }
+
+func scheduleLogRemoval() error {
+	logRemover := log_remover.LogRemover{}
+	c := cron.New()
+	_, err := c.AddJob(everyWeekCron, logRemover)
+	if err != nil {
+		return err
```
gbouv: stacktrace.Propagate

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -255,3 +264,13 @@ func CreateEngine(
 	shouldKillEngineContainer = false
 	return result, nil
 }
+
+func scheduleLogRemoval() error {
+	logRemover := log_remover.LogRemover{}
+	c := cron.New()
```
gbouv: You don't need to keep a reference to it to stop it or whatever?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/consts/consts.go
```diff
@@ -0,0 +1,16 @@
+package consts
```
gbouv: All those constants are very log oriented, maybe name this more explicitly? Do you need to have a dedicated file for those consts? Can't they live at the top of a regular code file like we do everywhere else?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
```diff
@@ -0,0 +1,42 @@
...
+type LogRemover struct {
+}
+
+// Run implements the Job cron interface. It removes logs a week older than the log retention period.
+func (remover LogRemover) Run() {
+	// get the current week
+	_, week := time.Now().UTC().ISOWeek()
+
+	// compute the next oldest week
+	diff := week - (consts.LogRetentionPeriodInWeeks + 1) - 1
```
gbouv: Why are we doing `-2` here?
metapoint: Are we somehow testing this thing? I wouldn't mind us doing a weird `-2` here if next to this file I have decent unit tests. Maybe they are somewhere else though, not yet at this end of this PR

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
```diff
@@ -0,0 +1,42 @@
...
+	var weekPastRetentionPeriod int
+	if diff >= 0 {
+		weekPastRetentionPeriod = diff
+	} else {
+		weekPastRetentionPeriod = maxWeekNum + diff
```
gbouv: This seems overly complex... Can't you just do:
```
oneWeekDuration := 7 * 24 * time.Hour
time.Now().Add(- LogRetentionPeriodInWeeks * oneWeekDuration).ISOWeek()
```
and let go std library handle the logic?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	// create directory path for that week
+	oldLogsDirPath := fmt.Sprintf("%s%s/", consts.LogsStorageDirpath, strconv.Itoa(weekPastRetentionPeriod))
+	if _, err := os.Stat(oldLogsDirPath); !os.IsNotExist(err) {
```
gbouv: Straight from `RemoveAll` doc:
```
If the path does not exist, RemoveAll returns nil (no error).
```
You don't need to check for folder exsitence i believe

engine/server/engine/centralized_logs/client_implementations/persistent_volume/persistent_volume_logs_database_client.go
```diff
@@ -138,7 +126,7 @@ func (client *persistentVolumeLogsDatabaseClient) FilterExistingServiceUuids(
 // ====================================================================================================
 func streamServiceLogLines(
```
gbouv: Does this function need to be static? Since it now accepts the entire client object, can't we just make it a class method?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/consts.go
```diff
@@ -21,17 +21,19 @@ const (
 	fileSinkIdSuffix = "file"
 	fileTypeId       = "\"file\""
 
-	// We store log files in the volume per-enclave, per-service
+	// We instruct vector to store log files per-week (00-53), per-enclave, per-service
 	// To construct the filepath, we utilize vectors template syntax that allows us to reference fields in log events
 	// https://vector.dev/docs/reference/configuration/template-syntax/
-	uuidLogsFilepath = "\"" + logsStorageDirpath + "{{ enclave_uuid }}/{{ service_uuid }}.json\""
+	baseLogsFilepath = "\"" + logsStorageDirpath + "%%U/"
```
gbouv: I suppose the `%%U` is a placeholder for the week number, but we probably also need the year in there no? I know it's quite unlikely, but if at any point we have an engine which runs for more than 1 year, you will have conflicts

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
```diff
@@ -0,0 +1,42 @@
...
+	// compute the next oldest week
+	diff := week - (consts.LogRetentionPeriodInWeeks + 1) - 1
+	var weekPastRetentionPeriod int
+	if diff >= 0 {
+		weekPastRetentionPeriod = diff
+	} else {
+		weekPastRetentionPeriod = maxWeekNum + diff
+	}
+
+	// create directory path for that week
```
gbouv: `create`? Aren't we removing it here?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_file_stream_logs_strategy.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/kurtosis-tech/kurtosis/engine/server/engine/centralized_logs/client_implementations/persistent_volume/consts"
+	"github.com/kurtosis-tech/kurtosis/engine/server/engine/centralized_logs/client_implementations/persistent_volume/volume_filesystem"
+	"github.com/kurtosis-tech/kurtosis/engine/server/engine/centralized_logs/logline"
+	"github.com/kurtosis-tech/stacktrace"
+	"github.com/sirupsen/logrus"
+	"io"
+	"strings"
+)
+
+// This strategy pulls logs from filesytsem where there is a log file per enclave, per service
```
gbouv, line 20: Interesting. The reason we need both is for backward compat?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -0,0 +1,206 @@
...
+// eg.
+// [.../28/d3e8832d671f/61830789f03a.json] is the file containing logs from service with uuid 61830789f03a, in enclave with uuid d3e8832d671f,
+// in the 28th week of the current year
+type PerWeekStreamLogsStrategy struct {
+	// Current week of the year based in range 00-53
+	currentWeek int
+}
+
+func NewPerWeekStreamLogsStrategy(currentWeek int) *PerWeekStreamLogsStrategy {
+	return &PerWeekStreamLogsStrategy{
```
gbouv, line 38: Wait how does that work in practice? I see this is called in engine main.go. What happens if I start my engine a Sunday at 11:59PM? This currenWeek will be wrong for the entire lifetime of my engine. What will be the consequences?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -0,0 +1,206 @@
...
+// going back ([retentionPeriodInWeeks] + 1) back from [currentWeek].
+// Notes:
+// - File paths are of the format '/week/enclave uuid/service uuid.json' where 'week' is %W strftime specifier
+// - The +1 is because we retain an extra week of logs compared to what we promise to retain for safety.
+// - The list of file paths is returned in order of oldest logs to most recent logs e.g. [ 3/80124/1234.json, /4/801234/1234.json, ...]
+// - If a file path does not exist, the function with exits and returns whatever file paths were found
+func getRetainedLogsFilePaths(
+	filesystem volume_filesystem.VolumeFilesystem,
+	retentionPeriodInWeeks int,
+	currentWeek int,
```
gbouv: Same remark as above, why does this method need to be static if you're taking as input `currentWeek`, which is the only attribute of you *Strategy class?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -0,0 +1,206 @@
...
+	ctx context.Context,
+	fs volume_filesystem.VolumeFilesystem,
+	logsByKurtosisUserServiceUuidChan chan map[service.ServiceUUID][]logline.LogLine,
+	streamErrChan chan error,
+	enclaveUuid enclave.EnclaveUUID,
+	serviceUuid service.ServiceUUID,
+	conjunctiveLogLinesFiltersWithRegex []logline.LogLineFilterWithRegex,
+	shouldFollowLogs bool,
+) {
+	paths := getRetainedLogsFilePaths(fs, consts.LogRetentionPeriodInWeeks, strategy.currentWeek, string(enclaveUuid), string(serviceUuid))
```
gbouv: That's what I thought, I'm pretty sure this doesn't work if I start my engine on a Sunday at 23:59PM

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -0,0 +1,206 @@
...
+			logrus.Debugf("Context was canceled, stopping streaming service logs for service '%v' in enclave '%v", serviceUuid, enclaveUuid)
+			return
+		default:
+			var jsonLogStr string
+			var err error
+			var readErr error
+			var jsonLogNewStr string
+			var shouldReturnAfterStreamingLastLine = false
+
+			for {
```
gbouv, line 95: Isn't this the exact same as the `per_file_stream_logs_strategy`? It seems very similar at first glance, can't we DRY this a little bit?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -0,0 +1,206 @@
...
+	if len(paths) == 0 {
+		streamErrChan <- stacktrace.NewError(
+			`No logs file paths for service '%v' in enclave '%v' were found. 
+					This is a bug, indicating either:
+					1) Logs for the current week are not being stored.
+					2) Logs were manually removed.`,
+			serviceUuid, enclaveUuid)
+		return
+	}
+	if len(paths) < consts.LogRetentionPeriodInWeeks+1 {
```
gbouv: Do we need to postfilter log lines that are older than the retention periodm and not send them to the stream?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// get log file paths as far back as they exist
+	for i := 0; i < (retentionPeriodInWeeks + 1); i++ {
+		diff := currentWeek - i
+		var pathWeekStr string
+		if diff >= 0 {
+			pathWeekStr = strconv.Itoa(diff)
+		} else {
+			pathWeekStr = strconv.Itoa(maxWeekNum + diff)
+		}
```
gbouv: same remark as above here, this is way too complex for a human, let's have this logic handled by golang `time` lib

engine/server/engine/centralized_logs/client_implementations/persistent_volume/volume_filesystem/volume_filesystem.go
```diff
@@ -15,12 +15,16 @@ const (
 // primarily for the purpose of enabling unit testing persistentVolumeLogsDatabaseClient
 type VolumeFilesystem interface {
```
gbouv, line 16: I am not sure I understand the purpose of this object TBH. How does the two function below differ from `os.Open` and `os.Stat`?

engine/server/engine/main.go
```diff
@@ -272,3 +286,8 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
 	output.WriteString("]")
 	return output.String()
 }
+
+func getCurrentWeek() int {
+	_, week := time.Now().UTC().ISOWeek()
```
gbouv: nit: wrapping this inside its own function seemd a bit overkill IMO

engine/server/engine/server/engine_connect_server_service.go
gbouv: If this works already we can keep it, it's good to be fully backward compatible, and we can remove it in a few weeks from now. Maybe stick a TODO somewhere saying that past 2023-09-XX, this can be removed

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_file_stream_logs_strategy.go
tedim52: ye that was the intent

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
tedim52: ah I meant creating the directory path string, but that's a bit confusing I'll change

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
tedim52: Manually tested this, by changing the times around a bit, but I'll add some unit tests.

engine/server/engine/centralized_logs/client_implementations/persistent_volume/consts/consts.go
tedim52: They initially were, but I found a lot of the consts being reused throughout this package. I will rename to be more specific though.

engine/server/engine/server/engine_connect_server_service.go
tedim52: sounds good.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
tedim52: It is once per week. So in the scenario you're mentioning - logs existing past the retention period would get removed the next Sunday. I can easily make this once a day though

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
tedim52: lol very valid point I'll change

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
tedim52: The primary differences are the way the log file path is created and log reader is created (we use a `MultiReader` to combine readers of all the files in `per_week`). There is also a small difference in error handling logic because of how the `MultiReader` works (maybe I can get this in parity though) - I'd prefer not to DRY right now but if you feel strongly, I don't mind.

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
tedim52: big bug, this is fixed

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
tedim52: changed to use ticker instead as discussed

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
tedim52: changed to use ticker every six hours as discussed

engine/server/engine/centralized_logs/client_implementations/persistent_volume/volume_filesystem/volume_filesystem.go
tedim52: explained over g meet

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
tedim52: discussed over g meet - going to do this

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -27,6 +30,11 @@ const (
 	maxWaitForEngineAvailabilityRetries         = 10
 	timeBetweenWaitForEngineAvailabilityRetries = 1 * time.Second
 	logsStorageDirpath                          = "/var/log/kurtosis/"
+	removeLogsWaitHours                         = 6
```
gbouv: nit: you can directly do `6 * time.Hour` and avoid having `*Hours` in your variable name (which will be a duration)

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -27,6 +30,11 @@ const (
...
+)
+
+var (
+	logRemovalTicker = time.NewTicker(removeLogsWaitHours * time.Hour)
```
gbouv: do you need that to be a static variable here? Can't you create it where you need it?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -100,6 +108,18 @@ func CreateEngine(
...
 			"An error occurred attempting to create logging components for engine with GUID '%v' in Docker network with network id '%v'.", engineGuidStr, targetNetworkId)
 	}
+
+	// schedule log removal for log retention
+	go func() {
+		osFs := volume_filesystem.NewOsVolumeFilesystem()
+		realTime := logs_clock.NewRealClock()
+		logRemover := log_remover.NewLogRemover(osFs, realTime)
+
+		for range logRemovalTicker.C {
```
gbouv, line 117: Reading the docs of time.Ticket, it seems the ticker will first wait for the duration to elapse before ticking for the first time. So if users always uses the engine for less than 6 hours, you'll never cleanup the logs. We probably want to do a first cleanup immediately when the engine starts

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/consts.go
```diff
@@ -21,17 +21,19 @@ const (
...
+	uuidLogsFilepath = baseLogsFilepath + "{{ enclave_uuid }}/{{ service_uuid }}.json\""
 
 	// Right now, we store duplicate logs files by service name and service's shortened uuid
 	// This is to enable retrieving logs by name and short uuid after enclaves have been stopped
 	// As right now, when an enclave is stopped all identifier info is lost
 	// TODO: Find a better way to capture or persist identifier info without storing duplicate log files
-	nameLogsFilepath      = "\"" + logsStorageDirpath + "{{ enclave_uuid }}/{{ service_name }}.json\""
-	shortUUIDLogsFilepath = "\"" + logsStorageDirpath + "{{ enclave_uuid }}/{{ service_short_uuid }}.json\""
+	nameLogsFilepath      = baseLogsFilepath + "{{ enclave_uuid }}/{{ service_name }}.json\""
+	shortUUIDLogsFilepath = baseLogsFilepath + "{{ enclave_uuid }}/{{ service_short_uuid }}.json\""
```
gbouv, line 36: out of curiosity, why do we use short uuids here and not the long ones? long ones have less chance of collisions, and I imagine there's not a big difference as to which one we're using here. Curious why we did not fallback on the safest option.

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/kurtosis-tech/kurtosis/engine/server/engine/centralized_logs/client_implementations/persistent_volume/logs_clock"
+	"github.com/kurtosis-tech/kurtosis/engine/server/engine/centralized_logs/client_implementations/persistent_volume/volume_consts"
+	"github.com/kurtosis-tech/kurtosis/engine/server/engine/centralized_logs/client_implementations/persistent_volume/volume_filesystem"
+	"github.com/sirupsen/logrus"
+	"strconv"
+	"time"
+)
+
+const (
+	oneWeekDuration = 7 * 24 * time.Hour
```
gbouv: nit: You variable is already of type "duration" here, no need to add it to its name

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+// Run implements the Job cron interface. It removes logs a week older than the log retention period.
+func (remover LogRemover) Run() {
+	// [LogRetentionPeriodInWeeks] weeks plus an extra week of logs are retained so remove logs a week past that, hence +1
+	numWeeksBack := volume_consts.LogRetentionPeriodInWeeks + 1
+
+	// compute the next oldest week
+	_, weekToRemove := remover.time.Now().Add(time.Duration(-numWeeksBack) * oneWeekDuration).ISOWeek()
```
gbouv: I believe this should work:
```
.Add(-numWeeksBack * oneWeekDuration)
```

`time.Duration(-numWeeksBack)` is actually super confusing when you know what a duration is in Golang (i.e. just an nanosecond count), b/c this literally represents 4 nanoseconds

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// remove directory for that week
+	oldLogsDirPath := fmt.Sprintf("%s%s/", volume_consts.LogsStorageDirpath, strconv.Itoa(weekToRemove))
```
gbouv: Wait, does that work? Didn't you just add the year in the logs dir path above line 27 of the consts.go file? I'm super confused...

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// log remover should remove week 49 logs
+	logRemover.Run()
+
+	_, err := mockFs.Stat(week49filepath)
+	require.Error(t, err) // check the file doesn't exist
+}
+
+func getWeekFilepathStr(week int) string {
+	return fmt.Sprintf("%s%s/%s/%s%s", logsStorageDirpathForTests, strconv.Itoa(week), testEnclaveUuid, testUserService1Uuid, volume_consts.Filetype)
```
gbouv: Same here like I'm a bit confused, is that really the path that will get produced by vector, even with the `%%Y` you just added? Where is the year here?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -0,0 +1,221 @@
...
+					1) Logs for the current week are not being stored.
+					2) Logs were manually removed.`,
+			serviceUuid, enclaveUuid)
+		return
+	}
+	if len(paths) < volume_consts.LogRetentionPeriodInWeeks+1 {
+		logrus.Warnf(
+			`We expected to retrieve logs going back '%v' weeks, but instead retrieved logs going back '%v' weeks. 
+					This indicates either:
+					1) The enclave has not been running longer than the log retention period.
```
gbouv: nit: is that really a warn then? I imagine this will be pretty frequent. I'd be more worried though is `len(paths) > volume_consts.LogRetentionPeriodInWeeks+1`

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
gbouv: ACK, I guess it's fine given that at some point we will remove the other one

engine/server/go.mod
```diff
@@ -82,6 +82,7 @@ require (
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/morikuni/aec v1.0.0 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/robfig/cron/v3 v3.0.0 // indirect
```
gbouv: do you still need this in here?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/consts.go
tedim52: we do store the longer uuid, this pr has context on why the short uuid is stored: https://github.com/kurtosis-tech/kurtosis/pull/1213

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
tedim52: This is a bug. I forgot to update the log remover dirpath to include the year. I refactored `consts` so every filepath depends on the same format and this doesn't happen.

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover_test.go
tedim52: No - added the year.

engine/server/go.mod
tedim52: no gonna remove this package.

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
tedim52: good point,  I'll change the condition and message.

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
tedim52: it actually doesn't work weirdly due to `Cannot use '-numWeeksBack * oneWeek' (type int) as the type Duration` - even though `7 * 24 * time.Hour` works...

tedim52: FWIW, casting to time.Duration should still work though, [`time.Duration`](https://pkg.go.dev/time#Duration) is just an int64 underlying so casting `numWeeksBack` to a duration, has still has the effect of multiplying `oneWeekDuration` by 4 even if it technically represents 4 nanoseconds (but you may not have been disputing that)

gbouv: This is weird... This does seem to work here: https://go.dev/play/p/tKH0tLEQFM8?v=goprev
Anyway, not the end of the world...

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// log remover should remove week 49 logs
+	logRemover.Run()
+
+	_, err := mockFs.Stat(week49filepath)
+	require.Error(t, err) // check the file doesn't exist
+}
+
+func getWeekFilepathStr(year, week int) string {
+	return fmt.Sprintf(volume_consts.PerWeekFilePathFmtStr, logsStorageDirpathForTests, strconv.Itoa(year), strconv.Itoa(week), testEnclaveUuid, testUserService1Uuid, volume_consts.Filetype)
```
gbouv: Since you're implementing an entire function for this `Sprintf`, implement it in log_remover.go and call it from the Run() and from your tests, this way you're sure you're testing the actual behavior!!!

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_file_stream_logs_strategy.go
```diff
@@ -0,0 +1,124 @@
...
+	fs volume_filesystem.VolumeFilesystem,
+	logsByKurtosisUserServiceUuidChan chan map[service.ServiceUUID][]logline.LogLine,
+	streamErrChan chan error,
+	enclaveUuid enclave.EnclaveUUID,
+	serviceUuid service.ServiceUUID,
+	conjunctiveLogLinesFiltersWithRegex []logline.LogLineFilterWithRegex,
+	shouldFollowLogs bool,
+) {
+	// logs are stored per enclave id, per service uuid, eg. <base path>/123440231421/54325342w2341.json
+	logsFilepath := fmt.Sprintf("%s%s/%s%s", volume_consts.LogsStorageDirpath, string(enclaveUuid), string(serviceUuid), volume_consts.Filetype)
```
gbouv: have a constant for this one as well maybe. live computing those log file paths all over the place is suuuper fragile and worrying me a little TBH

gbouv (approved):

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_remover/log_remover.go
tedim52: so `numWeeksBack` gets cast as an int so there's a mismatch when doing `.Add(-numWeeksBack * oneWeek)` but `-4` is an untyped constant that gets cast to a Duration when `.Add(-4 * oneWeek) is evaluated.

PR #1243, opened 2023-09-06
fix: handle default string value properly

Peeeekay:
This PR fixes the bug found recently where the default value in the starlark package was being overwritten by the "" passed by the frontend.


engine/frontend/src/component/PackageCatalogForm.js
```diff
@@ -101,6 +101,20 @@ const checkValidIntType = (data) => {
...
+        return (/^-?[\d]*(\.[\d]+)?$/g).test(value);
+    }
+    if (data === "undefined") {
+        return false
+    }
+    return isValidFloat(data)
+}
+
+const checkValidBooleanType = (data) => {
+    return ["TRUE", "FALSE"].contains(data)
```
adschwartz: @Peeeekay I'd recommend lowercasing the data to make sure you can match invariantly to casing

engine/frontend/src/component/PackageCatalogForm.js
```diff
@@ -101,6 +101,20 @@ const checkValidIntType = (data) => {
     }
 }
 
+const checkValidFloatType = (data) => {
+    const  isValidFloat = (value) => {
+        return (/^-?[\d]*(\.[\d]+)?$/g).test(value);
```
adschwartz: @Peeeekay It's not wrong to use a regex per se (I haven't tested the validity of this), but can we not use Number.parsefloat or [what this guy is saying](https://stackoverflow.com/a/52740766)?

engine/frontend/src/component/PackageCatalog.js
```diff
@@ -45,12 +45,12 @@ const PackageCatalog = ({kurtosisPackages: defaultPackages}) => {
         }
         const filteredPackages = defaultPackages.filter(pack => {
                 if ("name" in pack) {
-                    return pack.name.includes(value)
+                    // lowercase everything so that it works for both cases
+                    return pack.name.toLowerCase().includes(value.toLowerCase())
```
adschwartz: @Peeeekay maybe also add left and right trim to remove any space?

adschwartz (changes requested):
Couple of comments to look at. What do you think about adding some unit tests for this to catch any obvious mistakes?

engine/frontend/src/component/PackageCatalogForm.js
Peeeekay: I have used the regex that was was the second answer in the slack link. IsNaN has an edge-case where "" returns true; for our initial usecases the regex seems appealing. Parsefloat also has issues where it returns true for some of failure modes. @adschwartz

adschwartz: I think we can just check for an empty string no? i.e. if you get `""` then it's not a float. We should do early check for empty string null/undefined as well.

Peeeekay: hm ok; i can do that.

Peeeekay:
> Couple of comments to look at. What do you think about adding some unit tests for this to catch any obvious mistakes?

@adschwartz is it cool if i do that in another PR because it would require me to add some plumbing - but this can unblock nam.

engine/frontend/src/component/PackageCatalogForm.js
Peeeekay: oh nevermind; i was looking at different link; doing it rn

engine/frontend/src/component/PackageCatalog.js
```diff
@@ -45,12 +45,13 @@ const PackageCatalog = ({kurtosisPackages: defaultPackages}) => {
         }
         const filteredPackages = defaultPackages.filter(pack => {
                 if ("name" in pack) {
-                    return pack.name.includes(value)
+                    // lowercase everything so that it works for both cases
+                    const trimmedValue = value.trim()
```
adschwartz, line 49: @Peeeekay do we need to check for undefined and null pointer?

adschwartz: Same comment other places

adschwartz (approved):
@Peeeekay Unblocking you but I think we should follow up with some increased quality and a unit test if possible.

engine/frontend/src/component/PackageCatalog.js
Peeeekay: The default value is `""`; the initial state for all the inputs is an empty string.. so it shouldn't be an issue

PR #1248, opened 2023-09-06
docs: update container registry docs

tedim52:
## Description:
Updates docs to reflect that we can indeed pull images from public container registries, just not private ones.

YES


leeederek (approved):
Can you just add a 1 liner in the description next time? I know this change is trivial, but making our PRs consistent is helpful and good habit to get in to :)

PR #1249, opened 2023-09-06
fix: the runtime value store now supports `starlark.Bool` value types

leoporoli:
## Description:
the runtime value store now supports `starlark.Bool` value types

NO

to fix this issue: https://app.circleci.com/pipelines/github/kurtosis-tech/awesome-kurtosis/1069/workflows/5a87b29f-1ab0-4ed1-94fa-1640ec5b329d/jobs/3297


core/server/api_container/server/startosis_engine/runtime_value_store/recipe_result_repository.go
```diff
@@ -143,7 +143,12 @@ func (repository *recipeResultRepository) Get(
 			var comparableValue starlark.Comparable
 			comparableInt, err := strconv.Atoi(stringifiedComparable)
 			if err != nil {
-				comparableValue = starlark.String(stringifiedComparable)
```
victorcolombo, line 146: Can we reduce the nesting here?

Creating a function returning `Comparable, error` with the following structure:

```
comparableInt, err := strconv.Atoi(stringifiedComparable)
if err != nil {
    return comparableInt
}
comparableBool, err := strconv.ParseBool(stringifiedComparable)
if err != nil {
    return comparableBool
}
...
```

victorcolombo (approved):

PR #1250, opened 2023-09-06
feat: folks can delete enclaves from the frontend

Peeeekay:
Users can delete enclave(s) from the frontend once this change get merged in!


engine/frontend/src/component/Home.js
```diff
@@ -39,6 +38,16 @@ const Home = () => {
     const [enclaveLoading, setEnclaveLoading] = useState(false)
     const {appData, setAppData} = useAppContext()
 
+    const handleDeleteClick = (enclaveName) => {
+        console.log("name ", enclaveName)
+        const filteredEnclaves = enclaves.filter(enclave => enclave.name !== enclaveName)
+        const makeRequest = async () => {
+            await removeEnclave(appData.jwtToken, appData.apiHost, enclaveName)
```
adschwartz: @Peeeekay can we listen to the response and determine if the removal was successful or not?

engine/frontend/src/component/Home.js
```diff
@@ -40,9 +40,14 @@ const Home = () => {
 
     const handleDeleteClick = (enclaveName) => {
         const makeRequest = async () => {
-            const filteredEnclaves = enclaves.filter(enclave => enclave.name !== enclaveName)
-            await removeEnclave(appData.jwtToken, appData.apiHost, enclaveName)
-            setEnclaves(filteredEnclaves)
+            try {
+                const filteredEnclaves = enclaves.filter(enclave => enclave.name !== enclaveName)
+                await removeEnclave(appData.jwtToken, appData.apiHost, enclaveName)
```
adschwartz, line 45: @Peeeekay doesn't removeEnclave return anything? THe endpoint must return an error code if it fails?

Peeeekay: No; the response failure normally are returned as errors - if you look at the proto schema it's just an empty response

engine/frontend/src/api/enclave.js
```diff
@@ -19,6 +19,13 @@ export const getEnclavesFromKurtosis = async (token, apiHost) => {
     return []
 }
 
+export const removeEnclave = async (token, apiHost, enclaveName) => {
+    const response = await removeEnclaveFromEnclaveManager(enclaveName, token, apiHost)
+    const enclave = response.enclaveInfo;
```
adschwartz, line 24: what does the `enclave` variable do?

adschwartz (approved):

engine/frontend/src/api/enclave.js
Peeeekay: just prints undefined, was using this for testing, forgot to remove it.

PR #1252, opened 2023-09-07
feat: CLI run command option to disable user services port forwarding

### LINKED ISSUE #1236: CLI option to disable user services port forwarding
### Background & motivation

User services port forwarding is done automatically even if some users don't need it.  Some users would like to be able to disable port forwarding and do it themselves for the required ports.  This is needed until we can provide a complete [Connect API](https://www.notion.so/kurtosistech/Upgraded-Connectivity-ITERATION-1-86693f7143ad49d8958f9abc671b8f4e) where the user has control over what gets forwarded.

### Desired behaviour

The user should be able to disable port forwarding by passing the flag `--no-connect` to the CLI run command.

laurentluce:
## Description:
Add a `--no-connect` flag to the CLI run command.  If set the user service ports are not forwarded locally.  In the case of K8S, the port forward logic in the Gateway is disabled and the user services public ip/port is not displayed in the enclave info.  In the case of Docker, the forward port calls are not sent to the Portal client but the ports are still exposed on the Docker instance since this is managed by Docker backend.  It is acceptable for now since the K8S use case is higher priority.  A subsequent PR will be created to clean up the Docker use case.

This change is part of the Connect API implementation: https://www.notion.so/kurtosistech/Upgraded-Connectivity-ITERATION-1-86693f7143ad49d8958f9abc671b8f4e

YES

Closes #1236


api/golang/core/lib/binding_constructors/binding_constructors.go
```diff
@@ -39,6 +39,7 @@ func NewRunStarlarkScriptArgs(
 	dryRun bool,
 	parallelism int32,
 	experimentalFeatures []kurtosis_core_rpc_api_bindings.KurtosisFeatureFlag,
+	connect kurtosis_core_rpc_api_bindings.Connect,
```
laurentluce: We use the same working `connect` as in the Connect API.

api/protobuf/core/api_container_service.proto
```diff
@@ -107,6 +107,17 @@ message ServiceInfo {
 // ==============================================================================================
 //                               Execute Starlark Arguments
 // ==============================================================================================
+
+// User services port forwarding
+enum Connect {
```
laurentluce, line 115: Use an enum to support future use cases described in the Connect API doc.

cli/cli/commands/run/run.go
```diff
@@ -144,14 +143,6 @@ var StarlarkRunCmd = &engine_consuming_kurtosis_command.EngineConsumingKurtosisC
 			Type:    flags.FlagType_Bool,
 			Default: fullUuidFlagKeyDefault,
 		},
-		{
-			Key: mapPortsFlagKey,
```
laurentluce, line 148: Remove this flag which disabled port forwarding in the Portal only.  This flag is not documented and not used at Kurtosis.  The new `no-connect` flag supersedes it.

cli/cli/commands/service/add/add.go
```diff
@@ -203,14 +200,6 @@ var ServiceAddCmd = &engine_consuming_kurtosis_command.EngineConsumingKurtosisCo
 			Type:    flags.FlagType_Bool,
 			Default: fullUuidFlagKeyDefault,
 		},
-		{
-			Key: mapPortsFlagKey,
```
laurentluce, line 207: Unused and undocumented flag.  We decided to not support `service add` with the new `no-connect` flag.

cli/cli/kurtosis_gateway/server/api_container_gateway/api_container_gateway_service_server.go
```diff
@@ -31,6 +31,9 @@ type ApiContainerGatewayServiceServer struct {
 	// ServiceMap and mutex to protect it
 	mutex                               *sync.Mutex
 	userServiceNameToLocalConnectionMap map[string]*runningLocalServiceConnection
+
+	// User services port forwarding
+	connect kurtosis_core_rpc_api_bindings.Connect
```
laurentluce: We store the connect attribute at the enclave level in the Gateway.  It is set in the RunStarlark functions and read in GetServices.  A cleaner way would be to add a Connect RPC method and not rely on GetServices.  The plan is to deprecate the Gateway and replace it with the Portal so we don't want to spend too much time here.

api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -84,9 +85,10 @@ func (enclaveCtx *EnclaveContext) RunStarlarkScript(
 	dryRun bool,
 	parallelism int32,
 	experimentalFeatures []kurtosis_core_rpc_api_bindings.KurtosisFeatureFlag,
+	connect kurtosis_core_rpc_api_bindings.Connect,
```
h4ck3rk3y: this ( and the relevant package call) would be a breaking change to the SDK we should rename the PR as such to reflect

it seems like we only change the non blocking endpoints and not the blocking SDK calls, perhaps as we only want to change the CLI. Should we be consistent?

cli/cli/commands/import/import.go
```diff
@@ -57,6 +57,7 @@ const (
 	doNotShowFullUuids    = false
 	doNotDryRun           = false
 	noParallelism         = 1
+	connectConnect        = kurtosis_core_rpc_api_bindings.Connect_CONNECT
```
h4ck3rk3y: Should we track a TODO here to make it passable overtime

cli/cli/commands/service/add/add.go
h4ck3rk3y: 1. Should we track a TODO to support this overtime?
2. What is the behavior if i create an enclave with `no-connect` and then do this? Happy for us to address this concern in a later PR

cli/cli/kurtosis_gateway/server/api_container_gateway/api_container_gateway_service_server.go
```diff
@@ -56,6 +59,7 @@ func NewEnclaveApiContainerGatewayServer(connectionProvider *connection.GatewayC
 		mutex:                               &sync.Mutex{},
 		userServiceNameToLocalConnectionMap: userServiceToLocalConnectionMap,
 		enclaveId:                           enclaveId,
+		connect:                             kurtosis_core_rpc_api_bindings.Connect_CONNECT,
```
h4ck3rk3y: nit can we rename this to `userServiceConnect` so that its clear that this is for user services and not the APIC container itself

h4ck3rk3y (approved):
Mostly looks good and I am happy with the incremental work + minimalism - somethings come to mind

1. Barnabas' as far as I remember uses `kurtosis enclave add` followed by run as he wants to spin up Chaos Mesh between enclave add & and run. Can we add no-connect to enclave add as well in this PR?
2. We need to update the SDK docs with the `no-connect` flag changes
3. Need to change the title as its a breaking change

h4ck3rk3y:
> Barnabas' as far as I remember uses kurtosis enclave add followed by run as he wants to spin up Chaos Mesh between enclave add & and run. Can we add no-connect to enclave add as well in this PR?

Resolved this on Slack. This is a non issue, the current codepath already tackles this!

cli/cli/commands/service/add/add.go
laurentluce: With this change, the ports for this newly added service won't be forwarded if the flag `no-connect` was passed to the previous `run` command.

laurentluce: > 1. Should we track a TODO to support this overtime?

From the Connect API design doc: "We could build kurtosis service connect, kurtosis port connect, etc. as shorthands for connecting an entire service or port if we need. At the moment, I'm not convinced that it's worth adding the extra surface area/mental load to users."

laurentluce: > 2. What is the behavior if i create an enclave with `no-connect` and then do this? Happy for us to address this concern in a later PR

With this change, the ports for this newly added service won't be forwarded if the flag no-connect was passed to the previous run command.

api/golang/core/lib/enclaves/enclave_context.go
laurentluce: I updated the non blocking calls.

cli/cli/commands/import/import.go
laurentluce: TODO added.

Issue #1103, opened 2023-08-11
Always pull newest image


barnabasbusa:
### Background & motivation

When a new enclave is created, a trial attempt to fetch the latest images should be made.

This would ensure that someone is not spending endless time troubleshooting why an image not working, then realizing that the image is outdated already. 

### Desired behaviour

I want to always run the latest images at a certain tag.

### How important is this to you?

Painful; the lack of this feature makes using Kurtosis frictionful.

mieubrisse:
Thanks for filing this @barnabasbusa ! How would you deal with the case where you've built a local version of an image that also exists on Dockerhub, and so pulling would overwrite your locally-built version?

barnabasbusa:
Locally built images are nothing special. If someone wants to test a local image they can tag it with anything they want to ensure it doesn’t clash with the same tag as one ok dockerhub.

mieubrisse:
Looking into this with @tedim52 ; I'd actually thought our current behaviour was to always try and pull the latest but seems like that's not the case

parithosh:
Just an FYI: Usually most tools don't pull `latest` by default if an image tagged `latest` exists in the local registry (docker-compose for e.g). Its considered bad practice in production systems unless you add an `image_pull: true` or something to explicitly force that behavior. 

However, I see kurtosis as a testing system and its incredibly valuable for test setups to have the behavior of pulling latest by default. I'm fine with it being a flag and enabled in all runs by default, allowing prod setups to disable it. If kurtosis wants to focus more on prod users, then perhaps a `mode: test/prod` would achieve the same for this and future defaults. Just my 2c on the topic :)

tedim52:
Thanks for weighing in here @parithosh ! We'll definitely keep that in mind as we address this : )

mieubrisse:
Yep @parithosh , thinking along very similar lines here :) Ideally, everybody in the world would version-tag their images and follow proper GitOps of updating everything. In practice, many people use `latest` despite it being bad practice. We want Kurtosis to be able to handle dev, test, and prod equally (the "write your environment definition once" ideal), and enclave modes are probably our ticket here (e.g. "Prod"-like enclaves could have extra safeguards added to them).

We're only now starting to support the Prod usecase, so for now we might just make the "always try to pull the latest" default and in the future give a flag in Prod enclaves to flip it off

mieubrisse:
[...]

PR #1267, opened 2023-09-08
feat: always pull latest image

tedim52:
## Description:
Prior to this PR, Kurtosis would use the local image if it was available, and if it wasn't, pulled the image from Dockerhub (or other container registry). This PR reverses that behavior by always attempting to pull the latest image, and using the local image, if available, only if the pull fails. For users, this means they will always be using the latest version of the images they're working with. 

Through testing, I've determined the impact on speed is negligible but I've kept a `FetchLocalImage` function with the old behavior in case we need to rollback this change. I've also removed a redundant `FetchImage` from the `StartServices` as it is already being called for each service during validation phase and in `CreateAndStartContainerArgs`.

NO

## References
https://github.com/kurtosis-tech/kurtosis/issues/1103


container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -1121,7 +1121,46 @@ func (manager *DockerManager) GetContainersByLabels(ctx context.Context, labels
...
+	// this allows us to be deterministic in our behaviour
+	if !strings.Contains(dockerImage, dockerTagSeparatorChar) {
+		dockerImage = dockerImage + dockerTagSeparatorChar + dockerDefaultTag
+	}
+	logrus.Infof("Checking if image '%v' is available locally...", dockerImage)
+	doesImageExistLocally, err := manager.isImageAvailableLocally(ctx, dockerImage)
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred checking for local availability of Docker image '%v'", dockerImage)
+	}
+	logrus.Infof("Is image available locally?: %v", doesImageExistLocally)
```
adschwartz: Is this for debugging?

tedim52: yes I should change that log level to trace

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -1121,7 +1121,46 @@ func (manager *DockerManager) GetContainersByLabels(ctx context.Context, labels
 	return result, nil
 }
 
+// [FetchImage] always attempts to retrieve the latest [dockerImage].
+// If retrieving the latest [dockerImage] fails, the local image will be used.
+// Returns error, if no local image is available after retrieving latest fails.
 func (manager *DockerManager) FetchImage(ctx context.Context, dockerImage string) error {
```
adschwartz, line 1127: @tedim52 is there any way to return the result of fetching the image? It seems a little strange to call this function but not knowing if it was fetched successfully or not (other than the absence of the error)? Maybe you are limited by what the `isImageAvailableLocally` and `pullImage` functions return, so maybe there's not much you can do, but maybe worth having a look at?

adschwartz (approved):

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: great point, send some ideas over slack -> lmk what you think and i'm happy to make this change. gonna merge for now and open another PR for this

mieubrisse:
I added "user-facing" because this does indeed affect users because it's a potentially-significant change in beahviour. @leeederek FSA for the announce!

PR #1278, opened 2023-09-08
fix: adjust em streaming logic to prevent panic

tedim52:
## Description:
This is my solution to fixing the panic bug. The panic bug was being cause by a send on a closed channel that was being used to send log lines to frontend. Instead of using the channel to send the log lines, I just iterate over the `engineClient` receiving messages and forward those messages to the stream, removing the channel, and thus removing the possibility of the panic entirely.

NO


enclave-manager/server/server.go
```diff
@@ -158,29 +158,31 @@ func (c *WebServer) GetServiceLogs(
...
+			err := engineClient.Close()
 			if err != nil {
 				logrus.Errorf("Error ocurred: %+v", err)
+				return err
 			}
-			close(logs)
 			return nil
-		case resp := <-logs:
+		default:
+			resp := engineClient.Msg()
```
Peeeekay, line 177: Does this work when the stream is open but no new logs are available yet? @tedim52

tedim52: Yep I just tested it. Good callout too - I looked into how [`Receive()`](https://pkg.go.dev/connectrpc.com/connect#ServerStreamForClient.Receive) is working - it only evaluates to false if an error occurs or stream is closed. So as long as streams open, that for loop will keep going, but I added a check afterwards to handle the error via `engineClient.Error()`.

tedim52:
Closing, as it was merged in the `-n X` PR.

PR #1290, opened 2023-09-11
ci: Require version bump if proto breaks

victorcolombo:
## Description:

This PR introduces a check that will force that any PR that changes protobuf in a breaking way MUST bump the minor version (by adding `!` on title). This avoids cases where API from Engine or APIC introduce breaks but the SDK is not aware, since minor and major are equal.

Screenshot 2023-09-11 at 19 43 58

NO


h4ck3rk3y (approved):

.github/workflows/check-proto-break.yml
```diff
@@ -0,0 +1,30 @@
...
+        fetch-depth: 0
+    - name: Check if PR is breaking
+      run: |
+        title="${{ github.event.pull_request.title }}"
+        if [[ "$title" == *'!'* ]]; then
+          echo "PR is a breaking change by title"
+          exit 0
+        else
+          echo "Checking if PR is a breaking change"
+          buf breaking --path ./api/protobuf/ --against '.git#branch=origin/main'
```
h4ck3rk3y, line 29: is this checking ou the right branch

.github/workflows/check-proto-break.yml
```diff
@@ -0,0 +1,30 @@
...
+  check-proto-break:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-go@v3
+      with:
+        go-version: 1.19
+    - run: go install github.com/bufbuild/buf/cmd/buf@v1.26.1
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
```
h4ck3rk3y, line 20: is this checking ou the right branch

.github/workflows/check-proto-break.yml
victorcolombo: I believe there is no material difference between picking main here and the latest tag

PR #1291, opened 2023-09-12
fix: Reset the module global cache on every new interpretation to avoid using outdated modules

### LINKED ISSUE #1226: updating image of `add_service` doesn't change the running service
### What's your CLI version?

0.82.18

1. Clone the eth-network-package locally - checking out current main 30d2ccf8feae4fadc859a19c4a0697f2659cc90b
2. `kurtosis run . --enclave test`
3. Change https://github.com/kurtosis-tech/eth-network-package/blob/main/package_io/input_parser.star#L2 to `ghcr.io/paradigmxyz/reth` or `ethereum/client-go:v1.12.2-arm64`
4. `kurtosis run . --enclave test`

You will see that the package just runs fine and the image isn't updated in either the reth case or in the v1.12.2-arm64 case. 

### Desired behavior

I'd expect `reth` to not start as flags would be wrong, and for `v1.12.2-arm64` I'd expect this image in `add_service`

gbouv:
## Description:
Using a single module cache lead to issues like [this one](https://github.com/kurtosis-tech/kurtosis/issues/1226) b/c the interpreter was pulling the outdated cache.

Fixes #1226 

NO

### DIFF
--- a/core/server/api_container/server/startosis_engine/startosis_interpreter.go
+++ b/core/server/api_container/server/startosis_engine/startosis_interpreter.go
@@ -51,14 +51,13 @@ var (
 
 type StartosisInterpreter struct {
 	// This is mutex protected as interpreting two different scripts in parallel could potentially cause
-	// problems with the moduleGlobalsCache & moduleContentProvider. Fixing this is quite complicated, which we decided not to do.
-	mutex              *sync.Mutex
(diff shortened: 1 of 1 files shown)


h4ck3rk3y (approved):
LGTM

PR #1295, opened 2023-09-13
refactor!: rename assert to verify

h4ck3rk3y:
## Description:

We have renamed `plan.assert` to `plan.verify`. Users are encouraged to replace any usage of `plan.assert` in their Starlark with `plan.verify`

YES/NO


h4ck3rk3y:
@leoporoli if you can review this before Monday then I can get it out in the next release cycle

leoporoli (approved):

Issue #1238, opened 2023-09-05
Page not found


Prathiksha-Nataraja:
### Where did you see the issue in our documentation?

https://docs.kurtosis.com/0.80.11/SDK

### Issue description

[Kurtosis SDK](https://docs.kurtosis.com/0.80.11/SDK) page not working

<img width="1440" alt="Screenshot 2023-09-05 at 5 28 57 PM" src="https://github.com/kurtosis-tech/kurtosis/assets/90592522/d7bd7f12-00af-400d-a8c0-48c9d568f299">

leeederek:
Thank you for surfacing this, @Prathiksha-Nataraja! I will look into this.

mieubrisse:
Verified with @leeederek that what's happening here is, Google is indexing old versions of the SDK and so folks arriving to the SDK from Google are getting stale versions that have been deleted

mieubrisse:
@h4ck3rk3y , there's probably an easy fix here which is, rather than deleting the oldest X versions chronologically, keep the latest `X.Y.Z`, `X.(Y-1).Z`, `X.(Y-2).Z`, etc. with the highest `Z` version. This would let us keep less versions (hopefully breaking Algolia less) and also mean that the old versions stick around for longer

leeederek:
Hey I'm not seeing this issue anymore - perhaps @h4ck3rk3y's removal of versioned docs fixed this :) closing out!

PR #1297, opened 2023-09-13
docs: rearrange root-level sidebar

galenmarchetti:
## Description:
Rearranging root-level sidebar to make reference material locations more obvious at a first glance, as a first step to more in detail reference doc revamp

YES


docs/docs/runtime-api-reference.md
```diff
@@ -1,7 +1,7 @@
 ---
-title: SDK
-sidebar_label: SDK
-slug: /SDK
+title: Runtime API Reference
```
leeederek: @galenmarchetti - NACK from me re: this name change. We should keep it as SDK.

Context: We previously had this section titled as `Client Libraries` and more than a few people asked me where our SDK was (Chainlink asked, and also a user asked too when we were doing a user feedback session). I changed the wording to SDK with Kevin's +1 [here](https://kurtosistech.slack.com/archives/C04MBCY49HR/p1686151655619479). Another data point we got recently was that our users search for the "Kurtosis SDK" on Google - https://github.com/kurtosis-tech/kurtosis/issues/1238

docs/sidebars.js
```diff
@@ -14,57 +14,63 @@
...
         {
             "type": "link",
             "label": "Kurtosis for Web3",
             "href": "https://web3.kurtosis.com",
         },
-        'quickstart',
-        'code-examples',
         {
             type: 'category',
-            label: 'Guides',
```
leeederek, line 27: Would prefer the `Guides` section be near the top of the sidebar, proposing that it be right below the `Quickstart`

leeederek (changes requested):
NACK from me on both the ordering and the name change from SDK --> runtime api reference

docs/docs/runtime-api-reference.md
h4ck3rk3y: Would prefer keeping the title as SDK as well

mieubrisse: +1

galenmarchetti: @h4ck3rk3y @leeederek  Roger - I'm looking for a specifier that identifies why "SDK" is different than _starlark_. With just SDK we don't give any information on a first glance about why there's a difference between starlark api and SDK. Thoughts on "runtime SDK" or <some other word> SDK that clarifies why we have an SDK _and_ a starlark API?

docs/sidebars.js
galenmarchetti: No issue with that will change

docs/docs/runtime-api-reference.md
h4ck3rk3y: Interesting, I see what you are getting at. I can get on board with `runtime SDK` - So Starlark is pure APIC no engine, it all runs inside an APIC, while the current SDK is both.

Starlark - InsdieEnclaveSDK
SDK - Everything but EnclaveSDK

leeederek (approved):

PR #1306, opened 2023-09-14
fix: revert always pull latest img

tedim52:
## Description:
Will now first check if image is available locally before trying to pull latest - reverting back to original functionality. Reverting because of slowdown experienced by users.

YES 

## References:
https://github.com/kurtosis-tech/kurtosis/pull/1267
https://github.com/kurtosis-tech/kurtosis/issues/1103

### DIFF
--- a/container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
+++ b/container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
@@ -1121,10 +1121,39 @@ func (manager *DockerManager) GetContainersByLabels(ctx context.Context, labels 
 	return result, nil
 }
 
-// [FetchImage] always attempts to retrieve the latest [dockerImage].
+// [FetchImage] uses the local [dockerImage] if it's available.
+// If unavailable, will attempt to fetch the latest image.
+// Returns error if local [dockerImage] is unavailable and pulling image fails.
+func (manager *DockerManager) FetchImage(ctx context.Context, dockerImage string) error {
+	// if the image name doesn't have version information we concatenate `:latest`
+	// this behavior is similar to CreateAndStartContainer above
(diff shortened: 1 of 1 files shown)


container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -1121,9 +1121,9 @@ func (manager *DockerManager) GetContainersByLabels(ctx context.Context, labels
 	return result, nil
 }
 
-// [FetchImage] always attempts to retrieve the latest [dockerImage].
-// If retrieving the latest [dockerImage] fails, the local image will be used.
-// Returns error, if no local image is available after retrieving latest fails.
+// [FetchImage] uses the local [dockerImage] if it's available.
```
tedim52, line 1124: context: every image pull call uses `FetchImage` function

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -1138,29 +1138,22 @@ func (manager *DockerManager) FetchImage(ctx context.Context, dockerImage string
...
 	}
 
 	return nil
 }
 
-// [FetchLocalImage] uses the local [dockerImage] if it's available.
-// If unavailable, will attempt to fetch the latest image.
-// Returns error if local [dockerImage] is unavailable and pulling image fails.
-func (manager *DockerManager) FetchLocalImage(ctx context.Context, dockerImage string) error {
+// [FetchLatestImage] always attempts to retrieve the latest [dockerImage].
```
tedim52, line 1153: this call isn't called by any function, just here so functionality exists if we want to add flag down the line

mieubrisse (approved):

PR #1312, opened 2023-09-15
feat: print a "downloaded container images" summary after pulling images from remote or locally

### LINKED ISSUE #1292: FR: print and persist ALL the downloaded container image tags on each run
### Background & motivation

Today, on every Kurtosis run at the validation step, the image tags that get pulled are printed in the command line _only if the images do not exist locally:_
image

But after they are downloaded, the text disappears. I would like to know, on each run, that Kurtosis is using the image tags I expected and so this would allow me to cancel the run earlier if I see that Kurtosis is using an image tag that I didnt expect.

### Desired behaviour

This Feature Request is to track a few improvements as requested by Barnabas:
1. Persist the log lines that read `"Downloading ethpandaops/etheruem-genesis-generator:2.0.0-rc6"` when Kurtosis is downloading the image. Right now, they disappear after a while.
2. When the download is complete, update the line with: `ethpandaops/etheruem-genesis-generator:2.0.0-rc6 download complete.`
3. Show *ALL* the image tags - right now, Kurtosis only prints 4 lines max but if I am using 7 images, I want to see Kurtosis download all 7 and show that the downloads are complete.

When the image is built locally, I do not expect to see Kurtosis say anything about those image tags.

leoporoli:
## Description:
print a "downloaded container images" summary after pulling images from remote or locally

Final result in terminal:
python3 11-alpine - local

YES

Fix #1292


leoporoli:
it got merged in #1312

PR #1319, opened 2023-09-18
feat: Add format flag to kurtosis port print

### LINKED ISSUE #1316: Add --format flag to `kurtosis port print`
### Background & motivation

`kurtosis port print` should be more friendly for tooling, allowing easy piping into another UNIX commands

Motivated by:
https://github.com/gobitfly/eth2-beaconchain-explorer/issues/2548

### Desired behaviour

`--format protocol,ip,number` would be the default, and print http://127.0.0.1:1200/. if the port doesn't have a protocol, it's skipped.
`--format ip,number` prints 127.0.0.1:1200
`--format number` prints 1200
`--format ip` prints 127.0.0.1
`--format protocol` would print http if the port has an application protocol, and emptystring if not (still a success 0 exit code though)

victorcolombo:
## Description:

Allows for customization via `--format` flag of the pieces to be shown. Options are:
1. Protocol
2. Port
3. Ip

YES

Closes https://github.com/kurtosis-tech/kurtosis/issues/1316


cli/cli/commands/port/print/print.go
```diff
@@ -42,7 +48,14 @@ var PortPrintCmd = &engine_consuming_kurtosis_command.EngineConsumingKurtosisCom
 	LongDescription:           "Get information for port using port id",
 	KurtosisBackendContextKey: kurtosisBackendCtxKey,
 	EngineClientContextKey:    engineClientCtxKey,
-	Flags:                     []*flags.FlagConfig{},
+	Flags: []*flags.FlagConfig{
+		{
+			Key:     formatFlagKey,
+			Usage:   "Allows selecting what pieces of port are printed, using comma separated values (examples: 'ip', 'number', 'protocol,ip'). Default 'protocol,ip,number'.",
```
mieubrisse: Can we make these named constants, that correspond to the values down below?

cli/cli/commands/port/print/print.go
```diff
@@ -113,13 +131,32 @@ func run(
...
 	out.PrintOutLn(fullUrl)
 	return nil
 }
+
+func formatOutput(format string, ipAddress string, spec *services.PortSpec) (string, error) {
+	parts := strings.Split(format, ",")
+	var resultParts []string
+	for _, part := range parts {
+		switch part {
+		case "protocol":
```
mieubrisse: Same - can we make these named constants so we're DRY?

cli/cli/commands/port/print/print.go
```diff
@@ -113,13 +131,32 @@ func run(
...
+			if spec.GetMaybeApplicationProtocol() != "" {
+				resultParts = append(resultParts, spec.GetMaybeApplicationProtocol()+"://")
+			} else {
+				logrus.Warnf("Expected protocol but was empty, skipping")
```
mieubrisse: Like the idea, but we should remove this else it's going to pollute the output and prevent users from doing things like `curl $(kurtosis port print) ...`

mieubrisse (approved):

cli/cli/commands/port/print/print.go
victorcolombo: In all the other commands in the CLI do not have their usage string extracted to constants: https://github.com/kurtosis-tech/kurtosis/blob/main/cli/cli/commands/run/run.go#L109-L113

I ended up going for keeping the pattern as we have it currently.

victorcolombo: Make sense, will do

cli/cli/commands/port/print/print.go
victorcolombo: Will do

cli/cli/commands/port/print/print.go
victorcolombo: Yeah this is pretty awkward. The idea that the default is to contain the application protocol that might not be present, leaves us on a spot where we cannot just ignore it, nor error it (since the default should never error IMO.

A workaround here would be to pipe stderr to `/dev/null` on `$()` which could be fine. Alternatively the user could specify the format that makes sense for the port (omitting the protocol if the protocol is indeed not present).

WDYT @mieubrisse ?

victorcolombo: Added this as a TODO item and moved to debug so we still have the log line if needed

PR #1322, opened 2023-09-18
feat: changes to the package manager config and the files artifact view

adschwartz:
## Description:
This change improves the UX of the package manager and the files artifact view. It's not a breaking change

YES


engine/frontend/src/component/CodeEditor.js
```diff
@@ -0,0 +1,213 @@
...
+    const getEditor = () => {
+        if (!monacoRef.current) return null;
+        return monacoRef.current.editor.getEditors()[0];
+    }
+
+    const isEditorReadOnly = () => {
+        try {
+            return getEditor().getOption(monacoReadOnlyEnumId)
+        } catch (e) {
+            return undefined
```
Peeeekay, line 38: is it necessary to return undefined ?? also should we instead just log the error message here

Peeeekay: maybe return false or null.

Peeeekay (approved):

engine/frontend/src/component/CodeEditor.js
adschwartz: The use case for this to throw an exception is because the editor is not ready yet. In which case I don't want to return false becasue we actually don't know if it's read only or not. We could return null, but I feel undefined captures it better since we don't actually know until the editor is ready. Let me know if you think it's bad practice to use undefined, or I'm overlooking something else!

PR #1340, opened 2023-09-19
build: support arm64 via multi arch builds of engine/apic/files-artifacts-expander

h4ck3rk3y:
## Description:
We bundled the amd64 binary onto a fake arm64 image

The image is now real arm64 as of #1317
This fixes the binary part

YES

Needs no docs though as we now will truly support arm64


.circleci/config.yml
```diff
@@ -942,6 +942,10 @@ jobs:
       - run: |
           skip_docker_image_building=true
           core/files_artifacts_expander/scripts/build.sh "${skip_docker_image_building}"
+      - run: |
```
h4ck3rk3y, line 945: we only add this to the publish step as the rest of the CI runs with `amd64`

.circleci/config.yml
```diff
@@ -980,6 +984,10 @@ jobs:
       - run: |
           skip_docker_image_building=true
           core/server/scripts/build.sh "${skip_docker_image_building}"
+      - run: |
+          skip_docker_image_building=true
+          build_architecture=arm64
+          core/server/scripts/build.sh "${skip_docker_image_building}" ${build_architecture}
```
adschwartz: @h4ck3rk3y are you intentionally missing the quotes? In this case it won't matter much but I like to add just as a general best practice.

core/files_artifacts_expander/scripts/build.sh
```diff
@@ -28,6 +29,11 @@ if [ "${skip_docker_image_building}" != "true" ] && [ "${skip_docker_image_build
     echo "Error: Invalid skip-docker-image-building arg '${skip_docker_image_building}'" >&2
 fi
 
+architecture_to_build="${2:-"${DEFAULT_ARCHITECTURE_TO_BUILD}"}"
+if [ "${architecture_to_build}" != "amd64" ] && [ "${architecture_to_build}" != "arm64" ]; then
```
adschwartz, line 33: @h4ck3rk3y are there cases when we'd want to specify `arm64/v7` or `arm64/v8`? In that case we'd error out. not a huge issue and you can skip if you don't think it's worth checking if the string begins with arm64.

adschwartz (approved):

.circleci/config.yml
h4ck3rk3y: It was a mistake! Thanks :)

gbouv (approved):
At this point it might be better to just use go-releaser. It will build everything we need with a single command. But I guess that's an acceptable alternative

core/files_artifacts_expander/scripts/build.sh
gbouv: those arch distinctions are for docker. For Go it doesn't make a difference AFAICT

h4ck3rk3y: +1 ^, resolving this

PR #1343, opened 2023-09-19
fix: scan for first week of existing logs

tedim52:
## Description:
Prior to this change, if there were no logs for the current week, we'd return saying logs did not exist even though they do - logs for prior weeks could exist. This PR fixes this bug by scanning for the first week of existing logs, then scanning for as many week of logs exist beyond that, within the retention period. 

YES


engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -162,8 +162,20 @@ func (strategy *PerWeekStreamLogsStrategy) getRetainedLogsFilePaths(
 	enclaveUuid, serviceUuid string) []string {
 	var paths []string
 
-	// get log file paths as far back as they exist
+	// scan for first existing log file
+	firstWeekWithLogs := 0
 	for i := 0; i < (retentionPeriodInWeeks + 1); i++ {
+		year, week := strategy.time.Now().Add(time.Duration(-i) * oneWeek).ISOWeek()
```
adschwartz: @tedim52 is it possible this can behave unexpectedly during winter/summer time transitions, leap years etc? May not be problematic but potentially worth making a function for this and running a few unit tests to make sure it works as you expect.

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -162,8 +162,20 @@ func (strategy *PerWeekStreamLogsStrategy) getRetainedLogsFilePaths(
...
+		filePathStr := fmt.Sprintf(volume_consts.PerWeekFilePathFmtStr, volume_consts.LogsStorageDirpath, strconv.Itoa(year), strconv.Itoa(week), enclaveUuid, serviceUuid, volume_consts.Filetype)
+		if _, err := filesystem.Stat(filePathStr); err == nil {
```
adschwartz, line 173: @tedim52 is there a difference between a file not existing and a "real" error happening during `stat`? If there is we may want to handle the `err != nil` case as well

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
adschwartz: This is nitpicking but it's a standard I've adopted. I like to place calls to get the current time outside of loops unless I truly want to get a new Now everytime. If you are not intending to get a new Now it could lead to bugs. Like what if the user made this call 0.1 sec before midnight? :) 

Proposal: Put `strategy.time.Now()` out side the loop and `Add` inside the loop.

adschwartz: I guess if you don't do this and the user crosses over midnight they might get the logs for the same week twice?

adschwartz (approved):
@tedim52 looks good with a few comments

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
tedim52: very good point, will make this fix

tedim52: Good callout. This line primarily depends on the `year` and `week` returned by [`ISOWeek()`](https://pkg.go.dev/time#Time.ISOWeek). `ISOWeek` will handle decrementing weeks across a year correctly, `(year, 1) -> (year-1, 52)` which I've got a unit test for. The only interesting things that can happen here are that some years have either 52 or 53 weeks, but `ISOWeek` will still handle this transition. I'll add a test for the 53 week case just to be sure!

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
tedim52: Yep, good to handle in case. Done.

PR #1350, opened 2023-09-19
feat: Support YAML as Package param

### LINKED ISSUE #581: `args` should be specifiable as YAML
### Background & motivation

It's really, really annoying that I can only pass in JSON, because it doesn't allow me to comment my args. This becomes especially painful for large params, like the eth2-package

### Desired behaviour

`kurtosis run . "$(cat /tmp/params.yml)"` will work as expected

victorcolombo:
## Description:

Add support for providing YAML instead of JSON in both SDK and CLI:
`kurtosis run github.com/kurtosis-tech/datastore-army-package 'num_datastores: 2 # THIS IS A COMMENT'`
YES

Closes https://github.com/kurtosis-tech/kurtosis/issues/581


cli/cli/commands/run/run.go
```diff
@@ -551,17 +552,21 @@ func getOrCreateEnclaveContext(
...
-		return stacktrace.Propagate(err, "Error validating args, likely because it is not a valid JSON.")
+	var jsonError error
+	if jsonError = json.Unmarshal([]byte(serializedArgs), &result); jsonError == nil {
+		return nil
 	}
-	return nil
+	if err := yaml.Unmarshal([]byte(serializedArgs), &result); err == nil {
+		return nil
+	}
+	return stacktrace.Propagate(jsonError, "Error validating args, likely because it is not a valid JSON or YAML.")
```
leoporoli: it's propagating `jsonError` but what happens if `err is also. != nil` could be confusing for users if both are != nil because we will printing only the error for the first one.

leoporoli (approved):

cli/cli/commands/run/run.go
victorcolombo: Done!

PR #1362, opened 2023-09-21
fix: Manually locate docker socket

### LINKED ISSUE #1140: `unix:///var/run/docker.sock` isn't found
### What's your CLI version?

0.81.9

On some user machines (unknown what exactly causes this), there's no symlink at `/var/run/docker.sock`, pointing to the actual socket in the user's home directory (`~/.docker/run/docker.sock`). I'm not sure what causes this, and I've only seen it twice so far, but it MIGHT be related to installing Docker without `sudo`.

The bad news is, this is a confusing and hard blocker for users (especially those unfamiliar with Docker). This would have been a blocker for a user onboarding I was doing if I didn't know what the problem was (letting me guide the user to creating the symlink).

### Desired behavior

Regardless of whether `/var/run/docker.sock` exists, Kurtosis Just Works

gbouv:
## Description:
Kurtosis tries to manually locate the docker daemon socket before falling back to docker default behavior. In descending order of preference, it will:
1. Use the manually set environment variable `DOCKER_HOST` (if not empty)
2. Use the system wide docker socket `/var/run/docker.sock` if it exists
3. Use the user docker socket `${HOME}/.docker/run/docker.sock` if it exists
4. Fallback to `docker.FromEnv` which was what we were doing before

Fixes #1140 

YES


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
```diff
@@ -73,10 +78,35 @@ func getLocalDockerKurtosisBackend(
 	optionalApiContainerModeArgs *APIContainerModeArgs,
 ) (backend_interface.KurtosisBackend, error) {
 	dockerClientOpts := []client.Opt{
-		client.FromEnv,
 		client.WithAPIVersionNegotiation(),
 	}
 
+	// If the DOCKER_HOST env variable is set, use it. Otherwise, try to locate the daemon socket. Otherwise, fall back
+	// to env variables as we were doing before
+	userHomeDir, err := os.UserHomeDir()
```
h4ck3rk3y, line 86: i wonder if we can look t his up after `client.EnvOverrideHost` and `systemDaemonSocket` have failed to yield a `dockerClientOpts`; as we don't use this otherwise and the debug line could be mis-leading

but the current way is definitely more readable

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
```diff
@@ -73,10 +78,35 @@ func getLocalDockerKurtosisBackend(
...
+	} else if _, err := os.Stat(userOwnedSocketPath); err == nil {
+		logrus.Debugf("Connecting to Docker daemon via unix socket '%s'", userOwnedSocketPath)
+		fullyQualifiedUnixSocket := fmt.Sprintf("%s%s", unixSocketPrefix, userOwnedSocketPath)
+		dockerClientOpts = append(dockerClientOpts, client.WithHost(fullyQualifiedUnixSocket))
+	} else {
+		logrus.Debugf("Unable to locate Docker daemon socket and '%s' environment variable wasn't set. Falling "+
+			"back to Docker's own way to connect to locally running daemon. If it fails, make sure docker is running "+
+			"locally and try setting '%s' to help Kurtosis connect to it.", client.EnvOverrideHost,
+			client.EnvOverrideHost)
+		dockerClientOpts = append(dockerClientOpts, client.FromEnv)
```
h4ck3rk3y, line 107: Trusting this has been discussed somewhere, but the previous default way is now the final fallback, so people who might have the above configured will get some different behavior than before maybe?

I suspect this is an extreme edge case and isn't much to worry about

h4ck3rk3y (approved):
Thanks @gbouv ! Its been a pleasure :)

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
gbouv: We don't throw if it returns an error, so technically it doesn't affect the code flow, and I find it more readable this way.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
gbouv: I think it's the "new" expected behaviour yes, at least reading for the the associated tickets. Basically we is DOCKER_HOST is set, we want to use it in priority. Which I find also more comprehensive to users. If they made the effort of setting this env variable, it's probably for something and we should default to using it. I won't the here to roll back if this change causes issues, but I guess it's fairly small and can be reverted easily :)

PR #1364, opened 2023-09-21
fix: Relative import breaks for 'non-main branchs'

### LINKED ISSUE #1361: Relative import breaks for 'non-main branchs'
### What's your CLI version?

0.83.2

I modified the avalanche package to use relative locators, [here are the changes](https://github.com/kurtosis-tech/avalanche-package/compare/main...lporoli/local-absolute-locators), if we run this package locally it works but if we run the package using the “branch” it fails, it seems the relative import fails on this case. You can tes it by doing this `kurtosis run [github.com/kurtosis-tech/avalanche-package@lporoli/local-absolute-locators`

it’s the issue:
There was an error interpreting Starlark code 
```
Evaluation error: An error occurred while loading the module 'github.com/src/node_launcher.star'
        Caused by: The path 'github.com/src/node_launcher.star' needs to point to a specific file but it didn't. Users can only read or import specific files and not entire packages.
        at [github.com/kurtosis-tech/avalanche-package@lporoli/local-absolute-locators/main.star:1:30]: <toplevel>
        
Error encountered running Starlark code.
```

if you pay attention to this `github.com/src/node_launcher.star` it seems that the loader is removing the `/kurtosis-tech/avalanche-package` from the path.

### Desired behavior

kurtosis run `github.com/kurtosis-tech/avalanche-package@lporoli/local-absolute-locators` should work

leoporoli:
## Description:
Relative import breaks for 'non-main branchs'

NO

fix #1361


leoporoli:
@h4ck3rk3y don't forget this guy ;)

h4ck3rk3y:
@leoporoli it seems like the `build` is failing!!

h4ck3rk3y (approved):

PR #1368, opened 2023-09-25
feat: improved log experience on the UI.

Peeeekay:
This PR:
I have the logging experience now, where users can see large amount of logs. It behaves like terminal where if at bottom it automatically tracks the new logs, otherwise stays in place until user goes to bottom.

I have included copy and download functionality as well, where users can do that for service logs and logs generated during enclave creation.


enclave-manager/server/server.go
```diff
@@ -227,26 +227,23 @@ func (c *WebServer) RunStarlarkPackage(ctx context.Context, req *connect.Request
...
+			logrus.Errorf("Error occurred: %+v", errWhileSending)
+			return stacktrace.Propagate(errWhileSending, "Error occurred while sending streams")
 		}
 	}
+
+	if errWhileReceive := apiClient.Err(); errWhileReceive != nil {
+		logrus.Errorf("Error occurred: %+v", errWhileReceive)
+		return stacktrace.Propagate(errWhileReceive, "Error occurred while receiving data from engine")
+	}
+	ctx.Done()
```
tedim52: what's `ctx.Done()` doing here? From my understanding, [`Context.Done()`](https://pkg.go.dev/context#Context) just returns a channel.

engine/frontend/src/component/CreateEnclaveView.js
```diff
@@ -27,38 +29,47 @@ export const CreateEnclaveView = ({packageId, enclave, args}) => {
...
+    const readStreamData = (result) => {
+        if (result.case === "instruction") {
+            setLogs(logs => [...logs, result.value.executableInstruction])
+        }
 
-              if (result.case === "progressInfo" && result.value.currentStepInfo.length > 0) {
-                  console.log("progressinfo: ", result)
-                  setLogs(logs => [...logs, result.value.currentStepInfo[result.value.currentStepNumber]])
-              }
+        if (result.case === "progressInfo" && result.value.currentStepInfo.length > 0) {
```
adschwartz: @Peeeekay can we use constants for these cases?

engine/frontend/src/component/CodeEditor.js
```diff
@@ -87,6 +88,10 @@ export const CodeEditor = (
     // Start by manually setting the content of the editor. From hereafter user interaction will update it:
     useEffect(() => {
         handleEditorChange(value)
+        console.log("hello!")
+        return () => {
+            console.log("BYE BYE")
+        }
```
adschwartz: @Peeeekay this looks like test code?

engine/frontend/src/component/log/Log.js
```diff
@@ -0,0 +1,178 @@
...
+  const [bottomMap, setMaybeBottom] = useState([0, 0])
+  const [previousOffset, setPreviousOffset] = useState(0)
+
+
+  const setSize = useCallback((index, size) => {
+    sizeMap.current = { ...sizeMap.current, [index]: size };
+    listRef.current.resetAfterIndex(index);
+  }, []);
+
+  const getSize = index => sizeMap.current[index] || 50;
```
adschwartz, line 58: @Peeeekay could we make this a default param or a const somewhere?

Peeeekay: Like that 50? Because getSize is return the height of the Child component.

engine/frontend/src/component/log/useWindowResize.js
```diff
@@ -0,0 +1,18 @@
+import { useState, useLayoutEffect } from "react";
```
adschwartz: @Peeeekay this is almost identical to this file:
https://github.com/kurtosis-tech/kurtosis/blob/main/engine/frontend/src/utils/windowheight.js
Can we refactor it to become one file

enclave-manager/server/server.go
Peeeekay: oh lol; you are right; will remove this.

engine/frontend/src/component/CodeEditor.js
Peeeekay: true; let me clean this up.

engine/frontend/src/utils/download.js
```diff
@@ -0,0 +1,27 @@
+export const saveTextAsFile = (text, fileName) => {
+    const blob = new Blob([text], {type: "text/plain"});
+    const downloadLink = document.createElement("a");
```
adschwartz: can we make `"a"` a default parameter so it can be overwritten. This would be to avoid collisions with other elements :)

engine/frontend/src/component/log/Log.js
```diff
@@ -0,0 +1,178 @@
...
+            setSize={setSize}
+            windowWidth={windowWidth}
+            style={style}
+          />
+        </div>
+      )}
+      </List>
+      <Flex className="bg-black" style={{height: `80px`}}>
+        <Spacer />
+        <Box p='2' m="4" onClick={onCopy} className="rounded-md border-white border-1" height={"40px"}>
```
adschwartz: @Peeeekay the stuff inside `className` is for tailwind right? Can you move them out into the Box. Chakra has options for this too, including rounded borders as well.

engine/frontend/src/component/EnclaveInfo.js
```diff
@@ -93,7 +93,8 @@ const EnclaveInfo = ({enclaves}) => {
     }, [name, enclaves])
 
     const handleServiceClick = (service, services) => {
-        navigate(`/enclaves/${name}/services/${service.uuid}`, {state: {services, selected: service}})
+        console.log(service)
```
adschwartz: @Peeeekay delete?

engine/frontend/src/utils/download.js
```diff
@@ -0,0 +1,27 @@
+export const saveTextAsFile = (text, fileName) => {
```
adschwartz: @Peeeekay you'll need to remove this from the CodeEditor.js so we don't have two differing versions of this.

engine/frontend/src/component/log/CreateEnclaveLog.js
```diff
@@ -0,0 +1,114 @@
...
+        let stream;
+        const fetchLogs = async () => {
+            try {
+                stream = await runStarlark(enclave.host, enclave.port, packageId, args, appData.jwtToken, appData.apiHost);
+                for await (const res of stream) {
+                    const result = res["runResponseLine"]
+                    readStreamData(result)    
+                }
+            } catch (ex) {
+                console.log("Error occured while reading data from the enclave: ", enclave.name)
```
adschwartz: console.error?

engine/frontend/src/component/log/Log.js
```diff
@@ -0,0 +1,178 @@
+import React, { useCallback, useEffect, useRef, useState } from "react";
+import { VariableSizeList as List } from "react-window";
+import { useWindowResize } from "./useWindowResize";
```
adschwartz: @Peeeekay we should use the one in utils if at all possible

engine/frontend/src/component/log/Log.js
```diff
@@ -0,0 +1,178 @@
...
+  }, [setSize, index, windowWidth]);
+
+  if (data[index] !== undefined && data[index].length !== 0) {
+    let txt = data[index]
+    if (hasAnsi(data[index])) {
+      const parsedAnsi = convert.toHtml(data[index])
+      txt = parse(parsedAnsi)
+    }
+    return (
+        <div
```
adschwartz: @Peeeekay ideally we wouldn't put in these `div`s but use chakra components instead. It'll be more techdebt to clean up.

engine/frontend/src/component/log/Log.js
```diff
@@ -0,0 +1,178 @@
...
+          ref={rowRef}
+          style={styles.row}
+        >
+         {txt}
+        </div>
+      );
+  } 
+  return null;
```
adschwartz: return `<></>`?

engine/frontend/src/component/log/Log.js
```diff
@@ -0,0 +1,178 @@
...
+      return stripAnsi(log)
+    })
+
+    setCopyValue(logsWithoutAnsi.join("\n"))    
+    // automatically scroll the bottom if it;s at the bottom
+    if (logs.length > 0 && isBottom) {
+      scrollToBottom()
+    }
+
+  }, [logs]);
```
adschwartz, line 75: @Peeeekay does this mean that on every change to the logs we process the ansi stripping on all the logs? This sounds very labor intensive but let me know if I'm overlooking something. Is there anything we can do to avoid recomputing the same data over and over? Maybe storing with useMemo?

engine/frontend/src/component/log/Log.js
```diff
@@ -0,0 +1,178 @@
...
+  const isItAtBottom = () => {
+    console.log(logs.length - 1, bottomMap[0])
+    return bottomMap[0] === logs.length - 1
+  }
+
+  const handleDownload = () => {
+    const logsWithoutAnsi = logs.map((log)=> {
+      return stripAnsi(log)
+    })
+    const logsAsText = logsWithoutAnsi.join("\n")
```
adschwartz: @Peeeekay wondering if we can use a platform agnostic value?
https://stackoverflow.com/a/12989543

adschwartz (changes requested):
I'm excited to roll out this change, but there are some things we can do to avoid adding (non-needed) tech-debt. Happy to give it another pass when you have updates. Thanks!

engine/frontend/src/component/log/Log.js
Peeeekay: I am going for simplicity here; will stick to this. If there are issues in terms of performance, I can add more complex logic to only strip ansi from newly seen logs; but for now lets keep this simple. But good point! @adschwartz

engine/frontend/src/component/log/Log.js
Peeeekay: not sure why this is tech-debt? I am  not really using tailwind here; just `div` so we can keep this here for now. I don;t think we need to overkill this with chakra and react-window only accepts `inline-style`.. so additional apis provided by Chakra won't be useful.

enclave-manager/server/server.go
```diff
@@ -227,26 +227,23 @@ func (c *WebServer) RunStarlarkPackage(ctx context.Context, req *connect.Request
...
-				logrus.Errorf("error occurred: %+v", errWhileSending)
-				return stacktrace.Propagate(errWhileSending, "Error occurred while sending streams")
-			}
+	apiClient, err := (*apiContainerServiceClient).RunStarlarkPackage(ctx, serviceRequest)
+
+	for apiClient.Receive() {
+		resp := apiClient.Msg()
+		errWhileSending := str.Send(resp)
+		if errWhileSending != nil {
+			logrus.Errorf("Error occurred: %+v", errWhileSending)
```
tedim52: I think it's worth making the error messages here more descriptive so that it's easier for an engineer debugging to locate where it's coming from in our codebase. Maybe by adding some context about where the error surfaced (eg. `while streaming from enclave manager...`) Same for the `stacktrace` error here and below.

engine/frontend/src/api/enclave.js
```diff
@@ -42,14 +42,16 @@ export const createEnclave = async (token, apiHost, enclaveName, productionMode)
...
-        "enclaveIdentifier": enclaveName,
-        "serviceUuidSet": {
+        enclaveIdentifier: enclaveName,
+        serviceUuidSet: {
             [serviceUuid]: true
         },
-        followLogs: true,
+        followLogs: followLogs,
+        returnAllLogs: false,
+        numLogLines: numLogLines,
```
tedim52, line 54: nice

engine/frontend/src/component/CodeEditor.js
tedim52: you have the funniest log messages PK

engine/frontend/src/component/CreateEnclaveModal.js
```diff
@@ -11,9 +11,16 @@ export const CreateEnclaveModal = ({enclaveName, handleSubmit, name, setName, ar
...
     e.preventDefault();
     const fetch = async () => {
-      const enclave = await createEnclave(token, apiHost, enclaveName, productionMode);
-      addEnclave(enclave)
-      handleSubmit(enclave);
+      try {
+        const enclave = await createEnclave(token, apiHost, enclaveName, productionMode);
+        addEnclave(enclave)
+        handleSubmit(enclave);
+      } catch(ex) {
```
tedim52, line 18: `nit`: is `ex` supposed to stand for exception? used to seeing `err` here, but all the same

engine/frontend/src/component/CreateEnclaveView.js
```diff
@@ -27,38 +29,47 @@ export const CreateEnclaveView = ({packageId, enclave, args}) => {
...
+        let stream;
+        const fetchLogs = async () => {
+            try {
+                stream = await runStarlark(enclave.host, enclave.port, packageId, args, appData.jwtToken, appData.apiHost);
+                for await (const res of stream) {
+                    const result = res["runResponseLine"]
+                    readStreamData(result)    
+                }
+            } catch (ex) {
+                console.log("Error occured while reading data from the enclave: ", enclave.name)
```
tedim52: Is there a reason you're not logging `ex` here? Saw you were logging in it in an earlier try catch.

engine/frontend/src/component/CreateEnclaveView.js
```diff
@@ -27,38 +29,47 @@ export const CreateEnclaveView = ({packageId, enclave, args}) => {
...
+            } finally {
+                setLoading(false)
+            }
+        }
+        fetchLogs();
```
tedim52: question for my understanding: Given that this `fetchLogs` is in `CreateEnclaveView` and the stream processes `runResponseLine`, I'm guessing this is for fetching enclave logs and not service logs. Is that accurate?

engine/frontend/src/component/ServiceInfo.js
```diff
@@ -30,32 +30,43 @@ const ServiceInfo = () => {
...
     const {name: enclaveName, uuid: serviceUuid} = params;
 
-    useEffect(() => {
-        let stream;
-        const ctrl = new AbortController();
+    const startServiceLogCollector = (ctrl) => {
+        let stream
         const fetchLogs = async () => {
-            stream = await getServiceLogs(ctrl, enclaveName, serviceUuid, appData.apiHost);
+            stream = await getServiceLogs(ctrl, enclaveName, serviceUuid, appData.apiHost, true, 6000);
```
tedim52: maybe create constants for these? eg. `DEFAULT_SHOULD_FOLLOW_LOGS`, `DEFAULT_NUM_LOG_LINES`

engine/frontend/src/component/ServiceInfo.js
```diff
@@ -30,32 +30,43 @@ const ServiceInfo = () => {
...
             try {
                 for await (const res of stream) {
                     const log = res["serviceLogsByServiceUuid"][serviceUuid]["line"][0]
-                    setLogs(logs => [...logs, log])
+                    if (log !== "") {
```
tedim52, line 43: Nice check although backend shouldn't be sending empty log lines. If you're still seeing this happens a lot, let's chat via slack.

engine/frontend/src/component/ServiceInfo.js
```diff
@@ -30,32 +30,43 @@ const ServiceInfo = () => {
     const params = useParams()
     const {name: enclaveName, uuid: serviceUuid} = params;
 
-    useEffect(() => {
-        let stream;
-        const ctrl = new AbortController();
+    const startServiceLogCollector = (ctrl) => {
+        let stream
         const fetchLogs = async () => {
```
tedim52: very `nit`: might be worth calling this `fetchServiceLogs` to distinguish the `fetchLogs` from the enclave view

engine/frontend/src/component/log/Log.js
tedim52: maybe add this context in a comment here somewhere so it doesn't get forgotten?

tedim52 (approved):
nice work @Peeeekay . Jus a few comments

engine/frontend/src/component/log/Log.js
adschwartz: Yeah not noting use of tailwind, just not idiomatic use of Chakra. I would use Box, Container, etc. instead of div.

engine/frontend/src/component/CreateEnclaveModal.js
Peeeekay: yeah in javascript; it's in an exception :)

engine/frontend/src/component/CreateEnclaveView.js
Peeeekay: yeah logs while creating enclave and services inside it.

engine/frontend/src/component/CreateEnclaveView.js
Peeeekay: good catch!

engine/frontend/src/component/log/Log.js
Peeeekay: don't think i need it lel; just removed them.

engine/frontend/src/component/CreateEnclaveView.js
Peeeekay: not using this file; I am using log/CreateEnclaveLog.js and it has this change in it.

engine/frontend/src/component/ServiceInfo.js
```diff
@@ -30,32 +33,43 @@ const ServiceInfo = () => {
...
     }, [serviceUuid])
 
     const handleServiceClick = (service) => {
-        const fullPath = `/enclaves/${enclaveName}/services/${service.uuid}`
+        const fullPath = `/enclaves/${enclaveName}/services/${service.serviceUuid}`
         navigate(fullPath, {state: {services, selected: service}})
     }
 
+
+
```
adschwartz: if you are removing these lines, can you also remove the func below that is not in use (not your work)

engine/frontend/src/component/ServiceInfo.js
```diff
@@ -94,7 +108,7 @@ const ServiceInfo = () => {
             displayData = "Error while retrieving information"
         }
         return (
-            <Tr key={heading}>
+            <Tr key={heading} className="text-white">
```
adschwartz: use Chakra's `color`

engine/frontend/src/component/ServiceInfo.js
```diff
@@ -134,24 +166,35 @@ const ServiceInfo = () => {
...
-                                {tableRow("ENTRYPOINT", () => codeBox(1, "entrypoint", selectedSerialized.container.entrypointArgs))}
-                                {tableRow("CMD", () => codeBox(2, "cmd", selectedSerialized.container.cmdArgs))}
-                                {tableRow("ENV", () => codeBox(3, "env", selectedSerialized.container.envVars))}
-                            </Tbody>
-                        </Table>
-                    </TableContainer>
-
-                    {/*<LogView heading={`Service Logs`} logs={logs}/>*/}
+                <div className={`flex-col flex h-full space-y-1 bg-[#171923]`}>
+                    <Flex className={"bg-[#171923]"} style={{height: `80px`}}>
```
adschwartz: @Peeeekay use `height` and `bg` rather than className and style

engine/frontend/src/component/log/CreateEnclaveLog.js
```diff
@@ -0,0 +1,120 @@
...
+            <div className="flex h-full">
+                <LeftPanel 
+                    home={false} 
+                    heading={"Services"} 
+                    isServiceInfo={true}
+                    renderList={ ()=> renderServices(services, handleServiceClick)}
+                />
+                <div className="flex h-full w-[calc(100vw-39rem)] flex-col space-y-5">
+                    <div className='flex flex-col h-full space-y-1 bg-[#171923]'>
+                        <Flex className={"bg-[#171923]"} style={{height: `80px`}}>
```
adschwartz: Mixing here as well

engine/frontend/src/component/log/Log.js
```diff
@@ -0,0 +1,171 @@
...
+        ref={listRef}
+        height={(windowHeight-100)*0.8}
+        width={"100%"}
+        itemCount={logs.length}
+        itemSize={getSize}
+        itemData={logs}
+        onScroll={handleScroll}
+        outerRef={containerRef}
+        innerRef={testRef}
+        style={{backgroundColor: "black"}}
```
adschwartz, line 129: use chakra bg

engine/frontend/src/utils/download.js
```diff
@@ -0,0 +1,34 @@
+const DEFAULT_ELEMENT_NAME = "a"
+const DEFAULT_FILE_TYPE = "text/plain"
+
+export const saveTextAsFile = (text, fileName, options = {}) => {
```
adschwartz, line 4: @Peeeekay can't you just say: `elementName="box"`  rather than having this generic options hashmap with the associated code below? A default parameter is straight forward to do.

Peeeekay: the method will look like: saveTextAsFile(text, fileName, fileType="", elementName="") .. so no good way of distinguish between fileType and/or elementName. ( or I couldn;t find any).. this seems to be more cleaner.

adschwartz (commented):
I've made some additional comments. I would address those but removing my request for changes to unlock the PR

engine/frontend/src/component/CreateEnclaveView.js
adschwartz: @Peeeekay if we are not using this file shouldn't we delete it?

engine/frontend/src/component/ServiceInfo.js
Peeeekay: are you sure; we use that for `CodeEditor; it;s callback for data . @adschwartz

adschwartz: yeah you can just make an anonymous function, but not a big deal. Feel free to skip this.

engine/frontend/src/component/CreateEnclaveView.js
Peeeekay: yeah deleted the file

engine/frontend/src/component/log/Log.js
Peeeekay: this requires inline.

adschwartz (approved):

PR #1373, opened 2023-09-25
fix: handle missing arg types

adschwartz:
## Description:
Some packages will not have a type defined. This currently causes issues with the EM UI.

YES


engine/frontend/src/component/PackageCatalogForm.js
```diff
@@ -527,7 +535,11 @@ const PackageCatalogForm = ({createEnclave}) => {
             Object.keys(formData).map(key => {
                 const arg = kurtosisPackage.args[key]
                 const argName = getArgName(arg)
-                const type = getType(arg)
+                let type = ""
+                try {
```
tedim52, line 539: jus curious what's the reasoning behind wrapping this in a `try catch` if you don't handle the error/do anything on the catch?

tedim52 (approved):
lgtm, jus one question for my understanding

engine/frontend/src/component/PackageCatalogForm.js
adschwartz: The function will throw an exception and I don't really care what the error is as it will be due to data missing. So in the absence of the data I fill it in with a default that wont break the UI.

PR #1397, opened 2023-09-26
fix: bring back args

adschwartz:
## Description:
We should only ignore "plan" arguments

YES


Peeeekay:
@adschwartz could you build the frontend artifacts; will approve after that.

Peeeekay (approved):

PR #1407, opened 2023-09-27
fix: text off center

### LINKED ISSUE #1406: No file artifacts message is not aligned
### What's your CLI version?

0.83.9

the `There are no file artifacts in this enclave` is misaligned, it could be more up

image

### Desired behavior

Align the message

adschwartz:
## Description:
Text was off center. This is no longer the case. 

YES

Closes #1406


tedim52 (approved):
ahhh frontend. good work Anders

PR #1421, opened 2023-09-28
fix: stale data in code editor

### LINKED ISSUE #1393: The entrypoint field in the 'detailed service info' form is not refreshing the value
### What's your CLI version?

0.83.8

I don't see the `entrypoint` field getting refreshed meanwhile a change between the services, I can see changes in other fields like 'UUID' and 'name'

image

image

### Desired behavior

should refresh the value for the selected service

adschwartz:
## Description:
The monaco editor was not configured to use proper ids to manage different models being present at the same time. This is fixed with this change.

YES

Closes #1393


tedim52 (approved):
LGTM

PR #1427, opened 2023-09-28
fix: relative locators for read_file and upload_files instructions

### LINKED ISSUE #1412: bug: relative locators fails for quickstart package
### What's your CLI version?

0.83.7

There is another bug related to relative imports this time happens with the  quickstart packagein theawesome repository and it seems it happens when a package uses relative reference to a sibling package, data-packgein this case.
[If you check this change](https://github.com/kurtosis-tech/awesome-kurtosis/pull/187/files#diff-abf5ef9d44740e5825e0f95de966f699d74df8f7948ef4183ff1a821a57947c7R7) you will see  the data-package/main.star file is uploading a file inside the same repo using a relative path, and then [if you check here](https://github.com/kurtosis-tech/awesome-kurtosis/blob/main/quickstart/main.star#L1)  you will see that quickstart package is importing this module from its sibling, so when you run `kurtosis run [github.com/kurtosis-tech/awesome-kurtosis/quickstart](http://github.com/kurtosis-tech/awesome-kurtosis/quickstart) it fails (I reverted the change so it won’t fails now) you can also [see the error in the mono-repo CI test](https://app.circleci.com/pipelines/github/kurtosis-tech/kurtosis/7072/workflows/ba012562-5cb2-480e-b49b-806592107a01/jobs/95120) because it’s using it.

It seems the relative path works when quickstart is executed in awesome CI (because I think it’s doing kurtosis run .) but the relative breaks when it’s executed from the mono repo, because, It does `kurtosis run [github.com/kurtosis-tech/awesome-kurtosis/quickstart](http://github.com/kurtosis-tech/awesome-kurtosis/quickstart)

We created an example to reproduce the issue:

`kurtosis run github.com/h4ck3rk3y/package-panic/primary`   will break meanwhile 

`kurtosis run [github.com/h4ck3rk3y/package-panic/data`  will works.

### Desired behavior

Quickstar package can use relative locators

leoporoli:
## Description:
relative locators for `read_file` and `upload_files` instructions

NO
Because I think users are not using relative locators yet

Fix #1412


h4ck3rk3y (approved):
Nice! I wonder if we can get rid of the `thread.Name = locatorOfPackage` dependency all together from the `startosis_interpreter`; i mean does the Filename logic work for import_module as well?

leoporoli:
> Nice! I wonder if we can get rid of the `thread.Name = locatorOfPackage` dependency all together from the `startosis_interpreter`; i mean does the Filename logic work for import_module as well?

@h4ck3rk3y yes import_module is also using the new logic. Is this enough? or Do you want to rename the threads? 

I can remove the comment mentioning the old logic related to the thread name, I'll do it

h4ck3rk3y (approved):
@leoporoli I think with that comment removed this is in a good state :)

h4ck3rk3y (approved):

PR #1437, opened 2023-09-29
ci: reindex after docs are publsihed

h4ck3rk3y:
## Description:
We used to re-index in parallel to docs publish, that can be buggy.

YES - docs will work better


tedim52 (approved):
thanks for fixing!!!

PR #1446, opened 2023-09-29
feat!: block 'local absolute locators', users should replace' local absolute locators' with 'relative locators'

### LINKED ISSUE #1435: Block absolute imports for local imports
### Background & motivation

Right now we allow absolute imports for specifying local imports (imports within a package). This is a legacy from the days before we had relative imports, and it’s causing some of the painpoints above.

We should instead ban absolute imports within a package, and force relative imports (either via a leading `.` or from the package root using `/` )

from: https://www.notion.so/kurtosistech/2023-09-Forked-Packages-Cleanup-d8bc4f5165ae4b039c07cdad56892459

### Desired behavior

This has three benefits:

- Users don’t have the option to get confused about which syntax they should use
- Eliminate the “accidentally pull from Github when I didn’t want to” issue
- Make it clearer at a glance which imports are imported from outside the package, and which imports are imported from inside

leoporoli:
## Description:
block 'local absolute locators', users should replace' local absolute locators' with 'relative locators'

YES

This should be merged after https://github.com/kurtosis-tech/kurtosis/pull/1427
Fix #1435


core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/test_engine/static_constants.go
```diff
@@ -21,7 +21,10 @@ var (
 
 	TestSrcPath = "/path/to/file.txt"
 
-	TestModuleFileName = "github.com/kurtosistech/test-package/helpers.star"
+	TestModulePackageId       = "github.com/kurtosistech/test-package"
```
h4ck3rk3y, line 24: can these constants be private?

h4ck3rk3y (approved):

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/test_engine/static_constants.go
h4ck3rk3y: happy for you to tackle it separately as well; this PR is green and Good to Go

PR #1455, opened 2023-10-02
feat:  Get Starlark Run APIC endpoint

laurentluce:
## Description:
We need a way to retrieve the latest Starlark run info: package id, serialized params...  The same parameters that are passed to the RunStarlark* APIC methods.

YES

https://www.notion.so/kurtosistech/Run-Config-07b2dbdab74d45d59be76061500ee435


core/server/api_container/server/api_container_service.go
```diff
@@ -78,19 +78,26 @@ type ApiContainerService struct {
 	startosisRunner *startosis_engine.StartosisRunner
 
 	startosisModuleContentProvider startosis_packages.PackageContentProvider
+
+	isProduction bool
```
adschwartz: @laurentluce this is not really part of your work, but I wonder if we could/should? take this opportunity to rename this guy: the naming of this flag is not ideal imo. IMO we are conflating the use of a field with the name of a field. This flag is supposed to indicate if a service should be restarted if it dies. We can use that feature in any environment/use caes (whether it's development, test or production envs).

h4ck3rk3y:
Really excited about this PR! wish we can see Args i nCLI too

adschwartz (approved):

core/server/api_container/server/api_container_service.go
laurentluce: Renamed to `restartPolicy`.  More details [here](https://kurtosistech.slack.com/archives/C05T1BG3G6S/p1696279260415829?thread_ts=1696258247.950339&cid=C05T1BG3G6S).  Another PR will be created to rename it in `CreateEnclave` since it is a breaking change.

PR #1465, opened 2023-10-04
fix: bug where we passed cloud user id for cloud instance id

h4ck3rk3y:
## Description:
Metrics would show cloud user id instead of cloud instance id as we passed the wrong value; this fixes it

NO


adschwartz (approved):
Oh good find @h4ck3rk3y

PR #1471, opened 2023-10-04
fix: autoscroll

adschwartz:
## Description:
The autoscroll behvior was sticking even though the user was scrolling up and trying to escape it. With this change we are simplifying the scroll logic and it seems to be performing better.

YES


leoporoli (approved):

tedim52 (approved):
LGTM

Issue #1459, opened 2023-10-03
Data creep


barnabasbusa:
### What's your CLI version?

0.83.16

### Description & steps to reproduce

Each run increases the Docker.raw file size, which will lead to the virtual disk exhaustion eventually.

Use the following run example:
```json
{
  "participants": [
    {
      "el_client_type": "geth",
      "cl_client_type": "teku",
      "count": 3
    },
    {
      "el_client_type": "nethermind",
      "cl_client_type": "teku",
      "el_client_image": "nethermindeth/nethermind:cancun"
    },
    {
      "el_client_type": "besu",
      "cl_client_type": "teku"
    }
  ],
  "network_params": {
    "deneb_fork_epoch": 2,
    "seconds_per_slot": 3
  },
  "global_client_log_level": "info",
  "launch_additional_services": true,
  "mev_type": "full",
  "mev_params": {
    "mev_relay_image": "flashbots/mev-boost-relay:0.28.0a1",
    "mev_builder_image": "flashbots/builder:1.13.2.dev1"
  }
}
```

Steps to reproduce on MacOS:
* Install docker desktop
* `cd ~/Library/Containers/com.docker.docker/Data/vms/0/data`
* Delete Docker.raw (if you have one already), or mv to backup. 
* Start docker desktop (it will initiate a new Docker.raw file)
* `du -sh *` will now display that Docker.raw file is approx 1.6G
* `kurtosis engine start`
* `du -sh *` will now display that Docker.raw file is approx 1.9G
* `kurtosis run github.com/kurtosis-tech/ethereum-package --args-file ~/example.json`
* `du -sh *` will now display that Docker.raw file is approx 9.4GB.
* `kurtosis clean -a`
* `du -sh *` will now display that Docker.raw file is approx 9.3GB.
* `kurtosis run github.com/kurtosis-tech/ethereum-package --args-file ~/example.json`
* `du -sh *` will now display that Docker.raw file is approx 10GB.
* `kurtosis clean -a`
* `du -sh *` will now display that Docker.raw file is approx 9.7GB.
*  This leads to a creep of about 200-500MB each time you run a big test like this. It seems like kurtosis clean -a doesn't clean up all datadir and leaves some stuff just hanging around (not talking about docker images). 

[...]

Issue #1394, opened 2023-09-26
Kurtosis Logs not being cleaned


cbermudez97:
### What's your CLI version?

0.83.5

### Description & steps to reproduce

When running a package multiple times, and later removing the enclave( or running `kurtosis clean -a`) looks like the contents of the `kurtosis-logs-storage` volume related to previous runs are not correctly cleaned. This can lead to these volumes consuming high amounts of disk space especially when running with a high number of services(ie. https://github.com/kurtosis-tech/eth2-package).

Steps to reproduce:
1. Run `kurtosis run github.com/kurtosis-tech/eth2-package`.
2. Wait for the execution to finish.
3. Delete enclave through either `kurtosis enclave [stop|rm] <name>` or `kurtosis clean -a`
4. Restart from step 1. multiple times

<img width="332" alt="Screenshot 2023-09-26 at 12 41 34 PM" src="https://github.com/kurtosis-tech/kurtosis/assets/43155355/1f6de74a-42dc-4d28-ba8e-02c194e2fc5d">
<img width="637" alt="image" src="https://github.com/kurtosis-tech/kurtosis/assets/43155355/82810499-f5aa-4dbe-93ae-ccd62139e537">

### Desired behavior

Removing an enclave with either `kurtosis enclave rm <name>` or `kurtosis clean -a` should also remove logs stored on the `kurtosis-logs-storage` volume. Otherwise, a flag or command should be able to clean it more easily.

### What is the severity of this bug?

Papercut; this bug is frustrating, but I have a workaround.

tedim52:
Hey @cbermudez97. Thanks for creating this issue! Assigning this to myself and looking into it. I'll get back to you once I've got smth : )

tedim52:
Hey @cbermudez97 just merged this fix along with some other things that will reduce logs storage. This should be out in the next kurtosis release so I'll ping you then so you can upgrade!

tedim52:
Hey, @cbermudez97 the fix is released in [`v0.84.3`](https://github.com/kurtosis-tech/kurtosis/releases/tag/0.84.3) if you want to update. Will note that this only applies to future logs created - so if you want to free up current logs, I recommend manually removing the `kurtosis-logs-storage` Docker volume to get rid of old logs and start fresh. (if you don't need them of course)

PR #1472, opened 2023-10-05
fix: replace duplicate log files with symlinks

tedim52:
## Description:
Due to this [PR](https://github.com/kurtosis-tech/kurtosis/pull/1213), the logs aggregator stores duplicate log files which causes log storage to grow quickly. This change prevents storing duplicate logs by periodically replacing the duplicate log files with symlinks to a single log file. This allows logs to still be retrievable from stopped enclaves using any service identifier without increasing log storage.

Also includes a small refactor to use a better supported pkg (`spf13/afero`) to mock the filesystem that cleans up logging unit tests quite a bit.

NO

## References:
https://github.com/kurtosis-tech/kurtosis/pull/1213
https://github.com/kurtosis-tech/kurtosis/issues/1394
https://github.com/kurtosis-tech/kurtosis/issues/1459


engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_creator/log_file_creator.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	// replace with symlink
+	if err := creator.filesystem.Symlink(targetLogFilePath, symlinkLogFilePath); err != nil {
+		return stacktrace.Propagate(err, "An error occurred creating a symlink file path '%v' for target file path '%v'.", targetLogFilePath, targetLogFilePath)
+	}
+	return nil
+}
+
+// creates a filepath of format /<filepath_base>/year/week/<enclave>/serviceIdentifier.<filetype>
+func getFilepathStr(year, week int, enclaveUuid, serviceIdentifier string) string {
```
adschwartz, line 140: @tedim52 I think we should add a unit test for this guy alone to verify the format

engine/server/engine/centralized_logs/client_implementations/persistent_volume/persistent_volume_logs_database_client_test.go
```diff
@@ -256,12 +255,12 @@ func TestStreamUserServiceLogs_ThousandsOfLogLinesSuccessfulExecution(t *testing
...
 
-	file1 := fmt.Sprintf(volume_consts.PerFileFmtStr, volume_consts.LogsStorageDirpathForTests, string(enclaveUuid), testUserService1Uuid, volume_consts.Filetype)
-	underlyingFs := &fstest.MapFS{
-		file1: {
-			Data: []byte(logLinesStr),
-		},
-	}
+	underlyingFs := volume_filesystem.NewMockedVolumeFilesystem()
+
+	file1PathStr := fmt.Sprintf(volume_consts.PerFileFmtStr, volume_consts.LogsStorageDirpath, string(enclaveUuid), testUserService1Uuid, volume_consts.Filetype)
```
adschwartz, line 260: @tedim52 I don't think this is wrong per se, but a way to add another level of checking would be to write a literal string version of this code so that you know that the "coded" version of this (in `getFilepathStr`) is producing the string that you expect)

adschwartz (approved):

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_creator/log_file_creator.go
tedim52: completely agree, if you don't mind I've added some unit tests in my logs `enclave rm` and `clean -a` that address this so gonna resolve this for now

Issue #1457, opened 2023-10-02
deleting more images than necessary


h4ck3rk3y:
### What's your CLI version?

0.83.16

### Description & steps to reproduce

kurtosis clean -a 

<img width="1576" alt="image" src="https://github.com/kurtosis-tech/kurtosis/assets/6670312/85a0005e-4c13-49e1-8dd7-84eba9288da8">

### Desired behavior

This deletes 0.83.16 core/expander while it should keep that

### What is the severity of this bug?

Papercut; this bug is frustrating, but I have a workaround.

### What area of the product does this pertain to?

CLI: the Command Line Interface

victorcolombo:
Currently the image deleting algorithm is to delete all non-dev images that are not being used. Did you have an APIC running? If not, it makes sense that it was deleted

victorcolombo:
This was a ease of implementation compromise, without having to deal with parsing semver, only customers that upgraded Kurtosis, ran an APIC, stopped the APIC, than ran clean -a would be affected, which seems a corner case

h4ck3rk3y:
Yep, I didn't have an APIC running. I thought the algorithm was to delete previous versions, wasn't aware of this choice. My internet is alright so it doesn't affect me as much.

barnabasbusa:
`kurtosis clean -a` now deletes all images that I'm using, and every time I rerun kurtosis run, it has to download all images all over, this issue is super frustrating when you have slow internet connection:

kurtosis version 0.83.16

Container images used in this run:
> postgres:alpine - remotely downloaded
> prom/prometheus:latest - remotely downloaded
> hyperledger/besu:develop - remotely downloaded
> redis:alpine - remotely downloaded
> grafana/grafana-enterprise:9.2.3 - remotely downloaded
> flashbots/mev-boost - remotely downloaded
> ethpandaops/dora-the-explorer:master - remotely downloaded
> ethpandaops/tx-fuzz:master - remotely downloaded
> ethpandaops/beacon-metrics-gazer:master - remotely downloaded
> ethpandaops/ethereum-genesis-generator:2.0.0-rc.6 - remotely downloaded
> ethereum/client-go:latest - remotely downloaded
> ethpandaops/execution-monitor:master - remotely downloaded
> nethermindeth/nethermind:cancun - remotely downloaded
> sigp/lighthouse - remotely downloaded
> flashbots/mev-flood - remotely downloaded
> python:3.11-alpine - remotely downloaded
[...]

PR #1473, opened 2023-10-05
feat: Always keep latest released version of Kurtosis images

victorcolombo:
## Description:

Before this PR, images were deleted based on if they were being used by any container and were not dev versions. This caused the latest released version to be deleted if there were no enclaves active.

After this PR, the algorithm is:
- Keep all dev versions
- Keep the latest released version
- Keep all images that are used by at least one container

YES

https://github.com/kurtosis-tech/kurtosis/issues/1457


mieubrisse:
When you say, "keep latest released version" does that mean "keep the version of the CLI being currently run?" As written, it sounds like if the latest is 1.2.3. and I downgrade my CLI to 1.2.0, then the CLI will continually be deleting the images of 1.2.0

I'd been under the impression that the algorithm we'd discussed a while back was:

- Keep all dev versions (so you're not getting dev versions yanked out from under you)
- Keep all versions >= the version of the CLI doing the `clean` (so you're not deleting future versions, in case you've downgraded)
- Keep all images that are used by at least one container (so that we're not erroring)

This is the algorithm that makes the most sense to me as a user, though I could be convinced that point #2 should be "keep all versions == the version of the CLI doing the `clean`" so that if you downgrade, you're not left with future versions (you may never intend to reupgrade).

h4ck3rk3y:
Assigning the review to @mieubrisse as he's already on it!

victorcolombo:
That makes sense @mieubrisse, bullets 1 and 3 are already covered, now there are two viable options for 2:

1. Keep only images where version == CLI (This is the current behavior on this PR, and the one I find most intuitive). The current use case for `clean -a` for users is to minimize Kurtosis footprint to a minimum, keeping only what is absolutely necessary to run. In this case, the intuitive behavior for a user in my opinion is that, if you pinned a version by downgrading it on your cluster, you should only keep the version you actually need to run.
2. Keep images >= CLI (Behavior you are suggesting). This case is more aligned with what a Kurtosis dev would face (downgrading and upgrading Kurtosis to bisect a bug), and I agree that in this case it makes more sense to keep these images around.

Let me know what is your preference @mieubrisse

mieubrisse:
Honestly I'm okay with the "keep ONLY the CLI version's images" case. This keeps Kurtosis as lean as possible for the user, and Kurtosians can just deal with re-pulling if they're bisecting a bug.

mieubrisse (approved):

PR #1478, opened 2023-10-05
feat: Edit Enclave

adschwartz:
## Description:
With this change the user can click to edit a running enclave, change the configuration of the enclave (package), click Run and the enclave will update idempotently (using idempotent runs).

YES


enclave-manager/api/protobuf/kurtosis_enclave_manager_api.proto
```diff
@@ -55,3 +56,8 @@ message InspectFilesArtifactContentsRequest {
   int32 apic_port = 2;
   api_container_api.FilesArtifactNameAndUuid file_names_and_uuid = 3;
 }
+
+message GetStarlarkRunRequest{
+  string apic_ip_address = 1;
+  int32 apic_port = 2;
```
tedim52, line 62: why not a `uint32` since ports are nonnegative?

adschwartz: that's a good point! It should have been. We'll need to change this in the other requests too.

tedim52 (approved):
LGTM - sick feature

PR #1489, opened 2023-10-06
feat: remove logs on enclave rm and clean -a

tedim52:
## Description:
Addresses https://github.com/kurtosis-tech/kurtosis/issues/1394 and https://github.com/kurtosis-tech/kurtosis/issues/1459 by removing an enclaves logs on `kurtosis enclave rm ...` and removing all logs on `kurtosis clean -a`

YES

## References:
https://github.com/kurtosis-tech/kurtosis/issues/1394
https://github.com/kurtosis-tech/kurtosis/issues/1459


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -473,6 +473,18 @@ func (backend *DockerKurtosisBackend) DestroyLogsCollectorForEnclave(ctx context
 	return nil
 }
 
+func (backend *DockerKurtosisBackend) DestroyLogsStorage(ctx context.Context) error {
+	logsStorageAttrs, err := backend.objAttrsProvider.ForLogsStorageVolume()
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred attempting to destroy logs storage.")
```
leoporoli: nit: same error msg from both calls, you could be more specific here.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -111,20 +107,6 @@ func CreateEngine(
 			removeLogsAggregatorFunc()
 		}
 	}()
-
-	// schedule log removal for log retention
-	go func() {
```
leoporoli, line 116: is this code remotion fine? just to verify that it's not an accidental remotion because I didn't see something in the description

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
tedim52: yep good callout.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
tedim52: yes it is, I moved this logic into the `engine` (which is actually where it should be)

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
```diff
@@ -140,3 +171,12 @@ func (creator *LogFileCreator) createSymlinkLogFile(targetLogFilePath, symlinkLo
 func getFilepathStr(year, week int, enclaveUuid, serviceIdentifier string) string {
 	return fmt.Sprintf(volume_consts.PerWeekFilePathFmtStr, volume_consts.LogsStorageDirpath, strconv.Itoa(year), strconv.Itoa(week), enclaveUuid, serviceIdentifier, volume_consts.Filetype)
 }
+
+func getEnclaveLogsDirPath(year, week int, enclaveUuid string) string {
+	logsDirPathForYearAndWeek := fmt.Sprintf(volume_consts.PerWeekDirPathStr, volume_consts.LogsStorageDirpath, strconv.Itoa(year), strconv.Itoa(week))
```
leoporoli: you could call ` getLogsDirPathForWeek` and replace this line

engine/server/engine/enclave_manager/enclave_manager.go
```diff
@@ -253,6 +257,9 @@ func (manager *EnclaveManager) DestroyEnclave(ctx context.Context, enclaveIdenti
 		return stacktrace.Propagate(err, "An error occurred destroying the enclave")
 	}
 	if _, found := successfullyDestroyedEnclaves[enclaveUuid]; found {
+		if err = manager.enclaveLogFileManager.RemoveEnclaveLogs(enclaveIdentifier); err != nil {
```
leoporoli: I think that will be better if you use `enclaveUuid` instead here

engine/server/engine/main.go
```diff
@@ -136,30 +136,40 @@ func runMain() error {
...
 	perFileStreamStrategy := stream_logs_strategy.NewPerFileStreamLogsStrategy()
 	perFileLogsDatabaseClient := persistent_volume.NewPersistentVolumeLogsDatabaseClient(kurtosisBackend, osFs, perFileStreamStrategy)
 
 	// pulls logs /per week/per enclave/per service
-	realTime := logs_clock.NewRealClock()
 	perWeekStreamStrategy := stream_logs_strategy.NewPerWeekStreamLogsStrategy(realTime)
 	perWeekLogsDatabaseClient := persistent_volume.NewPersistentVolumeLogsDatabaseClient(kurtosisBackend, osFs, perWeekStreamStrategy)
 
+	logFileManager := log_file_manager.NewLogFileManager(kurtosisBackend, osFs, realTime)
+	go func() {
```
leoporoli: I think that will be worth it to move all this threads and logic inside the logFileManager and trigger it from here, just to encapsulate the logic in the right place, but you can do it in a next PR if this one needs to get merged now

tedim52: ooo yes this is a good idea ill do this

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -473,6 +473,18 @@ func (backend *DockerKurtosisBackend) DestroyLogsCollectorForEnclave(ctx context
 	return nil
 }
 
+func (backend *DockerKurtosisBackend) DestroyLogsStorage(ctx context.Context) error {
```
leoporoli: I don't see any call to this method.  I could be wrong. is it used?

leoporoli (approved):

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
tedim52: I was initially going to use this `DestroyLogsStorage` to remove logs during a `kurtosis clean -a` from the CLI but realized that wasn't possible since the volume can't be removed while the running engine container is still using it - so I opted to use `RemoveEnclaveLogs` from the engine to remove logs during `clean -a` instead. I'll remove the function on the principle of dead code is dangerous.

Issue #1299, opened 2023-09-14
Allow offline mode with local images


barnabasbusa:
### Background & motivation

I know I have requested a feature to always pull images, however, it would be great if there would be a mode where this would not be enabled by default.

### Desired behaviour

It would be great if there would be a flag e.g: `--offline` which disabled all image pulls and only works with local images.

Basically to allow kurtosis to operate in an offline only mode. (Very useful when your internet connection is unreliable, and you know that you already have the images locally available). 

### How important is this to you?

Painful; the lack of this feature makes using Kurtosis frictionful.

mieubrisse:
Hey @barnabasbusa , I'm keen to grab a chat with you to see explicitly what were the symptoms you experienced. I think there's the opportunity to put some smart logic in the product

mieubrisse:
Related to, but not superseded by, #1307

h4ck3rk3y:
hey @mieubrisse I have assigned this to you as Tedi told me there's some design work going on on this;

leeederek:
Hey @barnabasbusa, we've merged #1495 and #1660! Kurtosis can now be used while offline so long as your images are locally cached. By default, Kurtosis will not attempt to fetch new image tags if an image already exists locally. 

Can we close out this issue?

PR #1495, opened 2023-10-06
feat: Add cli argument to control image download

lostbean:
## Description:
This PR adds a new cli argument to control how to download the latest images.

YES

This PR is aimed to address #1299


api/golang/core/lib/binding_constructors/binding_constructors.go
```diff
@@ -41,11 +41,14 @@ func NewRunStarlarkScriptArgs(
...
 	cloudInstanceId string,
 	cloudUserId string,
+	imageDownloadMode string,
 ) *kurtosis_core_rpc_api_bindings.RunStarlarkScriptArgs {
 	cloudInstanceIdCopy := new(string)
 	*cloudInstanceIdCopy = cloudInstanceId
 	cloudUserIdCopy := new(string)
 	*cloudUserIdCopy = cloudUserId
+	imageDownloadModeCopy := new(string)
+	*imageDownloadModeCopy = cloudUserId
```
tedim52: `s/cloudUserId/imageDownloadMode`

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -1244,6 +1246,26 @@ func (manager *DockerManager) FetchLatestImage(ctx context.Context, dockerImage
...
+func (manager *DockerManager) FetchImage(ctx context.Context, image string, download_mode image_download_mode.ImageDownloadMode) (bool, error) {
+	var err error
+	var pulledFromRemote bool = true
+
+	switch image_pulling := download_mode; image_pulling {
+	case image_download_mode.Always:
+		err = manager.FetchLatestImage(ctx, image)
+	case image_download_mode.Missing:
+		pulledFromRemote, err = manager.FetchImageMissing(ctx, image)
+	case image_download_mode.Never:
```
tedim52: I'm assuming this is where the `FetchLocalOnly` implementation would go?

core/server/api_container/server/startosis_engine/startosis_validator/docker_images_validator.go
```diff
@@ -61,7 +63,7 @@ func fetchImageFromBackend(ctx context.Context, wg *sync.WaitGroup, imageCurrent
 	}()
 
 	logrus.Debugf("Starting the download of image: '%s'", imageName)
-	imagePulledFromRemote, err := (*backend).FetchImage(ctx, imageName)
+	imagePulledFromRemote, err := (*backend).FetchImage(ctx, imageName, image_download_mode.Missing)
```
tedim52: shouldn't this be be the `imageDownloadMode` that's set in the `startosisValidator`?

leeederek:
Hey @lostbean and @tedim52:
1. Can we make sure we have docs for this new flag in the `run` starlark reference?
2. I saw in a doc that there was an ownership transfer that took place - can the correct owner be reassigned on the parent issue? https://github.com/kurtosis-tech/kurtosis/issues/1299

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
lostbean: Yes, but we just decided to remove the `never` mode for now

core/server/api_container/server/startosis_engine/startosis_validator/docker_images_validator.go
lostbean: yes!

lostbean:
> Hey @lostbean and @tedim52:
> 
> 1. Can we make sure we have docs for this new flag in the `run` starlark reference?
> 2. I saw in a doc that there was an ownership transfer that took place - can the correct owner be reassigned on the parent issue? [Allow offline mode with local images #1299](https://github.com/kurtosis-tech/kurtosis/issues/1299)

@leeederek the issue is already assigned to me and I just updated the documentation to include the new flag.

cli/cli/commands/run/run.go
```diff
@@ -608,6 +624,23 @@ func parseVerbosityFlag(flags *flags.ParsedFlags) (command_args_run.Verbosity, e
...
+
+	defaultImageMode := kurtosis_core_rpc_api_bindings.ImageDownloadMode_missing
+
+	imageDownloadStr, err := flags.GetString(imageDownloadFlagKey)
+	if err != nil {
+		return defaultImageMode, stacktrace.Propagate(err, "An error occurred getting the image-download using flag key '%s'", imageDownloadFlagKey)
+	}
+	imageDownload, err := command_args_run.ImageDownloadString(imageDownloadStr)
+	if err != nil {
+		return defaultImageMode, stacktrace.Propagate(err, "Invalid image-download value: '%s'. Possible values are %s", imageDownloadStr, strings.Join(command_args_run.ImageDownloadStrings(), ", "))
```
tedim52: I see that if this returns, `err` you return the `defaultImageMode` with an `err` but then above you check for the error, then propagate the error, so the `defaultImageMode` actually never ends up being used afaict.

tedim52: If I'm right, I suggest removing the `defaultImageMode`, and then return `nil, stacktrace.Prop...` in case of an error, and then have the default image download value be configured as the default in `StarlarkRunConfig`, and be overrided by `starlark_run_config.WithImageDownloadMode(imageDownload)`. This way we only have to reason about the default image download mode in one place `StarlarkRunConfig` instead of two.

If you see a good reason for the default being returned here, I'd just suggest documenting the behavior of the default being returned in function spec comment.

container-engine-lib/lib/backend_impls/metrics_reporting/metrics_reporting_kurtosis_backend.go
```diff
@@ -2,18 +2,20 @@ package metrics_reporting
 
 import (
 	"context"
+	"io"
```
tedim52, line 5: hmm where are `io` and `time` being used/are they needed imports with this change?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/kubernetes_kurtosis_backend.go
```diff
@@ -2,6 +2,8 @@ package kubernetes_kurtosis_backend
 
 import (
 	"context"
+	"io"
```
tedim52, line 5: is `io` being used?

container-engine-lib/lib/backend_impls/docker/docker_manager/create_and_start_container_args.go
```diff
@@ -249,3 +255,13 @@ func (builder *CreateAndStartContainerArgsBuilder) WithContainerInitEnabled(cont
 	builder.containerInitEnabled = containerInitEnabled
 	return builder
 }
+
+func (builder *CreateAndStartContainerArgsBuilder) WithPullingLatestImageAlways() *CreateAndStartContainerArgsBuilder {
```
tedim52: I like that you separated the two functions in two as opposed to having the `image download` mode passed in! 

also very nit smth I care about having consistent naming of this behavior across the repo - I think `Pulling` makes sense here but now we're using `download` in CLI, `Fetch` in `container-engine-lib` and `Pulling` here to describe the same behavior. I wonder if there's a way to simplify this 

no ask here jus wanted to point out

container-engine-lib/lib/backend_impls/docker/docker_manager/create_and_start_container_args.go
```diff
@@ -249,3 +255,13 @@ func (builder *CreateAndStartContainerArgsBuilder) WithContainerInitEnabled(cont
...
+	builder.imageDownloadMode = image_download_mode.Always
+	return builder
+}
+
+func (builder *CreateAndStartContainerArgsBuilder) WithPullingLatestImageOnMissing() *CreateAndStartContainerArgsBuilder {
```
tedim52: could we change this to `WithPullingLatestImageIfMissing` , I think `s/On/If` will help express the behavior is conditional.

container-engine-lib/lib/backend_impls/docker/docker_manager/create_and_start_container_args.go
tedim52: maybe I'd consider changing `Pulling` to `Fetch` so we can keep it to `download` and `fetch` but up to you

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -1180,10 +1182,10 @@ func (manager *DockerManager) GetContainersByLabels(ctx context.Context, labels
 	return result, nil
 }
 
-// [FetchImage] uses the local [dockerImage] if it's available.
+// [FetchImageMissing] uses the local [dockerImage] if it's available.
 // If unavailable, will attempt to fetch the latest image.
 // Returns error if local [dockerImage] is unavailable and pulling image fails.
-func (manager *DockerManager) FetchImage(ctx context.Context, dockerImage string) (bool, error) {
+func (manager *DockerManager) FetchImageMissing(ctx context.Context, dockerImage string) (bool, error) {
```
tedim52: similar to above maybe change to `FetchImageIfMissing` - I think `If` expresses the conditional behavior of the function. Without reading the function spec, `FetchImageMissing` is a bit hard to interpret when reading code.

cli/cli/commands/run/run.go
```diff
@@ -608,6 +624,23 @@ func parseVerbosityFlag(flags *flags.ParsedFlags) (command_args_run.Verbosity, e
...
+// Get the image download flag is present, and parse it to a valid ImageDownload value
+func parseImageDownloadFlag(flags *flags.ParsedFlags) (kurtosis_core_rpc_api_bindings.ImageDownloadMode, error) {
+
+	defaultImageMode := kurtosis_core_rpc_api_bindings.ImageDownloadMode_missing
+
+	imageDownloadStr, err := flags.GetString(imageDownloadFlagKey)
+	if err != nil {
+		return defaultImageMode, stacktrace.Propagate(err, "An error occurred getting the image-download using flag key '%s'", imageDownloadFlagKey)
+	}
+	imageDownload, err := command_args_run.ImageDownloadString(imageDownloadStr)
```
tedim52: I propose we remove the `image_download_enumer` from `command_args_run` and translate `imageDownloadStr` straight to `kurtosis_core_rpc_api_bindings.ImageDownloadMode`. 

This is the only place that `image_download_enumer` is used and doesn't seeeem absolutely necessary, so ideally we could also remove the extra representation of the field in CLI so there are less places to keep track of it. Sorry I didn't catch this on my initial review.

core/server/api_container/server/api_container_service.go
```diff
@@ -137,14 +140,17 @@ func (apicService *ApiContainerService) RunStarlarkScript(args *kurtosis_core_rp
 	experimentalFeatures := args.GetExperimentalFeatures()
 	cloudUserId := shared_utils.GetOrDefaultString(args.CloudUserId, defaultCloudUserId)
 	cloudInstanceID := shared_utils.GetOrDefaultString(args.CloudInstanceId, defaultCloudInstanceId)
+	ApiDownloadMode := shared_utils.GetOrDefault(args.ImageDownloadMode, defaultImageDownloadMode)
+
+	downloadMode := image_download_mode.FromAPI(ApiDownloadMode)
```
tedim52: ahh - right idea here, we need some way to translate the representation of `imageDownloadMode` `kurtosis_core_rpc_api_bindings` to the representation in `container_engine_lib`. 

However, I propose to do this, we use an existing pattern in the codebase. 

First, let's change the representation of `image download` mode in `container-engine-lib` to how it's done for in `[service status](https://github.com/kurtosis-tech/kurtosis/blob/main/container-engine-lib/lib/backend_interface/objects/service/service_status.go)` 

Then, let's do the translation of `kurtosis_core_rpc_api_bindings.ImageDownloadMode` to `image_download_mode.ImageDownloadMode` to look like how it's done for service status [here](https://github.com/kurtosis-tech/kurtosis/blob/563798c82954d807d926d8e92d67bc39c1b832eb/core/server/api_container/server/api_container_service.go#L944).

All this is doing is taking the existing `switch` statement/translation you have in the `container-engine-lib` and moving it to this file in `apic`, but this pattern is preferred bc it removes the dependency of `container-engine-lib` on `apic`(`container-engine-lib` doesn't need to know about how `apic` is representing `ImageDownloadMode`, only `apic` needs to know about that)

This abides by the dependency flow of our system in the [world map](https://github.com/kurtosis-tech/kurtosis-internal-docs#world-map).

docs/docs/cli-reference/run.md
```diff
@@ -121,6 +121,8 @@ run(plan, some_parameter = struct(some_property = "Property value"))
 
 1. The `--no-connect` flag can be used to disable user services port forwarding (default behavior is to forward the ports)
 
+1. The `--image-download` flag can be used to determine what is the behavior to fetch images tagged as latest during the package deployment. Use `missing` to fetch the latest image only if not available in local cache and use `always` to always fetch the latest image from remote repository (default behavior is `missing`).
```
tedim52, line 129: docs driven development : )

core/server/api_container/server/startosis_engine/startosis_validator/docker_images_validator.go
```diff
@@ -43,13 +45,13 @@ func (validator *DockerImagesValidator) Validate(ctx context.Context, environmen
...
 		wg.Add(1)
-		go fetchImageFromBackend(ctx, wg, imageCurrentlyDownloading, validator.kurtosisBackend, image, pullErrors, imageDownloadStarted, imageDownloadFinished)
+		go fetchImageFromBackend(ctx, wg, imageCurrentlyDownloading, validator.kurtosisBackend, image, environment.imageDownloadMode, pullErrors, imageDownloadStarted, imageDownloadFinished)
 	}
 	wg.Wait()
 	logrus.Debug("All image validation submitted, currently in progress.")
 }
 
-func fetchImageFromBackend(ctx context.Context, wg *sync.WaitGroup, imageCurrentlyDownloading chan bool, backend *backend_interface.KurtosisBackend, imageName string, pullErrors chan<- error, imageDownloadStarted chan<- string, imageDownloadFinished chan<- *ValidatedImage) {
+func fetchImageFromBackend(ctx context.Context, wg *sync.WaitGroup, imageCurrentlyDownloading chan bool, backend *backend_interface.KurtosisBackend, imageName string, image_download_mode image_download_mode.ImageDownloadMode, pullErrors chan<- error, imageDownloadStarted chan<- string, imageDownloadFinished chan<- *ValidatedImage) {
```
tedim52: golang does camelCase as opposed to snake_case, thus `s/image_download_mode/imageDownloadMode/` for variable names.

core/server/api_container/server/startosis_engine/startosis_validator/docker_images_validator.go
```diff
@@ -61,7 +63,7 @@ func fetchImageFromBackend(ctx context.Context, wg *sync.WaitGroup, imageCurrent
 	}()
 
 	logrus.Debugf("Starting the download of image: '%s'", imageName)
-	imagePulledFromRemote, err := (*backend).FetchImage(ctx, imageName)
+	imagePulledFromRemote, err := (*backend).FetchImage(ctx, imageName, image_download_mode)
```
tedim52: `s/image_download_mode/imageDownloadMode/`

core/server/api_container/server/startosis_engine/startosis_validator/validator_environment.go
```diff
@@ -18,9 +19,10 @@ type ValidatorEnvironment struct {
 	isResourceInformationComplete bool
 	minCPUByServiceName           map[service.ServiceName]compute_resources.CpuMilliCores
 	minMemoryByServiceName        map[service.ServiceName]compute_resources.MemoryInMegaBytes
+	imageDownloadMode             image_download_mode.ImageDownloadMode
 }
 
-func NewValidatorEnvironment(serviceNames map[service.ServiceName]bool, artifactNames map[string]bool, serviceNameToPrivatePortIds map[service.ServiceName][]string, availableCpuInMilliCores compute_resources.CpuMilliCores, availableMemoryInMegaBytes compute_resources.MemoryInMegaBytes, isResourceInformationComplete bool) *ValidatorEnvironment {
+func NewValidatorEnvironment(serviceNames map[service.ServiceName]bool, artifactNames map[string]bool, serviceNameToPrivatePortIds map[service.ServiceName][]string, availableCpuInMilliCores compute_resources.CpuMilliCores, availableMemoryInMegaBytes compute_resources.MemoryInMegaBytes, isResourceInformationComplete bool, image_download_mode image_download_mode.ImageDownloadMode) *ValidatorEnvironment {
```
tedim52: `s/image_download_mode/imageDownloadMode/`

container-engine-lib/lib/backend_interface/objects/image_download_mode/image_download_mode.go
```diff
@@ -0,0 +1,26 @@
+/*
```
tedim52, line 1: in accord with the comment above, i propose we change this to reflect [service status](https://github.com/kurtosis-tech/kurtosis/blob/main/container-engine-lib/lib/backend_interface/objects/service/service_status.go)

core/server/api_container/server/startosis_engine/startosis_validator.go
```diff
@@ -46,7 +48,7 @@ func NewStartosisValidator(kurtosisBackend *backend_interface.KurtosisBackend, s
 	}
 }
 
-func (validator *StartosisValidator) Validate(ctx context.Context, instructionsSequence []*instructions_plan.ScheduledInstruction) <-chan *kurtosis_core_rpc_api_bindings.StarlarkRunResponseLine {
+func (validator *StartosisValidator) Validate(ctx context.Context, instructionsSequence []*instructions_plan.ScheduledInstruction, image_download_mode image_download_mode.ImageDownloadMode) <-chan *kurtosis_core_rpc_api_bindings.StarlarkRunResponseLine {
```
tedim52: `s/image_download_mode/imageDownloadMode`

tedim52 (changes requested):
looking very good, functionality is definitely there, jus requesting some changes around how we're representing image download mode to follow existing patterns in the codebase

cli/cli/commands/run/run.go
lostbean: No good apart from `nil` breaking the code because is the type is not a pointer. I can make it a pointer but not sure the tradeoffs.

cli/cli/commands/run/run.go
lostbean: Not sure if I agree with skip the internal data structure that represents the concept of the flag (the enum) but I see point that is not needed and this pattern is already used by `parseExperimentalFlag` (although `parseVerbosityFlag` which is very similar to image download uses a intermediary enumeration)

core/server/api_container/server/startosis_engine/startosis_validator/docker_images_validator.go
lostbean: thanks for clarifying it, I found it a bit confusing (even in golang itself is confused about it: https://github.com/golang/go/issues/36060). For instance, filename are snake_case and vars/func are CamelCase but protobuf definition has CamelCase for message names but snake_case for fields which generates :p

lostbean: oh, it would be nice to if we can find a linter that could help to enforce it.

cli/cli/commands/run/run.go
tedim52: fair were also a bit inconsistent in how we're representing flags. I interpreted the fact that `command_args_run` only being used for `verbosity` in it meant that the module was only for representing arguments that stay in the CLI/related to CLI (verbosity doesn't travel to apic/container engine lib). I don't have high conviction about this one so I'll leave it up to you!

cli/cli/commands/run/run.go
```diff
@@ -631,12 +631,18 @@ func parseImageDownloadFlag(flags *flags.ParsedFlags) (*kurtosis_core_rpc_api_bi
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "An error occurred getting the image-download using flag key '%s'", imageDownloadFlagKey)
 	}
-	imageDownload, err := command_args_run.ImageDownloadString(imageDownloadStr)
-	if err != nil {
-		return nil, stacktrace.Propagate(err, "Invalid image-download value: '%s'. Possible values are %s", imageDownloadStr, strings.Join(command_args_run.ImageDownloadStrings(), ", "))
+
+	imageDownloadCode := kurtosis_core_rpc_api_bindings.ImageDownloadMode_value[strings.ToLower(imageDownloadStr)]
+	if imageDownloadCode >= 1 {
```
tedim52: no need to change this but FWIW, often the codebase follows an ["eject early, eject often"](https://github.com/kurtosis-tech/kurtosis-internal-docs/blob/main/general-style-guide.md#default-to-eject-early-eject-often) pattern for conditional branches, just a stylistic choice so instead of 
```
if (success) {
    // do success code
} else {
    return err
}
```
we do 
``` 
if (error) {
    return err
}
// do success code
```
where it makes sense ofc

tedim52: also we try and avoid [magic constants](https://github.com/kurtosis-tech/kurtosis-internal-docs/blob/main/general-style-guide.md#replace-magic-constants-with-constants-named-for-their-purpose-not-their-value) - maybe `s/1/maxImageDownloadCode`? where `maxImageDownloadCode` is a const.

tedim52 (approved):
looks great! thanks for addressing comments, last thing is magic constant comment and then good to merge!

Issue #1434, opened 2023-09-29
Allow kurtosis to be used behind proxy


vanderheijden86:
### Background & motivation

When currently using kurtosis cli behind a corporate proxy, you run into issues with cloning git repositories (for example). 

```
INFO[2023-09-26T16:19:41+02:00] Uploading and executing package 'github.com/gobitfly/eth2-beaconchain-explorer/local-testnet' 
There was an error interpreting Starlark code 
Evaluation error: An error occurred while loading the module 'github.com/kurtosis-tech/eth2-package/src/package_io/parse_input.star'
    Caused by: Error in cloning git repository 'https://github.com/kurtosis-tech/eth2-package.git' to '/kurtosis-data/tmp-startosis-packages/tmp-repo-dir-383024695/kurtosis-tech/eth2-package'. Make sure that 'https://github.com/kurtosis-tech/eth2-package.git' exists and is a public repository.
    at [github.com/gobitfly/eth2-beaconchain-explorer/local-testnet/main.star:1:28]: <toplevel>

Error encountered running Starlark code.
```

This is caused by the fact that kurtosis doesn't pick up designated proxy environment variables like http_proxy, https_proxy, ftp_proxy and no_proxy. 

### Desired behaviour

In order for kurtosis to work behind a proxy, it would need to pick up these variables and pass them through to the relevant commands (like git) inside of it that connect to the internet via this proxy. 

At minimum these variables should be passed through to the git library used in kurtosis, but potentially also in other libraries or executables that are called via kurtosis. By doing so, users operating behind a proxy can use Kurtosis without encountering connectivity issues related to proxy restrictions.

### How important is this to you?

Critical; Kurtosis is unusable for me without it.

mieubrisse:
Hey @vanderheijden86 , thanks a ton for suggesting this! As @tedim52 mentioned in Discord, this is probably easiest to do by adding optional keys to the Kurtosis config file that then get passed in to the engine, which in turn will pass it to the Kurtosis API container when it creates enclaves ([architecture reference](https://docs.kurtosis.com/explanations/architecture)).

vanderheijden86:
[...]

PR #1496, opened 2023-10-07
feat: allow kurtosis cli to be used behind proxy

vanderheijden86:
## Description:
This update introduces the ability to specify a proxy URL for git operations, facilitating the cloning of repositories when behind a (corporate) proxy. Initially, this draft PR allows setting just the proxy URL, but the plan is to expand this by including a full proxy options struct that go-git utilizes, encompassing not just the URL but also the username and password for proxy authentication. The struct is outlined as follows:

```go
type ProxyOptions struct {
	URL      string
	Username string
	Password string
}
```

- [ ] Add proper test for loading of proxy config into `git_package_content_provider.go`
- [ ] Provide entire ProxyOptions struct as configurable parameter instead of just URL
- [ ] Add test to verify that ProxyOptions is picked when cloning a repo
- [ ] Fix linting issues

YES

https://github.com/kurtosis-tech/kurtosis/issues/1434

### DIFF
--- a/cli/cli/kurtosis_config/overrides_objects/v3/kurtosis_config_v3.go
+++ b/cli/cli/kurtosis_config/overrides_objects/v3/kurtosis_config_v3.go
@@ -1,0 +1,27 @@
+package v3
+
+import "github.com/kurtosis-tech/kurtosis/cli/cli/kurtosis_config/config_version"
+
+/*
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! WARNING !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+                           DO NOT CHANGE THIS FILE!
+  If you change this file, it will break config for users who have instantiated an
+           overrides file with this version of config overrides!
+    Instead, to make changes, you will need to add a new version of the config
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! WARNING !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+*/
+
+// NOTE: All new YAML property names here should be kebab-case because
+//  1. it's easier to read
+//  2. it's easier to write
+//  3. it's consistent with previous properties and changing the format of an already-written config file is very difficult
+type KurtosisConfigV3 struct {
+	// vvvvvvvvv Every new Kurtosis config version must have this key vvvvvvvv
+	ConfigVersion config_version.ConfigVersion `yaml:"config-version"`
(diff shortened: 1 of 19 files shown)


core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider.go
```diff
@@ -362,6 +373,16 @@ func (provider *GitPackageContentProvider) atomicClone(parsedURL *ParsedGitURL)
 	return nil
 }
 
+func getKurtosisConfig() (*resolved_config.KurtosisConfig, error) {
+	configStore := kurtosis_config.GetKurtosisConfigStore()
```
tedim52: ah so unfortunately we can't use `kurtosis_config.GetKurtosisConfigStore()` to retrieve the config because it pulls the config from a file on the filesystem. The `git_package_content_provider.go` logic lives and is executed in our `APIC` / `core` server, a container for managing each individual enclave - whereas the kurtosis config lives on the machine where the CLI is running. Bc of this we'll need to try another way of getting the proxy options into the `APIC`.

tedim52: The best way to do this is to pass the proxy config option from the `CLI` into the `engine` and then from the `engine` into the `APIC` when the enclave starts a new enclave. Going to look for links + reference implementations of this pattern to guide you here and will send via discord.

cli/cli/kurtosis_config/overrides_objects/v3/kurtosis_config_v3.go
```diff
@@ -0,0 +1,27 @@
...
+//  1. it's easier to read
+//  2. it's easier to write
+//  3. it's consistent with previous properties and changing the format of an already-written config file is very difficult
+type KurtosisConfigV3 struct {
+	// vvvvvvvvv Every new Kurtosis config version must have this key vvvvvvvv
+	ConfigVersion config_version.ConfigVersion `yaml:"config-version"`
+	// ^^^^^^^^^ Every new Kurtosis config version must have this key ^^^^^^^^
+
+	ShouldSendMetrics *bool                               `yaml:"should-send-metrics,omitempty"`
+	GitProxy          *string                             `yaml:"git-proxy,omitempty"`
```
tedim52: Another note, let's move `GitProxy` from the root level config to cluster config level (so it can be configured per cluster) so in `KurtosisClusterConfigV3`. Clusters can live at different locations - the default one is the docker cluster on the user machine, but the same proxy may not apply to every cluster so proxy should be configured depending on each cluster.

barnabasbusa:

PR #1498, opened 2023-10-09
fix: improve absolute locator checks

h4ck3rk3y:
## Description:
if we had something like

github.com/x/y calling github.com/a/b which uploads github.com/x/y/foo.star; we would have failures as absolute checks would be from running package; this fixes that and catches absolute calls for children packages as well

further improved the branch/tag/file parsing

YES

### DIFF
--- a/core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/builtin_argument/validators.go
+++ b/core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/builtin_argument/validators.go
@@ -110,16 +110,3 @@ func DurationOrNone(value starlark.Value, attributeName string) *startosis_error
 	}
 	return nil
 }
-
-func LocatorValidator(locatorStarlarkValue starlark.Value, packageId string, argNameForLogging string) *startosis_errors.InterpretationError {
-
-	if err := NonEmptyString(locatorStarlarkValue, argNameForLogging); err != nil {
-		return err
-	}
-
-	_, ok := locatorStarlarkValue.(starlark.String)
-	if !ok {
-		return startosis_errors.NewInterpretationError("Value for '%s' was expected to be a starlark.String but was '%s'", argNameForLogging, reflect.TypeOf(locatorStarlarkValue))
-	}
-	return nil
-}
--- a/core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/builtin_argument/validators.go
+++ b/core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/builtin_argument/validators.go
@@ -2,6 +2,7 @@ package builtin_argument
 
 import (
 	"github.com/kurtosis-tech/kurtosis/core/server/api_container/server/startosis_engine/startosis_errors"
+	"github.com/sirupsen/logrus"
 	"go.starlark.net/starlark"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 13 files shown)


core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/parsed_git_url.go
```diff
@@ -130,16 +132,28 @@ func parseOutTagBranchOrCommit(input string) (string, string) {
 	cleanInput := path.Clean(input)
 	pathWithoutVersion, maybeTagBranchOrCommitWithFile, _ := strings.Cut(cleanInput, tagBranchOrCommitDelimiter)
 
-	// input can have been set with version in two diff ways
+	// input can have been set with version in few diff ways
 	// 1- github.com/kurtosis-tech/sample-dependency-package/main.star@branch-or-version (when is called from cli run command)
 	// 2- github.com/kurtosis-tech/sample-dependency-package@branch-or-version/main.star (when is declared in the replace section of the kurtosis.yml file)
+	// 3- github.com/kurtosis-tech/sample-dependency-package/main.star@foo/bar - here the tag is foo/bar;
```
h4ck3rk3y, line 138: @leoporoli i am changing how parsing works to allow for `/` in the branch name

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/builtin_argument/validators.go
```diff
@@ -111,25 +111,15 @@ func DurationOrNone(value starlark.Value, attributeName string) *startosis_error
 	return nil
 }
 
-func RelativeOrRemoteAbsoluteLocator(locatorStarlarkValue starlark.Value, packageId string, argNameForLogging string) *startosis_errors.InterpretationError {
+func LocatorValidator(locatorStarlarkValue starlark.Value, packageId string, argNameForLogging string) *startosis_errors.InterpretationError {
```
h4ck3rk3y: the problem is that NewUploadFiles and NewReadFiles and NewImportModule is only built once and against the main package id

so that doesn't work if a child uses a passed import from a main package id

also it doesn't catch the following github.com/x/y import github.com/a/b which imports github.com/a/b/lib.star

leoporoli: @h4ck3rk3y could you give me an example of the first part?

h4ck3rk3y: image

basically the pattern Carlos from Nethermind is trying

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/parsed_git_url.go
leoporoli: @h4ck3rk3y great!

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/builtin_argument/validators.go
leoporoli: @h4ck3rk3y so you moved this validation inside `git package content provider` right?

leoporoli: @h4ck3rk3y I think that this validation is no longer essential and we can remove it and use `NonEmptyString` validation instead

leoporoli (approved):

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/builtin_argument/validators.go
h4ck3rk3y: Point taken! removed :)

Issue #1505, opened 2023-10-10
Logs not cleaned even on 0.84.3


mieubrisse:
### What's your CLI version?

0.84.3

### Description & steps to reproduce

On 0.84.2, my `kurtosis-logs-storage` was taking ~10GB of space:
```
Local Volumes space usage:

VOLUME NAME                                                        LINKS     SIZE
minikube                                                           1         924MB
kurtosis-logs-storage                                              2         9.277GB
```

I upgraded to 0.84.3:
```
$ kurtosis version
CLI Version:   0.84.3

To see the engine version (provided it is running): kurtosis engine status
```
ran `kurtosis clean -a`:
```
$ kurtosis clean -a
INFO[2023-10-10T09:16:07-03:00] Cleaning old Kurtosis engine containers...
INFO[2023-10-10T09:16:07-03:00] Successfully cleaned old Kurtosis engine containers
INFO[2023-10-10T09:16:07-03:00] Cleaning enclaves...
INFO[2023-10-10T09:16:07-03:00] Successfully cleaned enclaves
INFO[2023-10-10T09:16:07-03:00] Cleaning unused images...
INFO[2023-10-10T09:16:07-03:00] Successfully removed the following unused images:
kurtosistech/engine:0.84.2
INFO[2023-10-10T09:16:07-03:00] Successfully cleaned unused images
```
I now have no enclaves:
```
$ kurtosis enclave ls
UUID   Name   Status   Creation Time
```
yet, my log volume is still enormous:
```
$ docker system df -v
....bunch of omitted stuff....

Local Volumes space usage:

VOLUME NAME                                                        LINKS     SIZE
692ebadfc68f9f082fd4fbfd8e0a729f8374cdd7b2d24cbb4d82dd9dfd821ee8   0         39.91MB
kurtosis-logs-storage                                              2         9.277GB
```

### Desired behavior

Post-0.84.3, my `kurtosis-logs-storage` should have shrank tremendously

### What is the severity of this bug?

Painful; this is causing significant friction in my workflow.

### What area of the product does this pertain to?

CLI: the Command Line Interface

PR #1517, opened 2023-10-10
feat: make clean -a remove all logs

tedim52:
## Description:
Prior to this `kurtosis clean` removed logs for existing running and stopped enclaves. Now, `kurtosis clean -a` will get rid of all logs from this year. 

YES

## References
https://github.com/kurtosis-tech/kurtosis/issues/1505

### DIFF
--- a/engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
+++ b/engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
@@ -136,7 +136,8 @@ func (manager *LogFileManager) RemoveLogsBeyondRetentionPeriod() {
 }
 
 func (manager *LogFileManager) RemoveAllLogs() error {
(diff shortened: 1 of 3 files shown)


engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
```diff
@@ -136,7 +136,9 @@ func (manager *LogFileManager) RemoveLogsBeyondRetentionPeriod() {
 }
 
 func (manager *LogFileManager) RemoveAllLogs() error {
-	if err := manager.filesystem.RemoveAll(volume_consts.LogsStorageDirpath); err != nil {
+	// only removes logs for this year because Docker prevents all logs from base logs storage file path
```
tedim52, line 139: doing only for this year is a hack to get rid of all existing logs storage bc removing logs at the base path isn't allowed by Docker (for some reason) - but it should do the trick for clearing up all historical logs that are taking up current user log storage

leoporoli (approved):

Issue #1536, opened 2023-10-12
Missing documentation on Contexts 


omar711:
### Where did you see the issue in our documentation?

https://docs.kurtosis.com/cli

### Issue description

`kurtosis context` commands are missing documentation.   

I've also not found anything that explains what a context is.  Possibly https://docs.kurtosis.com/cloud touches on it but I'm not certain.

leeederek:
@omar711 thanks for filing this! Let me take care of this. I also notice that we don't have any docs for "portal" either - will add those as well.

leeederek:
Update: we have plans to remove the `context` concept altogether and opted to expose end users to just `kurtosis cloud` commands, and so the PR i've linked (#1557) will be for cloud cli commands and has no mention of contexts. I believe this addresses the pain points highlighted in this issue.

omar711:
Sounds great @leeederek

Issue #1508, opened 2023-10-10
In cluster ls have an * next to the currently chosen cluster


h4ck3rk3y:
### Background & motivation

This would match context ls

### Desired behaviour

kurtosis cluster ls

looks like

* docker
kubernetes
minikube

### How important is this to you?

Nice to have; this feature would make using Kurtosis more enjoyable.

galenmarchetti:
Would also be great to match the verbs in Kurtosis context and Kurtosis cluster ... right now one uses switch and the other uses set, and I keep mixing them up

omar711:
Small issue, on a new environment, the cluster-settings file doesn't exist yet, so `kurtosis cluster get` will fail:

```
√ kurtosis % kurtosis cluster get
Error:  An error occurred running command 'get'
  Caused by: Failed to get cluster setting.
  Caused by: Failed to get cluster setting from setting file.
  Caused by: An error occurred reading cluster setting file
  Caused by: open ~/Library/Application Support/kurtosis/cluster-setting: no such file or directory
```

omar711:
@galenmarchetti  @h4ck3rk3y I've updated this to match context, and also used the table_printer used there to match.  Currently it looks like this:

```
√ kurtosis % ./cli/cli/scripts/launch-cli.sh cluster ls
    Name
*   docker
    minikube
```

The column header is new, due to the table printer.  Is that ok or do you prefer the plain version?

omar711:
> Would also be great to match the verbs in Kurtosis context and Kurtosis cluster ... right now one uses switch and the other uses set, and I keep mixing them up

on this @galenmarchetti : do you prefer set or switch?  Any concerns around changing one (e.g. in cases users have scripted it)?

galenmarchetti:
great! I remember @mieubrisse had a strong opinion about this one, maybe @h4ck3rk3y  remembers this...one of these verbs represents idempotent changes so both should be whichever that verb is. i'm butchering it. @mieubrisse  thoughts?

galenmarchetti:
confirmed with @mieubrisse  it should be `set` because its a natural pairing with `get` and also implies idempotency

omar711:
`cluster ls` is done now https://github.com/kurtosis-tech/kurtosis/pull/1514 
I'll keep this open and tackle the `context switch/set` part in a separate pr tomorrow

omar711:
[...]

PR #1537, opened 2023-10-12
feat: rename kurtosis context "switch" to "set"

omar711:
## Description:
Changes the command `kurtosis context switch` to `kurtosis context set`, to match the similar command `kurtosis cluster set`.

I've changed the cli command itself, plus the context config store.  There's also the portal manager, which I've left as switch for now, as it extends into the kurtosis-portal project; I think we'll also need to do that?

We're missing documentation on contexts and this cli command.  I don't know what to write for it yet so I've added https://github.com/kurtosis-tech/kurtosis/issues/1536 

YES

https://github.com/kurtosis-tech/kurtosis/issues/1508


cli/cli/commands/cloud/load/load.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/kurtosis-tech/kurtosis/cli/cli/command_framework/lowlevel/flags"
 	"github.com/kurtosis-tech/kurtosis/cli/cli/command_str_consts"
 	"github.com/kurtosis-tech/kurtosis/cli/cli/commands/kurtosis_context/add"
-	"github.com/kurtosis-tech/kurtosis/cli/cli/commands/kurtosis_context/context_switch"
+	"github.com/kurtosis-tech/kurtosis/cli/cli/commands/kurtosis_context/set"
```
omar711, line 14: I've renamed this package too to mirror the other commands

mieubrisse: thank you!!

mieubrisse (approved):

PR #1545, opened 2023-10-12
docs: local replace package dependency

leoporoli:
## Description:
local replace package dependency docs

YES

This is part of the [Forked cleanup package project](https://www.notion.so/kurtosistech/Forked-Package-Cleanup-16d86c4e274547b28496f17154bf3d62)


leoporoli:
@leeederek I had to create the PR again because the previous one was accidentally closed, it's better because it has only the docs changes

laurentluce (approved):

Issue #1152, opened 2023-08-24
Improve SDK functionality and user experience


h4ck3rk3y:
### Background & motivation

There's a lot of boiler plating that needs to be done to get the SDK to run for example to run a simple package we have to produce this block

```
	// set up the input parameters
	logrus.Info("Parsing Input Parameters")
	inputParameters, err := os.ReadFile(inputFile)
	require.NoError(t, err, "An error occurred while reading the input file")
	require.NotEmpty(t, inputParameters, "Input parameters byte array is unexpectedly empty")
	inputParametersAsJSONString := string(inputParameters)
	require.NotEmpty(t, inputParametersAsJSONString, "Input parameter json string is unexpectedly empty")

	// set up enclave
	logrus.Info("Setting up Kurtosis Engine Connection & Enclave")
	ctx, cancelCtxFunc := context.WithCancel(context.Background())
	defer cancelCtxFunc()
	kurtosisCtx, err := kurtosis_context.NewKurtosisContextFromLocalEngine()
	require.NoError(t, err, "An error occurred while creating Kurtosis Context")

	enclaveName := fmt.Sprintf("%s-%d", enclaveNamePrefix, time.Now().Unix())
	enclaveCtx, err := kurtosisCtx.CreateEnclave(ctx, enclaveName)
	require.Nil(t, err, "An unexpected error occurred while creating Enclave Context")
	//defer kurtosisCtx.DestroyEnclave(ctx, enclaveName)

	// execute package
	logrus.Info("Executing the Starlark Package")
	packageRunResult, err := enclaveCtx.RunStarlarkRemotePackageBlocking(ctx, eth2Package, pathToMainFile, runFunctionName, inputParametersAsJSONString, isNotDryRun, defaultParallelism, noExperimentalFeatureFlags)
	require.NoError(t, err, "An unexpected error occurred while executing the package")
	require.Nil(t, packageRunResult.InterpretationError)
	require.Empty(t, packageRunResult.ValidationErrors)
	require.Nil(t, packageRunResult.ExecutionError)
```

One needs to know a lot of terms in order to get this going

1. What the KurtosisContext and engine's are
2. Enclaves
3. Enclave Creation before package execution
[...]

PR #1550, opened 2023-10-13
feat: Add create enclave utils to SDK

victorcolombo:
## Description:

A very idiomatic pattern in Go is to use `defer` and closer functions to cleanup resources, instead of relying on memory to call the correct SDK calls when the test/setup is done. This is very common on our test suit that we have a util, that now is exposed to an external user.

This PR also introduces a single SDK call to get from starlark to enclave, something that was not possible before.

YES

https://github.com/kurtosis-tech/kurtosis/issues/1152


victorcolombo:
It's unclear for me if this makes sense in TypeScript, since it doesn't feel very idiomatic

api/golang/util/create_enclave.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"context"
+	"github.com/kurtosis-tech/kurtosis/api/golang/core/lib/enclaves"
+	"github.com/kurtosis-tech/kurtosis/api/golang/core/lib/starlark_run_config"
+	"github.com/kurtosis-tech/kurtosis/api/golang/engine/lib/kurtosis_context"
+	"github.com/kurtosis-tech/stacktrace"
+)
+
+func CreateEmptyEnclave(ctx context.Context, enclaveName string) (*enclaves.EnclaveContext, func() error, func() error, error) {
```
h4ck3rk3y, line 11: Can we add docs for this?

h4ck3rk3y (approved):

PR #1553, opened 2023-10-13
fix: error clearly if there are no nodes on the Kubernetes cluster

### LINKED ISSUE #1507: fail with a precise error if Kubernetes cluster has no node groups
### Background & motivation

@laurentluce  ran into this when he set up Kubernetes for the first time and warned me when I was setting up my Kubernetes cluster

@galenmarchetti ran into this yesterday and I had to come in and help him

### Desired behaviour

The failure when there are no node groups is more obvious than it currently is

h4ck3rk3y:
## Description:
An AWS EKS cluster starts with no nodes which is quite confusing for the user if the engine doesn't start.  This checks if there are any nodes and otherwise tells the user there aren't

YES

#1507


h4ck3rk3y:
image

leoporoli (approved):

PR #1560, opened 2023-10-16
fix: show container status instead of service status in enclave inspect

h4ck3rk3y:
## Description:
If a service eventually dies we would still show `RUNNING` while in reality it would be stopped

1. Create an enclave
2. Stop a service
3. Use inspect -> this prints service status Running (Kurtosis was told to keep it running)

Post this PR

3. This would print container status -> Stopped - actually reflecting reality

## Is this change user facing?
YES

## References (if applicable):
Closes #1351

PR #1567, opened 2023-10-16
fix: show container status instead of service status  in UI

tedim52:
## Description:
This change is the sister to this PR https://github.com/kurtosis-tech/kurtosis/pull/1560 which makes the same change in the CLI.

YES

## References: 
https://github.com/kurtosis-tech/kurtosis/pull/1560


engine/frontend/src/component/ServiceView.js
```diff
@@ -65,7 +65,7 @@ const ServiceView = (service) => {
                 <Tbody>
                     {tableRow("Name", () => serviceInfo.name)}
                     {tableRow("UUID", () => <pre>{serviceInfo.serviceUuid}</pre>)}
-                    {tableRow("Status", () => statusBadge(serviceInfo.serviceStatus))}
+                    {tableRow("Status", () => statusBadge(serviceInfo.getContainer().Status()))}
```
tedim52: Screen Shot 2023-10-16 at 3 15 18 PM
hmm for some reason Status isn't being returned in the service info ill check why

tedim52: oops fixed

engine/server/webapp/asset-manifest.json
```diff
@@ -1,13 +1,13 @@
 {
   "files": {
     "main.css": "./static/css/main.d56b8d0f.css",
```
h4ck3rk3y: wonder if all of this needs to be checked in

h4ck3rk3y (approved):

engine/server/webapp/asset-manifest.json
tedim52: yes, these are files that are a result of rebuilding the frontend via `npm run build` in `engine/frontend`

PR #1569, opened 2023-10-16
feat: Cloud backend method to refresh the default payment method

laurentluce:
## Description:
Cloud backend API method to go with https://github.com/kurtosis-tech/kurtosis-cloud-backend/pull/59

NO


cloud/api/protobuf/kurtosis_backend_server_api.proto
```diff
@@ -13,6 +13,7 @@ service KurtosisCloudBackendServer {
   rpc GetOrCreateApiKey(GetOrCreateApiKeyRequest) returns (GetOrCreateApiKeyResponse) {};
   rpc GetOrCreateInstance(GetOrCreateInstanceRequest) returns (GetOrCreateInstanceResponse) {};
   rpc GetOrCreatePaymentConfig(GetOrCreatePaymentConfigArgs) returns (GetOrCreatePaymentConfigResponse) {};
+  rpc RefreshDefaultPaymentMethod (RefreshDefaultPaymentMethodArgs) returns (google.protobuf.Empty) {};
```
adschwartz, line 16: Not sure if it's useful, but would it perhaps make sense to return some receipt that the payment method was indeed refreshed (other than  the absence of error)? 

I can live without this but just wanted to propose it as an option.

adschwartz (approved):

cloud/api/protobuf/kurtosis_backend_server_api.proto
laurentluce: I am going to keep it as-is since no error means the default payment method has been refreshed and the frontend does not need more information.

PR #1586, opened 2023-10-18
refactor!: protobuf definitions for more idiomatic SDKs

### LINKED ISSUE #843: Refactor Protobuf definitions for more idiomatic SDKs
### Background & motivation

For example, `CreateEnclave` call takes in a optional parameters like `api_container_version_tag`. This field is not tagged as optional in the protobuf.

This is not a problem for Golang, where you can construct a struct by passing only a subset of the parameters, defaulting omitted values. In this specific case, building `CreateEnclaveRequest` omitting `api_container_version_tag` defaults it to `""`, which we interpret as `empty` in the backend.

In Rust, this is problematic, since there is a clear difference between null value (`""`) and absent value (`Option::None`), as seen in our current SDK: https://crates.io/crates/kurtosis-sdk/0.80.7

### Desired behaviour

Review our APIs, adding explicit `optional` field for elements that can be semantically absent (optional) vs protocol absent (regular field). See https://protobuf.dev/programming-guides/proto3/#field-labels

leoporoli:
## Description:
refactor protobuf definitions for more idiomatic SDKs

NO

Fix #843


engine/server/engine/server/engine_connect_server_service.go
```diff
@@ -75,21 +75,21 @@ func (service *EngineConnectServerService) GetEngineInfo(context.Context, *conne
...
 	isProduction := false
-	if args.Mode == kurtosis_engine_rpc_api_bindings.EnclaveMode_PRODUCTION {
+	if args.GetMode() == kurtosis_engine_rpc_api_bindings.EnclaveMode_PRODUCTION {
 		isProduction = true
 	}
 	enclaveInfo, err := service.enclaveManager.CreateEnclave(
 		ctx,
 		service.imageVersionTag,
-		args.ApiContainerVersionTag,
+		args.GetApiContainerVersionTag(),
```
h4ck3rk3y, line 90: Do we want `GetXXXX` and not just `ApiContainerVersionTag` I can't tell why one is preferred over the other but I don't like the new `&shouldFollowAll` syntax

leoporoli: The diff between using the field or the getter is that the getter returns the zero value (which is the default value in many cases) if it is `nil`

image

Why you don't like `&shouldFollowAll` syntax ? What do you suggest to use instead? 

What we are trying to solve with this modifications is this:
```
In Rust, this is problematic, since there is a clear difference between null value ("") and absent value (Option::None), as seen in our current SDK: https://crates.io/crates/kurtosis-sdk/0.80.7
```

leoporoli: Users won't have to use string pointers when using the SDK, it's internal. and only for setting the protobuf arguments, and it's necessary to do it on this way if we want to use optional arguments

h4ck3rk3y: That makes sense. Should we do any docs changes as well?

leoporoli: we don't need to change the docs because all the changes are internal, not changes on the SDK's API

h4ck3rk3y (approved):

PR #1590, opened 2023-10-18
feat: get package on demand

adschwartz:
## Description:
it's now possible to load a custom package using a package url param, e.g.:
```
http://localhost:3000/enclave-manager/?package=github.com%2Fadschwartz%2Fbasic-service-package
```

YES


engine/frontend/src/api/packageCatalog.js
```diff
@@ -1,16 +1,84 @@
...
+    return getSinglePackageManually(
+        parts.base,
+        parts.owner,
+        parts.name,
+    )
+}
+
+export const getOwnerNameFromUrl = (fullUrl) => {
+    const components = fullUrl.split('/');
+    if (components.length < 3) {
```
tedim52: magic number? or maybe comment why 3? it's pretty deducible but couldn't hurt (im assuming base owner and name)

tedim52 (approved):
very cool feature

engine/frontend/src/api/packageCatalog.js
adschwartz: This was fixed

PR #1595, opened 2023-10-19
chore: Simplify enclave-manager-sdk typescript dependencies

Dartoxian:
## Description:

The enclave manager sdk has a large number of (I think) unnecessary dependencies, this PR removes a lot of them and tidies up package.json.

Additionally, specify `yarn --frozen-lockfile` in the `enclave-manager/api/typescript/scripts/build.sh` so that builds will fail if dependencies are not committed properly.

This change removes the dependency in the enclave manager from the general typescript `kurtosis-sdk` - I'm not sure why this dependency is present as it's not required by any code generated by the enclave-manager-sdk.

NO

N/A


enclave-manager/api/typescript/package.json
```diff
@@ -1,11 +1,11 @@
 {
+    "version": "0.0.0",
     "name": "enclave-manager-sdk",
     "main": "./build/index",
     "description": "This repo contains a Typescript client for communicating with the Enclave Manager API server.",
     "types": "./build/index",
     "scripts": {
-        "build": "tsc && cp -R src/ build/ && cp -R node_modules/kurtosis-sdk/src/core/kurtosis_core_rpc_api_bindings/connect/* build/ && cp -R node_modules/kurtosis-sdk/src/engine/kurtosis_engine_rpc_api_bindings/connect/* build/",
-        "test": "ts-mocha -p tsconfig.json 'test/**/*.ts'"
```
adschwartz, line 8: We don't have an tests but should we leave this to make it possible to run them?

Dartoxian: I've added back a stub in case anything wants to call `yarn test` in the future.

I don't think we should leave `"test": "ts-mocha -p tsconfig.json 'test/**/*.ts'"` by default as to a reader it looks like it does something, when in fact we know that currently it does nothing.

adschwartz: I see, yes it could be misleading. I do prefer keeping it so that when we _do_ add unit tests they will automatically be picked up and run as part of the build script (also not giving extra burden to others to get unit tests going). This is pattern we have in other places too (especially our go projects): we look for test files and run them if they exist. This means all build scripts are basically identical in function. 

if you know of a better test command to run then please do feel free to change it.

Dartoxian:
@adschwartz thanks for the call earlier, I've now tested this PR inside a locally built kurtosis engine and confirmed it's working.

enclave-manager/api/typescript/package.json
Dartoxian: ACK

I've added Jest as the testing library and a basic test to the package - I'm not sure the tests will ever get more complicated that this for this package (instead integration tests of the enclave manager frontend will be more appropriate).

Dartoxian:
Another tweak to the `build` process of the typescript enclave manager sdk: by copying the generated `kurtosis-sdk` files into `src` pre build, `tsc` is able to generate more accurate types for the enclave manager sdk. The current type definition of the enclave manager sdk contains lots of `any` values, but with this change it now contains the correctly defined types.

enclave-manager/api/typescript/package.json
adschwartz: Perfect. Love this!

adschwartz (approved):

adschwartz:
@Dartoxian I'm ok with this if you have tested it and it works.

PR #1598, opened 2023-10-20
fix: bug in portal forwarding via run

h4ck3rk3y:


adschwartz (approved):

cli/cli/commands/run/run.go
```diff
@@ -425,7 +425,7 @@ func run(
 		return nil
 	}
 
-	if currentContext != nil {
+	if currentContext == nil {
 		return nil
```
omar711, line 429: I think some logging would be good here

PR #1601, opened 2023-10-23
ci: auto approve & merge dependabot PRs

h4ck3rk3y:
## Description:
We have 10s of them; this should merge those which pass checks automatically


omar711 (commented):
Looks ok to me @h4ck3rk3y but I think you should run it by the others.  These things depend on integration test coverage being good enough and I don't have a feel for that yet.

adschwartz (approved):

PR #1603, opened 2023-10-23
feat: initial new enclave manager

Dartoxian:
## Description:
This PR contains the beginnings of the new enclave manager ui. Specifically it adds a new project in `enclave-manager/web`. Notable inclusions:
* Continuing to use Chakra as a component libary, the KurtosisThemeProvider injects components themed based on the mocks.
* The enclave manager sdk cllient is refactored into authenticated (remote) and local versions, and provided to react components with `useKurtosisClient` by the `KurtosisClientProvider`. This provider blocks loading the application (with a spinner) until it's configured the client to use.
* react router is used to compose the application.
* react-table is used to create a (reusable) table for showing the enclaves list.

### Demo (WIP)

https://github.com/kurtosis-tech/kurtosis/assets/4419574/c026a1d9-313d-4e73-944a-2814399e039a

NO (this is not yet part of the build/deploy process)

This PR is rebased on top of https://github.com/kurtosis-tech/kurtosis/pull/1595 which should be merged first.


enclave-manager/web/src/client/KurtosisClientContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+  useEffect(() => {
+    const receiveMessage = (event: MessageEvent) => {
+      const message = event.data.message;
+      switch (message) {
+        case "jwtToken":
+          const value = event.data.value;
+          if (isDefined(value)) {
+            setJwtToken(value);
+          }
+          break;
```
adschwartz: This is a necessary evil that is tied to the loading the kurtosis client with authentication details. Maybe this should be part of the AuthenticatedKurtosisClient?

adschwartz: I have also since added another message to be passed. this is the `package` information that the UI should load. That part is not related to authentication and will be used whether we authenticate in the cloud context or if we run it locally.

enclave-manager/web/src/client/KurtosisClientContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+    window.addEventListener("message", receiveMessage);
+    return () => window.removeEventListener("message", receiveMessage);
+  }, []);
+
+  useEffect(() => {
+    const searchParams = new URLSearchParams(window.location.search);
+    const requireAuth = isStringTrue(searchParams.get("require_authentication"));
+    const requestedApiHost = searchParams.get("api_host");
+    // eslint-disable-next-line
+    const preloadedPackage = searchParams.get("package");
```
adschwartz: this will need to wire in the package into the package configuration page

enclave-manager/web/src/client/KurtosisClientContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+    try {
+      setError(undefined);
+      if (requireAuth) {
```
adschwartz: Is there any way we can move this checking out of the KurtosisClientContext? 
As a minimum I'd like to see if we can store the jwt token in a separate context provider that is tied to the authenticated client. Trying to see if we can minimize the bleeding of the need for the auth in here.
Let me know if we need to talk this over more.

enclave-manager/web/src/client/KurtosisClientContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { AuthenticatedKurtosisClient } from "./AuthenticatedKurtosisClient";
+import { LocalKurtosisClient } from "./LocalKurtosisClient";
+
+type KurtosisClientContextState = {
+  client: KurtosisClient | null;
+};
+
+const KurtosisClientContext = createContext<KurtosisClientContextState>({ client: null });
+
+export const KurtosisClientProvider = ({ children }: PropsWithChildren) => {
```
adschwartz: For this provider we can split the jwtToken handling out but keep the code related to requestedApiHost, preloadedPackage

enclave-manager/web/src/client/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const KURTOSIS_DEFAULT_HOST = "localhost";
+export const KURTOSIS_DEFAULT_PORT = 8081;
+export const KURTOSIS_DEFAULT_URL = `http://${KURTOSIS_DEFAULT_HOST}:${KURTOSIS_DEFAULT_PORT}`;
```
adschwartz: Is there a way we can have different ones for different environments? We have that in the cloud ui.

enclave-manager/web/src/components/AppLayout.tsx
```diff
@@ -0,0 +1,19 @@
...
+type AppLayoutProps = PropsWithChildren<{
+  Nav: React.ReactElement;
+}>;
+
+export const AppLayout = ({ Nav, children }: AppLayoutProps) => {
+  return (
+    <Flex flexDirection={"row"}>
+      {Nav}
+      <Flex as="main" w={"100%"} justifyContent={"flex-start"} p={"20px 40px"} className={"app-container"}>
+        <Flex maxWidth={"1320px"} w={"100%"}>
```
adschwartz: I think we are not making this the maxWidth right? 
But if there has to be a maxWidth can we pull it out into `constants.ts`?

enclave-manager/web/src/components/AppLayout.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type AppLayoutProps = PropsWithChildren<{
+  Nav: React.ReactElement;
+}>;
+
+export const AppLayout = ({ Nav, children }: AppLayoutProps) => {
+  return (
+    <Flex flexDirection={"row"}>
+      {Nav}
+      <Flex as="main" w={"100%"} justifyContent={"flex-start"} p={"20px 40px"} className={"app-container"}>
```
adschwartz: Just for my learning: what does the two p values (20, 40) mean?

enclave-manager/web/src/components/KurtosisThemeProvider.tsx
```diff
@@ -0,0 +1,205 @@
...
+    kurtosisGreen: {
+      100: "#005e11",
+      200: "#008c19",
+      300: "#00bb22",
+      400: "#00C223", // The true green
+      500: "#33ee55",
+      600: "#66f27f",
+      700: "#99f7aa",
+    },
+    kurtosisGray: {
```
adschwartz: Props for writing american english haha

enclave-manager/web/src/components/RelativeDateTime.tsx
```diff
@@ -0,0 +1,35 @@
+import { DateTime } from "luxon";
+import { Text, TextProps, Tooltip } from "@chakra-ui/react";
+import { isDefined } from "../utils";
+import { useEffect, useState } from "react";
+
+type RelativeDateTimeProps = TextProps & {
+  dateTime: DateTime | null;
+};
+
+export const RelativeDateTime = ({ dateTime, ...textProps }: RelativeDateTimeProps) => {
```
adschwartz: maybe consider using react-moment. I think it does the same?

https://www.npmjs.com/package/react-moment

enclave-manager/web/src/emui/enclaves/EnclaveList.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const kurtosisClient = useKurtosisClient();
+  const [enclaves, setEnclaves] = useState<EnclaveFullInfo[]>();
+  const [selectedEnclaves, setSelectedEnclaves] = useState<EnclaveFullInfo[]>([]);
+
+  useEffect(() => {
+    (async () => {
+      const enclavesResponse = await kurtosisClient.getEnclaves();
+      const enclaves = Object.values(enclavesResponse.enclaveInfo);
+      const [starlarkRuns, services, filesAndArtifacts] = await Promise.all([
+        Promise.all(enclaves.map((enclave) => kurtosisClient.getStarlarkRun(enclave))),
```
adschwartz: In the Cloud UI we've used `true-myth` for error handling. What do you think about adopting that for these API calls?

enclave-manager/web/src/emui/enclaves/Enclaves.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <Flex direction={"column"} gap={"36px"} width={"100%"}>
+        <EnclaveBreadcrumbs />
+        <Outlet />
+      </Flex>
+    ),
+    handle: { name: () => "Enclaves" },
+    children: [
+      { path: "/", element: <EnclaveList /> },
+      {
+        path: "/enclave/:enclaveName",
```
adschwartz: When the EM UI runs embedded using the cloud gateway URL, e.g. `cloud.kurtosis.com/gateway/ips/1-2-3-4/ports/9711` the path becomes prepended with all that junk so absolute paths become problematic. Basically the EM UI has to ignore whatever junk is prepended to the paths that it cares about. We have a setting enabled in the EM UI today that does that but it also creates issues. This is an important thing to work through to support the "embedded use case"

adschwartz (changes requested):
@Dartoxian Really good work! You did a lot so far and I love how clean your code is. Couple of comments and questions to look at. Great progress.

adschwartz:
One more thing: can you add a build script (put it under `scripts/build.sh` to be consistent with our normal practice) and perhaps add a section the readme on what needs to built to actually run this project (like the protobuf definitions). Or better yet, just add it to the build script so it's always generated.

enclave-manager/web/src/client/KurtosisClientContext.tsx
Dartoxian: I think it's better that this is in KurtosisClientProvider as it keeps responsibilities seperate.
* KurtosisClientProvider is responsible for `providing a KurtosisClient` - so is responsible for knowing about things like windows and browser events.
* The AuthenticatedKurtosisClient is responsible for `communicating with the cloud api` and to do so it requires a jwt token, which is provided to it in its constructor.

enclave-manager/web/src/client/KurtosisClientContext.tsx
Dartoxian: ah thanks - I wasn't really sure what this was for yet - is it a way to quicklink to a specific package for creating an enclave?

enclave-manager/web/src/client/KurtosisClientContext.tsx
Dartoxian: I don't think this is bleeding - this is the `KurtosisClientProvider` doing what it is responsible for - and this is the logic that switches between using a local or authenticated client.

enclave-manager/web/src/client/KurtosisClientContext.tsx
Dartoxian: Do you mean to encapsulate the jwt effect specifically? Something like:

```
type KurtosisJWTDetectorProps = {
  onJWTDetected: (jwt: string) => void;
};

export const KurtosisJWTDetector = ({ onJWTDetected }: KurtosisJWTDetectorProps) => {
  useEffect(() => {
    const receiveMessage = (event: MessageEvent) => {
      const message = event.data.message;
      switch (message) {
        case "jwtToken":
          const value = event.data.value;
          if (isDefined(value)) {
            onJWTDetected(value);
          }
          break;
      }
    };
    window.addEventListener("message", receiveMessage);
    return () => window.removeEventListener("message", receiveMessage);
  }, []);
  return <></>;
};
```

In my opinion this is excessive encapsulation - and `KurtosisClientProvider` would still have a state variable for storing the jwt token.

enclave-manager/web/src/client/constants.ts
Dartoxian: not sure what you mean - I've actually just used similar names to what's in the current engine/frontend - maybe it would be clearer if I just moved these constants into `LocalKurtosisClient`

enclave-manager/web/src/components/AppLayout.tsx
Dartoxian: just checked on slack what we are doing here. I agree making this a constant seems reasonable.

enclave-manager/web/src/components/AppLayout.tsx
Dartoxian: it's the same as setting `padding: 20px 40px` on the styles - https://chakra-ui.com/docs/styled-system/style-props

enclave-manager/web/src/components/RelativeDateTime.tsx
Dartoxian: interesting library - I've not come across it before. The main thing I like about luxon over moment is immutable objects.

enclave-manager/web/src/emui/enclaves/EnclaveList.tsx
Dartoxian: That pattern seems reasonable. Generally I like to swallow application breaking errors as soon as possible, rather than propagating the handling to components that shouldn't need to know the real details of working with the api. So my plan was going to be:
* wrap the `kurtosisClient` in `useKurtosisClient` with some sort of `ErrorHandlingKurtosisClient`
* ErrorHandlingKurtosisClient would delegate all requests to the actual client used
* ErrorHandlingKurtosisClient, being in the react tree, would be able to create toasts with the error messages

Perhaps a combination of these makes it possible to have components render their own error state, but keeps the location of api errors consistent on the page.

enclave-manager/web/src/emui/enclaves/Enclaves.tsx
Dartoxian: I think this will still be fine in this file, but point taken - I think it will have to be handled in App.tsx.

It would be great to go through how to test the embedded functionality on Thursday if possible?

enclave-manager/web/src/client/KurtosisClientContext.tsx
adschwartz: Got it, that makes sense. I wonder though if we can split out the jwt token part from this code and catch and provide it from some other place. The generic KurtosisClient shouldn't know about jwt tokens imo.

enclave-manager/web/src/client/KurtosisClientContext.tsx
adschwartz: Yes it is. You should be able to see the full code in `main` and track what it does.

enclave-manager/web/src/client/KurtosisClientContext.tsx
adschwartz: I see. I may be misunderstanding the design you had in mind here. Perhaps we should meet up and discuss this together so I can get on the same page. just let me know when you are around to chat.

enclave-manager/web/src/client/KurtosisClientContext.tsx
adschwartz: I think we can chat about these topics all together (referencing earlier comment) so I better understand your reasoning in this design.

enclave-manager/web/src/client/constants.ts
adschwartz: For reference, check out the `.env*` files in the cloud ui repo

enclave-manager/web/src/components/RelativeDateTime.tsx
adschwartz: That's neat. I think you can still use luxon here, it's just the formatting of the timestamp that you get for free with react-moment. I.e. you can skip the manual handling of state and timeouts etc.

enclave-manager/web/src/emui/enclaves/EnclaveList.tsx
adschwartz: I like the idea of the `ErrorHandlingKurtosisClient`. What I care about is that we have some explicit notion of handling errors and making consumers of the api calls aware if an error has happened so they can deal with it.

enclave-manager/web/src/client/KurtosisClientContext.tsx
Dartoxian: > The generic KurtosisClient shouldn't know about jwt tokens imo.

I agree - and it doesn't

enclave-manager/web/src/utils/index.ts
```diff
@@ -0,0 +1,54 @@
...
+};
+
+export async function asyncResult<T>(
+  p: Promise<T> | (() => Promise<T>),
+  errorMessage?: string,
+): Promise<Result<T, ErrorAndMessage<any>>> {
+  return (typeof p === "function" ? p() : p)
+    .then((r) => Result.ok<T, ErrorAndMessage<any>>(r))
+    .catch((err: any) => Result.err({ error: err, message: errorMessage }));
+}
```
Dartoxian: @adschwartz WDYT of using true-myth this way? it reduces the need to wrap in `try/catch` each method passed through the api, instead it's just wrapped in this method, which feels a little more concise.

Dartoxian:
Thanks for the extra comments and chatting last night @adschwartz . I've updated this PR with all of the feedback we discussed to change now.

I'm going to go through the design changes now in a fresh branch from this one.

adschwartz (approved):

PR #1610, opened 2023-10-24
feat: add full story script

adschwartz:
## Description:
Adding Full Story to our docs page. Full disclosure, I haven't been able to test if this actually works so a follow up PR may be needed

NO


docs/docusaurus.config.js
```diff
@@ -1,184 +1,190 @@
...
-        ],
-        copyright: `Copyright © ${new Date().getFullYear()} Kurtosis Technologies`,
-      },
-      prism: {
-        theme: lightCodeTheme,
-        darkTheme: darkCodeTheme,
-      },
-      algolia: {
-        appId: 'NTSX40VZB8',
+        ({
```
h4ck3rk3y: did we go from tabs to spaces?

docs/static/js/load-fullstory.js
```diff
@@ -0,0 +1,21 @@
+window['_fs_host'] = 'fullstory.com';
+window['_fs_script'] = 'edge.fullstory.com/s/fs.js';
+window['_fs_org'] = 'o-1RQ4W2-na1';
```
h4ck3rk3y, line 3: guessing this can be shared as this is analytics

h4ck3rk3y (approved):
don't have much context on how full story works! so trusting you on this :-)

docs/docusaurus.config.js
adschwartz: oh i'm just using the auto formatter in intellij. Is tabs our coding standard?

docs/static/js/load-fullstory.js
adschwartz: Yes public id.

docs/docusaurus.config.js
adschwartz: I'll fall back to that

PR #1612, opened 2023-10-24
feat: disable smooth scrolling for logs and default to select restart services

adschwartz:
## Description:
disable smooth scrolling for logs and default to select restart services

YES


tedim52 (approved):
LGTM

PR #1616, opened 2023-10-25
fix: name temporary python script for run_python with suitable name

h4ck3rk3y:
Earlier we would name it with a name like `tranquil-wind` etc and that would confuse the reader on what it might contain. The ideal solution here is to clean this up and not use a temporary files artifact at all and not use the service network inside interpret


core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_python.go
```diff
@@ -149,6 +151,9 @@ type RunPythonCapabilities struct {
 }
 
 func (builtin *RunPythonCapabilities) Interpret(_ string, arguments *builtin_argument.ArgumentValuesSet) (starlark.Value, *startosis_errors.InterpretationError) {
+	randomUuid := uuid.NewRandom()
+	builtin.name = fmt.Sprintf("task-%v", randomUuid.String())
```
leoporoli, line 155: you can put this "task-%v" in a constant

leoporoli (approved):

PR #1617, opened 2023-10-25
fix: handle error and fix rendering bug

### LINKED ISSUE #1587: The status spinner in EM UI doesn't update when script is fully completed
### What's your CLI version?

-

There's a bug where the spinner doesn't get replaced with the green checkmark. 

It looks like the backend is sending the right data.
I think this happens because the state doesn't get updated correctly. Maybe we are missing a dependency on another variable to trigger the render?

image

### Desired behavior

-

adschwartz:
## Description:
Added logic to handle error when script fails to run and returns an error. Before the status of the execution would just hang as being in progress, but now it updates the status to be an error (when the error is returned from the EM API).

As a followup I've created issue #1618 to provide better error reporting in the UI.

YES

Closes: https://github.com/kurtosis-tech/kurtosis/issues/1587


tedim52 (approved):
LGTM

PR #1642, opened 2023-10-27
fix: run_sh doesn't remove new lines from input

### LINKED ISSUE #1597: piping doesn't work for run_sh
### What's your CLI version?

0.84.9

`run = "cat /foo/bar.txt | tr -d '\n'"`

for a file that looks like

```
foo
```

still keeps the newline

### Desired behavior

the new line gets dropped due to the pipe

h4ck3rk3y:
## Description:
Earlier `cat foo.txt | tr -d '\n'` wouldn't work as it would run as `cat foo.txt | tr -d ' '`

### DIFF
--- a/core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_sh.go
+++ b/core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_sh.go
@@ -21,7 +21,6 @@ import (
 	"github.com/kurtosis-tech/stacktrace"
 	"github.com/xtgo/uuid"
 	"go.starlark.net/starlark"
-	"strings"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_sh.go
```diff
@@ -253,7 +252,6 @@ func getCommandToRun(builtin *RunShCapabilities) (string, error) {
 	if err != nil {
 		return "", stacktrace.Propagate(err, "An error occurred while replacing runtime values in run_sh")
 	}
-	commandWithNoNewLines := strings.ReplaceAll(maybeSubCommandWithRuntimeValues, newlineChar, " ")
```
h4ck3rk3y, line 256: i don't know why we did this though?

h4ck3rk3y: ah we have a failing test for this

h4ck3rk3y: My theory is that while writing the tests someone saw that `echo kurtosis` would echo

```
kurtosis

```

and that messed up the first test so they changed it to remove new lines

h4ck3rk3y: @leoporoli fysa

leoporoli (approved):

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_sh.go
leoporoli: @h4ck3rk3y you conclusion makes sense

PR #1649, opened 2023-10-30
feat: warn users of diff architecture image running

### LINKED ISSUE #1509: issue a warning if an image of a different platform is running
### Background & motivation

not all images publish on all architectures and sometimes there are failures and it isn't immediately clear that the container wasn't starting due to architecutre issues - not saying arm64 always fails on amd64 but sometimes it does; knowing that Kurtosis tried running a different platform than expected could help people debug

@galenmarchetti ran into this

### Desired behaviour

We tell the user that we are attempting to start an arm64 image on an amd64 device and vice versa

h4ck3rk3y:
## Description:
We now warn users in the beginning if the image is different from whats expected

YES/NO


h4ck3rk3y:
Screenshot 2023-10-31 at 10 53 41 AM

Its basically this but I don't log the architecture in the initial blue message any more - that was just for test

container-engine-lib/lib/backend_interface/kurtosis_backend.go
```diff
@@ -31,7 +31,8 @@ type KurtosisBackend interface {
 	// FetchImage always attempts to retrieve the latest image.
 	// If retrieving the latest [dockerImage] fails, the local image will be used.
 	// Returns True is it was retrieved from cloud or False if it's a local image
-	FetchImage(ctx context.Context, image string, downloadMode image_download_mode.ImageDownloadMode) (bool, error)
+	// Returns a string that represents the architecture of the image
```
leoporoli, line 34: why did you remove this comment: ` Returns True if it was retrieved from cloud or False if it's a local image` ?

leoporoli (approved):

container-engine-lib/lib/backend_interface/kurtosis_backend.go
h4ck3rk3y: I havne't remvoed it; its still tehre!

PR #1659, opened 2023-10-30
fix: block local absolute locators

### LINKED ISSUE #1637: Absolute within-package imports should be rejected, but aren't
### What's your CLI version?

0.84.11

The imports in this directory should fail, but don't because `GitPackagecontentProvider#isSamePackageLocalAbsoluteLocator` is simply using `strings.HasPrefix`

This also means these unit tests I just wrote, which should pass, fail:

```go
func Test_isSamePackageLocalAbsoluteLocator_TestDetectionInSubpath(t *testing.T) {
	result := isSamePackageLocalAbsoluteLocator("github.com/author/package/bang/lib.star", "github.com/author/package/main.star")
	require.True(t, result)
}

func Test_isSamePackageLocalAbsoluteLocator_TestDetectionInDifferentSubdirectories(t *testing.T) {
	result := isSamePackageLocalAbsoluteLocator("github.com/author/package/subdir1/file1.star", "github.com/author/package/subdir2/file2.star")
	require.True(t, result)
}
```

I believe the deeper problem is that `GetAbsoluteLocatorForRelativeLocator` doesn't actually have a notion of the package ID (which could be `github.com/author/package`, or nested like `github.com/author/somedir/otherdir/package`), so it has no way of telling if two paths are actually within the same package.

What seems to have happened is that when `GetAbsoluteLocatorForRelativeLocator` was written, the first argument was `packageId`, which would never end with `XXXX.star` (as evidenced by `PackageContentProvider#GetAbsoluteLocatorForRelativeLocator` having `packageId` as its first argument name), but then it got changed/overloaded to become module locator, which does end with `XXXX.star` (as evidenced by [GitPackageContentProvider#GetAbsoluteLocatorForRelativeLocator's first argument being called `parentModuleId`).

Here is a package that reproduces both issues:

[not-failing-imports.zip](https://github.com/kurtosis-tech/kurtosis/files/13183931/not-failing-imports.zip)

### Desired behavior

- The unit tests I wrote above should be incorporated
- We should actually reject absolute local locators, in `import_module`, `upload_files`, and anywhere else

This probably means that either a) the "reject absolute local locators" validation needs to be moved somewhere else or b) `GetAbsoluteLocatorForRelativeLocator` needs to know what package it's being executed inside of

leoporoli:
## Description:
block local absolute locators was not working appropriately 

YES

Fix #1637


core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
```diff
@@ -199,119 +200,154 @@ func TestGetAbsolutePathOnDisk_WorksForNonInMainBranchLocators(t *testing.T) {
...
+
+	packageId := "github.com/main-project/package1-in-subfolder"
+	locatorOfModuleInWhichThisBuiltInIsBeingCalled := "github.com/main-project/package1-in-subfolder/main.star"
+	maybeRelativeLocator := "github.com/main-project/package2-in-subfolder/file.star"
+
+	_, err := provider.GetAbsoluteLocator(packageId, locatorOfModuleInWhichThisBuiltInIsBeingCalled, maybeRelativeLocator, noPackageReplaceOptions)
+	require.Nil(t, err)
+}
+
+func TestGetAbsoluteLocator_AbsoluteLocatorIsInRootPackageButSourceIsNotShouldNotBeBlocked(t *testing.T) {
```
leoporoli, line 353: @h4ck3rk3y with this test I tried to simulate Carlo's case. It was like this example?

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
```diff
@@ -199,119 +200,175 @@ func TestGetAbsolutePathOnDisk_WorksForNonInMainBranchLocators(t *testing.T) {
...
+	provider := NewGitPackageContentProvider("", "", nil)
+
+	packageId := "github.com/root-package"
+	locatorOfModuleInWhichThisBuiltInIsBeingCalled := "github.com/child-package/main.star"
+	maybeRelativeLocator := "github.com/main-package/file.star"
```
h4ck3rk3y, line 358: this should read `github.com/root-pacakge/file.star`

can we add this as a in integration test?

h4ck3rk3y (approved):

PR #1660, opened 2023-10-30
fix: Improve description of the image-download flag

lostbean:
## Description:
 Improve description of the image-download flag in both CLI help and docs.

YES

- fix on #1495 for #1299


cli/cli/commands/run/run.go
```diff
@@ -200,7 +200,7 @@ var StarlarkRunCmd = &engine_consuming_kurtosis_command.EngineConsumingKurtosisC
 		},
 		{
 			Key:     imageDownloadFlagKey,
-			Usage:   "If unset, it defaults to `missing` for fetching the latest image only if not available in local cache. Use `always` to always fetch the latest image.",
+			Usage:   "If unset, it defaults to `missing` to fetch the image if no image with the latest tag is available in local cache. Use `always` to always fetch the latest image.",
```
leeederek:
```suggestion
			Usage:   "If unset, it defaults to `missing` which will only download the latest image tag if the image does not already exist locally (irrespective of the tag of the locally cached image). Use `always` to have Kurtosis always check and download the latest image tag, even if the image exists locally.",
```

docs/docs/cli-reference/run.md
```diff
@@ -126,7 +126,7 @@ run(plan, some_parameter = struct(some_property = "Property value"))
 
 1. The `--no-connect` flag can be used to disable user services port forwarding (default behavior is to forward the ports)
 
-1. The `--image-download` flag can be used to determine what is the behavior to fetch images tagged as latest during the package deployment. Use `missing` to fetch the latest image only if not available in local cache and use `always` to always fetch the latest image from remote repository (default behavior is `missing`).
+1. The `--image-download` flag can be used to determine what is the behavior to fetch images tagged as latest during the package deployment. Use `missing` to fetch the latest image only if no other version with the latest tag is available in local cache and use `always` to always fetch the latest image from remote repository (default behavior is `missing`).
```
leeederek:
```suggestion
1. The `--image-download` flag can be used to configure the download behavior for a given run. When set to `missing`, Kurtosis will only download the latest image tag if the image does not already exist locally (irrespective of the tag of the locally cached image). When set to `always`, Kurtosis will always check and download the latest image tag, even if the image exists locally.
```

leeederek (approved):
Proposed some suggestions - feel free to merge

PR #1662, opened 2023-10-31
build: build the right architecture locally

h4ck3rk3y:
## Description:
We would build amd64 even on our arm64 M1/M2 machines; this fixes that by building the local architecture by default for images and engine/core/files artifact expander servers

NO

### DIFF
--- a/scripts/docker-image-builder.sh
+++ b/scripts/docker-image-builder.sh
@@ -36,7 +36,6 @@ if [ -z "${image_tags}" ]; then
     show_helptext_and_exit
 fi
 
-# Get the system architecture using uname
 uname_arch=$(uname -m)
 
 architecture="amd64"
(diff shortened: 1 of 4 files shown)


scripts/docker-image-builder.sh
```diff
@@ -36,12 +36,22 @@ if [ -z "${image_tags}" ]; then
     show_helptext_and_exit
 fi
 
+uname_arch=$(uname -m)
+
+architecture="amd64"
```
leoporoli, line 41: in the other scripts, you used `unknown` as the default architecture. Why are you using "amd64" here instead?

leoporoli: oh, forget it I've just realized that the other files call down this one.

leoporoli (approved):

scripts/docker-image-builder.sh
h4ck3rk3y: also in this one the default was amd64; the others have mechanism to error if its unknown

PR #1663, opened 2023-10-31
ci: allow pull requests labelled with request-cf-deploy to deploy GitHub pages

h4ck3rk3y:
## Description:
Earlier pull requests created via forks would get blocked as the pages action wouldn't run. This is a work around for it following - https://github.com/cloudflare/pages-action/issues/94

### DIFF
--- a/.github/workflows/push-docs.yml
+++ b/.github/workflows/push-docs.yml
@@ -3,6 +3,12 @@ name: Deploy documentation
 on:
   push:
   merge_group:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


omar711 (approved):
To test my understanding: users making a PR from a fork would need to add a label to their PR, `request-cf-deploy`, to trigger a docs build?

h4ck3rk3y:
@omar711 They can't add the label as they don't have `write` but we do. After that gets added the job should run is my understanding.

PR #1665, opened 2023-10-31
refactor: move user constants to container-engine-lib

leoporoli:
## Description:
move user constants to container-engine-lib to make it accessible to the other modules like the `apic server` 

NO


h4ck3rk3y (approved):
@leoporoli I am on the fence with this - I am wondering if this should be a module of its own if its going to be used across. We have had this discussion in the past where felt like the container-engine-lib is becoming a dumping ground

Reasons for making this its own pacakge
1. Doesn't relate to Docker/Kubernetes at all

Reasons to not do this
1. We have a bunch of shared tiny modules - maybe thats okay?


@h4ck3rk3y your question is valid, I also had the same question and I decided not to create a new module just for these:
- I didn't have a name for the module, I mean a concept for describing what should be there.
- It's only one file, and I don't think is worth to create a module just for 1 file.

I know that this is not related to Docker/Kubernetes, and that is the downside of this, but I thought that it was a good place because `container-engine-lib` contains the Kurtosis primitives

if you have a good name or concept for the new package (describing what should we put there from now) I'm not against doing the change.

PR #1675, opened 2023-11-02
feat: add timestamps to log lines

tedim52:
## Description:
Attaches timestamps to all log lines to indicate when "kurtosis" logging infrastructure received the log line.

NO


engine/server/engine/centralized_logs/client_implementations/kurtosis_backend/kurtosis_backend_logs_database_client.go
```diff
@@ -192,7 +193,8 @@ func streamServiceLogLines(
 				return
 			}
 
-			logLine := logline.NewLogLine(logLineStr)
+			// TODO: Parse out the timestamp from docker container logs string instead of time.Now
```
adschwartz: @tedim52 I personally think what you have now is better, meaning: what you are logging is the timestamp we received the log whereas the timestamp from docker is when docker "managed" the log entry. I think it makes sense that we keep those two separate and it's fine to leave the docker timestamp embedded in the log line. Longer term I think we should move to structured logging so that we have a json entry that contains potentially many timestamps (for each actor that is involved in the parsing of the log line). With structured logging the user can then decide which timestamp is considered the "truth" (is it kurtosis timestamp, docker timestamp, service timestamp, etc.).

engine/server/engine/centralized_logs/client_implementations/kurtosis_backend/kurtosis_backend_logs_database_client.go
```diff
@@ -192,7 +193,8 @@ func streamServiceLogLines(
...
+			logLine := logline.NewLogLine(logLineStr, time.Now())
```
adschwartz, line 196: if I was being nitpicking, I would say we should move time.Now() up to being as close to when the line is received. It should be the first thing we do once the new line is ready. In this case it probably won't change the timestamp much but if you had more time consuming logic above this particular line, the timestamp would be later than when we actually received it.

adschwartz (approved):

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy_test.go
```diff
@@ -446,3 +452,37 @@ func TestGetJsonLogStringWithEOFAndNoValidJsonEnding(t *testing.T) {
...
+	require.NoError(t, err)
+
+	actualTime, err := parseTimestampFromJsonLogLine(jsonLogLine)
+
+	require.NoError(t, err)
+	require.Equal(t, expectedTime, *actualTime)
+}
+
+func TestParseTimestampFromJsonLogLineWithIncorrectlyFormattedTimeReturnsError(t *testing.T) {
+	timestampStr := "2023-09-06" // not UTC formatted timestamp str
```
adschwartz, line 487: This is a good test. Can we add another one to capture if the timezone information is absent like this:
`2023-09-06T00:35:15`

tedim52: good idea, will do

engine/server/engine/centralized_logs/client_implementations/kurtosis_backend/kurtosis_backend_logs_database_client.go
tedim52: Completely agree with ya. FWIW - right now we're not using the `KurtosisBackendLogsDatabaseClient`, but I wanted to implement timestamps for it to keep parity across the implementations of `LogsDatabaseClient`s

engine/server/engine/centralized_logs/client_implementations/kurtosis_backend/kurtosis_backend_logs_database_client.go
tedim52: Gotcha, this makes sense. I'll remove the TODO then.

engine/server/engine/centralized_logs/client_implementations/kurtosis_backend/kurtosis_backend_logs_database_client.go
tedim52: For the `PersistentVolumeLogsDatabaseClient` (the implementation we use currently), we use the time that the `LogsAggregator` claims to have seen the logs, which semantically I'm considering as the equivalent of when Kurtosis sees the log.

Issue #1651, opened 2023-10-30
Add missing flags to `kurtosis run --help`


barnabasbusa:
### Where did you see the issue in our documentation?

`kurtosis run --help`

### Issue description

When you run kurtosis run --help, you should see all the possible parameters that you can pass.

A few missing features that I have failed to see are: 
*`--args-file` flag now takes a URL. 
* `--always-pull` flag is missing.
* I'm sure there are some other flags that are now implemented but are missing from the `--help` window.

leeederek:
Hey @barnabasbusa - what version of the CLI are you on? I see both of the flags when I run `kurtosis run --help`:
```bash
Flags:
      --args-file string            The file (JSON/YAML) will be used as arguments passed to the Kurtosis Package
      --image-download missing      If unset, it defaults to missing for fetching the latest image only if not available in local cache. Use `always` to always fetch the latest image. (default "missing")
```

I believe you are on an older version of the CLI so thats why you might not see it!

leeederek:
Closing this out as completed! Just make sure you're on the latest and greatest Kurtosis :)

PR #1676, opened 2023-11-02
fix: mention URL JSON path in kurtosis run -h

leeederek:
## Description:
I noticed that `kurtosis run -h` does not mention the ability to support a URL for the --args-file. This PR adds a few words to highlight this support

YES

https://github.com/kurtosis-tech/kurtosis/issues/1651


cli/cli/commands/run/run.go
```diff
@@ -194,7 +194,7 @@ var StarlarkRunCmd = &engine_consuming_kurtosis_command.EngineConsumingKurtosisC
 		},
 		{
 			Key:     packageArgsFileFlagKey,
-			Usage:   "The file (JSON/YAML) will be used as arguments passed to the Kurtosis Package",
+			Usage:   "The file (JSON/YAML) that will be used to pass in arguments to the Kurtosis package. Can be a URL or file path to a file.",
```
h4ck3rk3y:
```suggestion
			Usage:   "The file (JSON/YAML) that will be used to pass in arguments to the Kurtosis package. Can be a URL or file path.",
```

h4ck3rk3y (approved):

PR #1677, opened 2023-11-02
build: move metrics library to monorepo and use it

h4ck3rk3y:
We move the metrics library to monorepo so that its easy to develop with


leoporoli (approved):
I didn't check the metrics-library code because I'm assuming that it was just moved from one place to another and there is no change in the base code.

PR #1687, opened 2023-11-03
feat: print Made with Kurtosis at the end of a run

h4ck3rk3y:
This is an attempt to solve the people wrap around us problem


h4ck3rk3y:
image

leeederek (approved):

PR #1689, opened 2023-11-03
feat: manage parameters and URL

adschwartz:
## Description:
Changes:
* Accessing the package config from anywhere in the UI
* Integration with URL parameters in the EM UI and the Cloud UI

YES


enclave-manager/web/src/client/enclaveManager/KurtosisClient.ts
```diff
@@ -73,11 +79,11 @@ export abstract class KurtosisClient {
       const apicInfo = enclave.apiContainerInfo;
       assertDefined(
         apicInfo,
-        `Cannot listFilesArtifactNamesAndUuids because the passed enclave '${enclave.name}' does not have apicInfo`,
+        `Cannot listFilesArtifactNamesAndUuids because the passed enclave '${enclave.name}' does not have apicInfo`
       );
       const request = new GetListFilesArtifactNamesAndUuidsRequest({
         apicIpAddress: apicInfo.bridgeIpAddress,
-        apicPort: apicInfo.grpcPortInsideEnclave,
+        apicPort: apicInfo.grpcPortInsideEnclave
```
Dartoxian: wonder if our prettiers aren't configured consistently - this change is just a formatting change. Could you run `yarn prettier:fix` and verify whether you keep this change.

enclave-manager/web/src/components/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const KURTOSIS_PACKAGE_PARAMS_URL_ARG = "package-args"
+export const KURTOSIS_PACKAGE_ID_URL_ARG = "package-id"
+export const KURTOSIS_CREATE_ENCLAVE_URL_ARG = "create-enclave"
+
+export const KURTOSIS_PACKAGE_URL_BASE="/"
```
Dartoxian: I'd prefer to have constants defined near to where they are used (ie components/mo- arguably having a `constants` file in a `components/enclaves/modals`) directory is a bad smell anyway - it suggests we've started working quite a lot of application logic into the components (which tbf is what I've given you...)

enclave-manager/web/src/components/enclaves/CreateEnclave.tsx
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+  return (
+    <>
+      <PreloadPackage onPackageLoaded={handleOnPackageLoaded} />
+      <ManualCreateEnclaveModal
+        isOpen={manualCreateEnclaveOpen}
+        onClose={handleCloseManualCreateEnclave}
+        onConfirm={handleManualCreateEnclaveConfirmed}
+      />
```
Dartoxian, line 47: took me a moment to work out how you're getting this open now, but makes sense. In a non SPA I think I'd not be that comfortable driving state changes like this, but for this application where we're relying a lot on react-router-dom to manage state changes this does seem more appropriate.

enclave-manager/web/src/components/enclaves/configuration/EnclaveConfigurationForm.tsx
```diff
@@ -98,7 +98,7 @@ export const EnclaveConfigurationForm = forwardRef<
         {},
       );
 
-    onSubmit({ ...data, args: newArgs });
+    onSubmit({enclaveName: "", restartServices: false, ...data, args: newArgs });
```
Dartoxian: I am not sure this is needed - these fields should already be supplied by `ConfigureEnclaveModal` (ie they should never be `null` or undefined in the supplied `data` object)

Dartoxian (commented):
Main points:
* please check prettier - if you want to change the rules could you also commit them. Consistent code formatters reduces noise in code review.
* Given you've gone part way to adding the build change could you also .gitignore the engine output?

enclave-manager/web/src/components/CopyButton.tsx
```diff
@@ -15,7 +15,7 @@ export const CopyButton = ({ valueToCopy, text, ...buttonProps }: CopyButtonProp
       const v = typeof valueToCopy === "string" ? valueToCopy : valueToCopy();
       navigator.clipboard.writeText(v);
       toast({
-        title: `Copied '${v}' to the clipboard`,
+        title: `Copied url to the clipboard`,
```
Dartoxian: It's not always a url that is copied - I think this is better exposed as a prop.

enclave-manager/web/src/components/enclaves/configuration/EnclaveConfigurationForm.tsx
adschwartz: yeah this was a fix to get it to compile. Wondering what data we are missing to wire in here.

enclave-manager/web/src/client/enclaveManager/KurtosisClient.ts
adschwartz: indeed, my client is not picking up the `prettier` formatter. Will fix that

adschwartz: We should add an ignore for protobuf generated files as it also formats those guys.

adschwartz: done

enclave-manager/web/src/components/enclaves/CreateEnclave.tsx
adschwartz: I'm using the url fragment to trigger it to open from any "page". 

Would be (genuinely) interested to hear what you'd propose alternatively and what specifically would not make this a good fit in a not SPA?

enclave-manager/web/src/components/enclaves/configuration/EnclaveConfigurationForm.tsx
adschwartz: I can't get this to compile if I don't add those members, is there something off in my config? How are you getting this to compile? It makes sense that the descruturing should provide this data but running yarn build fails on mine, so wondering what you have that works?
image

enclave-manager/web/src/components/CopyButton.tsx
adschwartz: Done

enclave-manager/web/src/components/enclaves/CreateEnclave.tsx
Dartoxian: One alternative would be to drive it entirely from react state/structure, which is what I was doing before (and is perhaps slightly more readable as it's explicit how it's used as you read the react tree) - but this does seem preferable in terms of getting the behaviour easily anywhere in the app.

The 'not a SPA' example I am thinking about is anything that uses a server rendering framework (specifically thinking nextjs), in those frameworks using effects to render content is generally avoided - and the page is hydrated from the server with relevant data if possible - so we'd probably prefer to use a configuration page, rather than a modal.

enclave-manager/web/src/components/enclaves/configuration/EnclaveConfigurationForm.tsx
Dartoxian: interesting, no I don't get this same issue when I run yarn run build.

In the interest of expediency, I'm not particularly interested in working out what the difference is atm.

enclave-manager/web/src/components/enclaves/CreateEnclave.tsx
adschwartz: Cool! This all makes sense. Thanks!

tedim52 (approved):

enclave-manager/web/src/client/enclaveManager/KurtosisClientContext.tsx
```diff
@@ -131,3 +138,12 @@ export const useKurtosisClient = (): KurtosisClient => {
 
   return client;
 };
+
+const paramToUrl = (searchParams: URLSearchParams, param: string) => {
+  let paramString = searchParams.get(param);
+  if (paramString === null) return null;
```
Dartoxian: I'll add eslint config for `curly` -https://eslint.org/docs/latest/rules/curly. Generally I dislike this form of conditional as it's less readable and a source of developer errors. Famous example of why not to do this https://www.synopsys.com/blogs/software-security/understanding-apple-goto-fail-vulnerability-2.html

scripts/docker-image-builder.sh
```diff
@@ -40,11 +40,11 @@ uname_arch=$(uname -m)
...
-#if [ "$uname_arch" == "x86_64" ] || [ "$uname_arch" == "amd64" ]; then
-#    architecture="amd64"
-#elif [ "$uname_arch" == "aarch64" ] || [ "$uname_arch" == "arm64" ]; then
-#    architecture="arm64"
-#fi
+if [ "$uname_arch" == "x86_64" ] || [ "$uname_arch" == "amd64" ]; then
+    architecture="amd64"
+elif [ "$uname_arch" == "aarch64" ] || [ "$uname_arch" == "arm64" ]; then
+    architecture="arm64"
+fi
```
Dartoxian, line 47: what is this change for?

enclave-manager/web/src/client/enclaveManager/KurtosisClient.ts
```diff
@@ -5,31 +5,51 @@ import {
...
+   * */
+  protected readonly parentUrl?: URL;
+
+  /* Full URL of the EM UI, covering two use cases:
+   * In local-mode this is the same as the `parentUrl`
+   * In authenticated mode : https://cloud.kurtosis.com/enclave-manager/gateway/ips/1-2-3-4/ports/1234/?searchparams... (this data/url is provided as a search param when the code loads)
+   *
+   * This URL is primarily used to set the react router basename so that the router is able to ignore any leading subdirectories.
+   * */
+  protected readonly childUrl?: URL;
```
Dartoxian, line 37: nbd, but knowing `childUrl` and `parentUrl` is not the responsibility of the KurtosisClient (which should only be concerned with talking to an enclave manager backend).

It would be more appropriate to just store these values in state inside the KurtosisClientContextProvider, and have a way of accessing them from there.

Dartoxian (commented):
I think prettier:fix also needs to be run over this change.

Dartoxian:
> I think prettier:fix also needs to be run over this change.

done - and added curly

enclave-manager/web/src/client/enclaveManager/KurtosisClientContext.tsx
adschwartz: That's fair. Thanks for fixing that!

PR #1697, opened 2023-11-06
fix: correct is_ci value in metrics from APIC

h4ck3rk3y:
## Description:
Earlier metrics from the APIC wouldn't get the right value for is_ci as they would look for the CI environment(TF_BUILD, CIRCLE_CI...) variables which our docker container wouldn't contain;

Now

1. CLI figures out if we are running in CI
2. CLI passes the information down to Engine
3. Engine Passes the information down to APIC

So if someone is running in Circle/GitHub actions - wherever the `kurtosis engine restart` happens we get the valid IS_CI value for that CLI


h4ck3rk3y:
https://github.com/kurtosis-tech/kurtosis/assets/6670312/d477729e-86c5-4819-a6a9-1e5399ab3bad

@leoporoli demo of the feature that reads from env variables on where engine restart is run

leoporoli (approved):

PR #1702, opened 2023-11-07
feat: new emui service logs

Dartoxian:
## Description:
This change adds support for showing enclave service logs in the new enclave manager ui.

Additionally, this pr contains a partial implementation of the service view ui - full implementation to follow.

### Demo:

In this demo `github.com/kurtosis-tech/log-load-package` is used to demonstrate large quantities of logs don't seem to break following. Additionally, my browser has it's locale set to san francisco - so that browser locale dependent timestamps can be demonstrated.

https://github.com/kurtosis-tech/kurtosis/assets/4419574/533b5f5a-7fdf-40b1-a6c5-e0b67970811c

YES

### DIFF
--- a/enclave-manager/web/src/client/enclaveManager/LocalKurtosisClient.ts
+++ b/enclave-manager/web/src/client/enclaveManager/LocalKurtosisClient.ts
@@ -6,7 +6,7 @@ import { KurtosisClient } from "./KurtosisClient";
 
 export class LocalKurtosisClient extends KurtosisClient {
   constructor() {
-    const defaultUrl = new URL(window.location.href);
+    const defaultUrl = new URL(`${window.location.protocol}//${window.location.host}`);
     super(
       createPromiseClient(
         KurtosisEnclaveManagerServer,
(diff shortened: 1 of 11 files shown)


enclave-manager/web/src/client/enclaveManager/LocalKurtosisClient.ts
```diff
@@ -6,7 +6,7 @@ import { KurtosisClient } from "./KurtosisClient";
 
 export class LocalKurtosisClient extends KurtosisClient {
   constructor() {
-    const defaultUrl = new URL(window.location.href);
+    const defaultUrl = new URL(`${window.location.protocol}//${window.location.host}`);
```
adschwartz, line 9: @Dartoxian is this actually correct?
1) We need the full url with params (and fragments?) so we can read the package id and parameters etc.
2) I think this url is missing the colon? (moot point given `1)` ).

Dartoxian: 1. I think when running against localhost at least we don't want that. I made this change mainly for my local development - I was finding that on every reload I was served the enclave list, rather than the correctly routed content (and links were broken because of this)
2. I agree this looks weird, but:
image

adschwartz (approved):

Issue #1619, opened 2023-10-25
fix the apic / cli / sdk metrics story


h4ck3rk3y:
### What's your CLI version?

0.84.11

### Description & steps to reproduce

The `Run` event is inside APIC but will get the wrong IS_CI value; in our dashboards we are using ENCLAVE_CREATE which only comes from the CLI for now

### Desired behavior

Run should have the right IS_CI vlaue

### What is the severity of this bug?

Painful; this is causing significant friction in my workflow.

### What area of the product does this pertain to?

Other: anything not covered by the above

h4ck3rk3y:
After #1703

- move metrics to the right places #1718 
- make kurtosis-cloud-config just work
- notify @galenmarchetti  about the new changes and cloud_user_id being the new correct thing instead of cloud-user-id

h4ck3rk3y:
cloud config just works!

h4ck3rk3y:
This is done as of 0.85.9

PR #1703, opened 2023-11-07
fix: pass cloud user id and cloud instance id while creating metrics client

h4ck3rk3y:
## Description:
Earlier we would pass it per event (only one event supported) - now we get it for every event

Part of #1619


api/golang/core/lib/starlark_run_config/starlark_run_config.go
```diff
@@ -85,12 +85,14 @@ func WithExperimentalFeatureFlags(experimentalFeatureFlags []kurtosis_core_rpc_a
 	}
 }
 
+// Deprecated: This isn't used anymore and the value passed will not be read anywhere
 func WithCloudInstanceId(cloudInstanceId string) starlarkRunConfigOption {
```
leoporoli, line 89: are you tracking the remotion in a ticket?

core/launcher/args/api_container_args.go
```diff
@@ -47,8 +48,16 @@ type APIContainerArgs struct {
...
 	IsCI bool `json:"is_ci"`
+
+	// The Cloud User ID of the current user if available
+	CloudUserID metrics_client.CloudUserID `json:"cloud_user_id"`
+
+	// The Cloud Instance ID of the current user if available
+	CloudInstanceID metrics_client.CloudInstanceID `json:"cloud_instance_id"`
 }
 
+var skipValidation = []string{"cloud_user_id", "cloud_instance_id"}
```
leoporoli: nit: you could use a set (map[string]bool) instead and avoid having to iterate the slice

engine/launcher/args/args.go
```diff
@@ -50,8 +51,16 @@ type EngineServerArgs struct {
...
 	IsCI bool `json:"is_ci"`
+
+	// The Cloud User ID of the current user if available
+	CloudUserID metrics_client.CloudUserID `json:"cloud_user_id"`
+
+	// The Cloud Instance ID of the current user if available
+	CloudInstanceID metrics_client.CloudInstanceID `json:"cloud_instance_id"`
 }
 
+var skipValidation = []string{"cloud_user_id", "cloud_instance_id"}
```
leoporoli: same here, you could use a set instead

engine/launcher/args/args.go
```diff
@@ -136,3 +153,12 @@ func (args EngineServerArgs) validate() error {
 	}
 	return nil
 }
+
+func contains(slice []string, value string) bool {
```
leoporoli: you can remove this helper function if you use a set instead

metrics-library/golang/lib/metrics_client/metrics_client_types.go
```diff
@@ -7,3 +7,6 @@ const (
 )
 
 type MetricsClientType string
+
+type CloudUserID string
```
leoporoli, line 11: nice!

leoporoli (approved):

engine/launcher/args/args.go
h4ck3rk3y: thanks! this makes sense :)

PR #1709, opened 2023-11-07
fix: new emui correct handling of Bool input type

Dartoxian:
## Description:
This PR corrects the enclave configuration form input to use boolean if a package argument type is boolean - which is enum value 0 and was previously treated as falsy.

### Screenshot of fix
image

YES


adschwartz:
@Dartoxian actually the default (in the absence of type set in the data coming from the API) should be `json`.

adschwartz (approved):

PR #1712, opened 2023-11-07
fix: stop all services first to update the service status in the service registration during a `kurtosis enclave stop` execution

### LINKED ISSUE #1711: The stop enclave CLI command is not updating the service status in the service registration repository
### What's your CLI version?

0.85.1

The `kurtosis enclave stop {enclave-identifier}` kill all the container in the Docker backend but this workflow doesn't unregister the user services in the user service registration repository

### Desired behavior

It should update the service status to "stop" in the service registration repository because otherwise we will have conflicts if we want to restart the services

leoporoli:
## Description:
unregister the enclave services during a `kurtosis enclave stop` execution

NO

Fix #1711


h4ck3rk3y (approved):
sounds alright - this puts the logic inside the CLI though; is there a world where we change `StopEnclave` inside engine to stop services first instead?

that way typescript/curl etc get it for free

cli/cli/commands/enclave/stop/stop.go
```diff
@@ -4,11 +4,13 @@ import (
 	"context"
 	"fmt"
 	"github.com/kurtosis-tech/kurtosis/api/golang/engine/kurtosis_engine_rpc_api_bindings"
+	"github.com/kurtosis-tech/kurtosis/api/golang/engine/lib/kurtosis_context"
```
h4ck3rk3y, line 7: leaving a comment here for discussion to prevent auto merge

wonder whether StopEnclave inside Engine Server can do a service stop so that ts/curl/any other frontend gets this functionality for free

leoporoli: As we talked in Slack the problem is that there is not connection between the Engine and the APIC server, so doesn't allow us to do it in the engine side

PR #1717, opened 2023-11-08
feat: generate enclave manager ui in build process and check prettier

Dartoxian:
## Description:
This pr includes the new enclave manager in the build process - specifically in the `engine` build process. This removes the need to commit the built emui to the codebase.

Additionally `yarn prettier` is added as a build step to the emui, which will fail the build if prettier has not been run (as is currently the case in main).

This change also includes a fix to all linting warnings, so that there are no warnings in the production build.

NO

* devconnect burndown


.circleci/config.yml
```diff
@@ -416,6 +446,11 @@ jobs:
 
       - <<: *abort_job_if_only_docs_changes
 
+      - attach_workspace:
+          at: "<< pipeline.parameters.workspace-with-cli-binary-and-images-mountpoint >>"
+
+      - run: cp -R << pipeline.parameters.workspace-with-cli-binary-and-images-mountpoint >>/webapp engine/server/webapp
```
adschwartz, line 452: @Dartoxian does this not happen as part of the `postbuild` step in `enclave-manager/scripts/build.sh` => `yarn build`?

adschwartz (changes requested):
@Dartoxian this looks good. I only have a single comment/question. We can try running it after your review.

.circleci/config.yml
Dartoxian: this step is what copies the build output between jobs:
* First the `build_enclave_manager_webapp` job will run, and save the build output into the circle ci workspace
* Second the `build_engine_server` job will run, this job reattaches the workspace and copies the build output into place  (this is what is happening here) before running it's own build steps

I chose to have a job for building emui separate from `build_engine_server` because the details of building (or testing) the emui are not relevant to the engine server, only the output

adschwartz: Cool. I _think_ this should work

Dartoxian: :pray:

adschwartz (approved):

PR #1718, opened 2023-11-08
refactor: move some metrics from CLI to engine/apic

h4ck3rk3y:
- CreateEnclave
- StopEnclave
- DestroyEnclave
- RunFinished

These metrics were previously tracked from the CLI and hence would miss cloud runs. Moved it to APIC/Engine so they are always tracked.


h4ck3rk3y:
Part of #1619

engine/server/engine/main.go
```diff
@@ -56,13 +58,20 @@ const (
...
+
+	shouldFlushMetricsClientQueueOnEachEvent = false
 )
 
 // Nil indicates that the KurtosisBackend should not operate in API container mode, which is appropriate here
 //
 //	because this isn't the API container
 var apiContainerModeArgsForKurtosisBackend *backend_creator.APIContainerModeArgs = nil
 
+type doNothingMetricsClientCallback struct{}
```
leoporoli: I think this should be provided by the metrics client library. can it?

leoporoli (approved):

PR #1728, opened 2023-11-09
feat: improve logging part 1

adschwartz:
## Description:
* service logs file name is now `<enclave-name>--<service-name>-logs.txt`
* Focus input box on Create an Enclave so you can type without clicking
* Change the logging layout so it's more compact, no borders
* Removed error messages from the logs themselves, will need to revisit how we should be displaying that.
* removed dropdown on the New Enclave button so it goes straight to the Create new Enclave modal

YES


enclave-manager/web/src/components/enclaves/tables/FilesTable.tsx
```diff
@@ -1,53 +1,46 @@
...
         header: "Name",
-        cell: ({ row, getValue }) => (
-          <Link to={`/enclave/${enclaveShortUUID}/file/${row.original.uuid}`}>
-            <Button size={"sm"} variant={"ghost"}>
-              {getValue()}
-            </Button>
-          </Link>
-        ),
+        cell: ({ row, getValue }) =>
+          //<Link to={`/enclave/${enclave.shortenedUuid}/file/${row.original.fileUuid}`}>
```
tedim52: can we remove dead code?

tedim52 (approved):

enclave-manager/web/src/components/enclaves/tables/FilesTable.tsx
adschwartz: yeah we should. Not sure if Ben is still working on this. He might be.

enclave-manager/web/src/emui/enclaves/enclave/service/logs/ServiceLogs.tsx
```diff
@@ -45,7 +45,8 @@ export const ServiceLogs = ({ enclave, service }: ServiceLogsProps) => {
         if (cancelled) {
           return;
         }
-        setLogLines((logLines) => [...logLines, { message: `Error: ${stringifyError(error)}`, status: "error" }]);
+        // TODO: We need a way to show this error
+        alert (`Error: ${stringifyError(error)}`)
```
Dartoxian, line 49: Easiest option that's nicer than `alert` would be to use a `toast`. Slightly nicer would be to store the message in state called `error` and use a `KurtosisAlert` component under/above the log viewer to display the error.

Dartoxian (approved):

enclave-manager/web/src/emui/enclaves/enclave/service/logs/ServiceLogs.tsx
adschwartz: Agreed and I thought about the toast but we don't actually want a toast. We need to auto-reconnect and not bother the user. As mentioned in the PR text we need to do more work here. It's just temporary to surface the error.

PR #1745, opened 2023-11-10
fix: run package with no args

tedim52:
YES (bug fix)


tedim52:
did i rebuild the EM UI correctly? I think so

adschwartz (approved):

PR #1750, opened 2023-11-10
chore: some fix for flaky tests

leoporoli:
## Description:
some fix for flaky tests

NO

More here: https://www.notion.so/kurtosistech/Flaky-tests-analysis-094fef0f64be428c9cd32b49c1c94f9e


h4ck3rk3y:
Continuity seems to fail still

leoporoli:
@h4ck3rk3y that error is related to this: https://kurtosistech.slack.com/archives/C04HLLL98J2/p1700165857744079 so, yeap I'll wait for that fix and I'll merge it here, then we can try again

leoporoli:
@h4ck3rk3y now it's ready

.circleci/config.yml
```diff
@@ -131,7 +131,11 @@ executors:
     machine:
       # This custom CircleCi Ubuntu image has a pre-installed Docker, so we are not using the CircleCI remote Docker when we
       # initiate this machine. More here: https://discuss.circleci.com/t/linux-machine-executor-images-2022-january-q1-update/42831
-      image: ubuntu-2004:2023.07.1
+      image: ubuntu-2004:2023.10.1
+  ubuntu_vm_xlarge:
```
adschwartz: @leoporoli do we hav an idea of the extra cost this will incur?

adschwartz (approved):

.circleci/config.yml
leoporoli: @adschwartz 
based on this info: https://circleci.com/product/features/resource-classes/
it will cost `100 credits/min` and the actual "large" costs `20 credits/min` so yeap probably is not worth it to activate it now, I'll downgrade it and see if the other changes help without this

PR #1818, opened 2023-11-17
feat: Unified REST API 

lostbean:
## Description:
This PR creates unified and first class REST API to manage the Kurtosis cluster (enclaves, services, ...). The REST APIs are defined using OpenAPI 3 specs, oapi-codegen to generate server and golang bindings and echo to server it. It also server Swagger UI to allow exploring the current API documentation.

!!! WARING !!! This PR also include bumping the Go compiler to `1.20` due requirement in some of the new dependencies for OpenAPI specs.

No

BEGIN_COMMIT_OVERRIDE
feat: Unified REST API
END_COMMIT_OVERRIDE


api/openapi/engine/engine_service.yaml
```diff
@@ -0,0 +1,337 @@
...
+      required:
+        - enclave_uuid
+        - name
+        - shortened_uuid
+        - containers_status
+        - api_container_status
+        - creation_time
+        - mode
+
+    EnclaveAPIContainerInfo:
```
omar711: In parallel I've added a couple of fields to `EnclaveAPIContainerInfo` and `EnclaveAPIContainerHostMachineInfo` to expose the tunnel server's ports. 
Reference is here: https://github.com/kurtosis-tech/kurtosis/pull/1849/files#diff-a24e12c03e41a9365ffc8601baddeee397de4b5547611039da5e15ac24d82a22R106 

Not sure on best timing, but we could add these fields right away, make them required and populate them with zeros to begin with?  Specifically this would be fields:

* `EnclaveAPIContainerInfo.tunnel_port_inside_enclave (integer, required)`
* `EnclaveAPIContainerHostMachineInfo.tunnel_port_on_host_machine (integer, required)`

which should avoid transitions at a later date

.circleci/config.yml
```diff
@@ -16,7 +16,7 @@ param_cli_cluster_backend: &param_cli_cluster_backend
     type: enum
     enum: ["docker", "kubernetes"]
 
-# Install go version 1.19 to run our integration tests on the minimal version we support
```
leoporoli, line 19: Make sure that you merge and release these changes on a breaking change day.

api/golang/core/kurtosis_core_rest_api_bindings/api_container_server.gen.go
leoporoli, line 1: TS bindings are missing, or do you plan to add them later?
What will happen with the Rust SDK? Will it be supported?

api/openapi/core/server.cfg.yaml
```diff
@@ -0,0 +1,6 @@
+package: kurtosis_core_rest_api_bindings
+generate:
+  embedded-spec: true
+  echo-server: true
+  strict-server: true
+output: api/golang/core/kurtosis_core_rest_api_bindings/api_container_server.gen.go
```
leoporoli: nit: end line missing

api/openapi/core/types.cfg.yaml
```diff
@@ -0,0 +1,4 @@
+package: kurtosis_core_rest_api_bindings
+generate:
+  models: true
+output: api/golang/core/kurtosis_core_rest_api_bindings/api_container_types.gen.go
```
leoporoli: nit: end line missing

api/openapi/engine/engine_service.yaml
```diff
@@ -0,0 +1,337 @@
+openapi: 3.0.0
+
+info:
+  title: Engine API
```
leoporoli: WDYTA specifying that it is a REST API?

api/openapi/engine/engine_service.yaml
```diff
@@ -0,0 +1,337 @@
...
+  version: 0.1.0
```
leoporoli: for acquiring knowledge. When do we have to bump this version? when we make any small change in the API right? what happens if, for instance, we add a 500 response line in an endpoint, do we have to bump the version in that case?

api/openapi/specs/core_service.yaml
```diff
@@ -0,0 +1,331 @@
...
+              schema:
+                $ref: "./api_types.yaml#/components/schemas/StarlarkDescription"
+
+  /enclaves/{enclave_identifier}/starlark/packages:
+    post:
+      description: Uploads a Starlark package. This step is required before the package can be executed with RunStarlarkPackage
+      parameters:
+        - $ref: './api_types.yaml#/components/parameters/enclave_identifier'
+      requestBody:
+        $ref: "./api_types.yaml#/components/requestBodies/fileUploadBody"
```
leoporoli: I don't see the fileUploadBody inside the "api_types.yaml" Do I missing something?

leoporoli: is this a stream right? so no type definition I guess

api/openapi/specs/core_service.yaml
```diff
@@ -0,0 +1,331 @@
...
+      responses:
+        default:
+          $ref: '#/components/responses/NotOk'
+        "200":
+          description: Success
+
+  /enclaves/{enclave_identifier}/starlark/packages/{package_id}:
+    post:
+      description: Executes a Starlark script on the user's behalf
```
leoporoli:
```suggestion
      description: Executes a Starlark package on the user's behalf
```

api/openapi/specs/core_service.yaml
```diff
@@ -0,0 +1,331 @@
...
+              schema:
+                type: string
+                format: binary
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "./api_types.yaml#/components/schemas/StarlarkRunResponseLine"
+
+  /enclaves/{enclave_identifier}/starlark/scripts:
```
leoporoli: suggestion, it could be also "/enclaves/{enclave_identifier}/starlark/module" because I single file script is a module and several modules compose a package

api/openapi/specs/core_service.yaml
```diff
@@ -0,0 +1,331 @@
...
+    post:
+      description: Executes a Starlark script on the user's behalf
```
leoporoli: it can also be "module"

leoporoli: mmm but now thinking about it, the methods in the backend are using "script" so yeap! maybe we should continue with that pattern

api/openapi/specs/core_service.yaml
```diff
@@ -0,0 +1,331 @@
...
+              schema:
+                type: string
+                format: binary
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "./api_types.yaml#/components/schemas/StarlarkRunResponseLine"
+
+  /enclaves/{enclave_identifier}/services/{service_identifier}:
```
leoporoli: missing description

api/openapi/specs/core_service.yaml
```diff
@@ -0,0 +1,331 @@
...
+        "200":
+          description: Successful request
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "./api_types.yaml#/components/schemas/ServiceIdentifiers"
+
+  /enclaves/{enclave_identifier}/services:
```
leoporoli: missing description

api/openapi/specs/core_service.yaml
```diff
@@ -0,0 +1,331 @@
...
+          content:
+            application/json:
+              schema:
+                type: object
+                additionalProperties:
+                  $ref: "./api_types.yaml#/components/schemas/ServiceInfo"
+
+  /enclaves/{enclave_identifier}/services/{service_identifier}/command:
+    post:
+      description: Executes the given command inside a running container
```
leoporoli:
```suggestion
      description: Executes the given command inside a running service's container
```

api/golang/core/kurtosis_core_rest_api_bindings/api_container_server.gen.go
lostbean: I'm planning to add the TS/Go bindings on a second PR.

Regarding the Rust SDK, didn't plan to include that. @leeederek / @h4ck3rk3y should I include a migration of RUST SDK now? Do we have active users relying on it?

h4ck3rk3y: i went searching on GitHub

https://github.com/search?q=-org%3Akurtosis-tech+kurtosis-sdk&type=code

i can't see it in a `cargo.toml` thats not a Fork of Kurtosis; LayerZero has a fork that they committed to in October; but not sure if they use rust/the fork

api/openapi/specs/core_service.yaml
```diff
@@ -0,0 +1,331 @@
...
+
+  /enclaves/{enclave_identifier}/services/{service_identifier}/endpoints/{port_number}/availability:
+    get:
+      description: Block until the given HTTP endpoint returns available, calling it through a HTTP request
+      parameters:
+        - $ref: './api_types.yaml#/components/parameters/enclave_identifier'
+        - $ref: './api_types.yaml#/components/parameters/service_identifier'
+        - $ref: './api_types.yaml#/components/parameters/port_number'
+        - $ref: './api_types.yaml#/components/parameters/http_method'
+        - $ref: './api_types.yaml#/components/parameters/path'
```
leoporoli: some of these parameters are query params right? should they be specified here?

api/openapi/specs/core_service.yaml
```diff
@@ -0,0 +1,331 @@
...
+        - $ref: './api_types.yaml#/components/parameters/initial_delay_milliseconds'
+        - $ref: './api_types.yaml#/components/parameters/retries'
+        - $ref: './api_types.yaml#/components/parameters/retries_delay_milliseconds'
+        - $ref: './api_types.yaml#/components/parameters/expected_response'
+        - $ref: './api_types.yaml#/components/parameters/request_body'
```
leoporoli: my only concern here is that GET request has char limit I think, and the request body could be long (in general should be small, but for corner cases), but probably is ok for now and if we see users having this problem we can move it to a POST request maybe

api/openapi/specs/engine_service.yaml
```diff
@@ -0,0 +1,149 @@
...
+      summary: Get Enclaves
+      responses:
+        default:
+          $ref: '#/components/responses/NotOk'
+        '200':
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                type: object
```
leoporoli: shouldn't it return an array?

api/openapi/specs/engine_service.yaml
```diff
@@ -0,0 +1,149 @@
...
+    post:
+      summary: Create Enclave
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: './api_types.yaml#/components/schemas/CreateEnclave'
+      responses:
+
```
leoporoli: nit: empty line

api/openapi/specs/engine_service.yaml
```diff
@@ -0,0 +1,149 @@
...
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                $ref: './api_types.yaml#/components/schemas/EnclaveInfo'
+                
+    delete:
+      summary: Delete Enclaves
+      parameters:
+        - $ref: './api_types.yaml#/components/parameters/remove_all'
```
leoporoli: probably the description could describe what happens when the value is false, I think only stopped enclave will be removed, right?

api/openapi/specs/engine_service.yaml
```diff
@@ -0,0 +1,149 @@
...
+        - $ref: './api_types.yaml#/components/parameters/enclave_identifier'
+      responses:
+        default:
+          $ref: '#/components/responses/NotOk'
+        '200':
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                $ref: './api_types.yaml#/components/schemas/EnclaveContainersStatus'
```
leoporoli: could it be just EnclaveStatus?

api/openapi/specs/websocket_service.yaml
```diff
@@ -0,0 +1,42 @@
...
+  title: Websocket API
+  version: 0.1.0
+
+paths:
+  /enclaves/{enclave_identifier}/logs:
+    get:
+      summary: Get Service Logs
+      parameters:
+        - $ref: "./api_types.yaml#/components/parameters/enclave_identifier"
+        - $ref: "./api_types.yaml#/components/parameters/service_uuid_set"
```
leoporoli: some of these are query params right?

api/openapi/specs/websocket_service.yaml
```diff
@@ -0,0 +1,42 @@
+openapi: 3.0.0
+
+info:
+  title: Websocket API
+  version: 0.1.0
+
+paths:
+  /enclaves/{enclave_identifier}/logs:
+    get:
+      summary: Get Service Logs
```
leoporoli:
```suggestion
      summary: Get Enclave's Services Logs
```

api/openapi/specs/websocket_service.yaml
```diff
@@ -0,0 +1,42 @@
...
+              schema:
+                $ref: "./api_types.yaml#/components/schemas/ServiceLogs"
+
+  /enclaves/{enclave_identifier}/services/{service_identifier}/logs:
+    get:
+      summary: Get Service Logs
+      parameters:
+        - $ref: "./api_types.yaml#/components/parameters/enclave_identifier"
+        - $ref: "./api_types.yaml#/components/parameters/service_identifier"
+        - $ref: "./api_types.yaml#/components/parameters/follow_logs"
```
leoporoli: some of these are query params right?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -17,13 +19,13 @@ import (
...
 	"github.com/kurtosis-tech/stacktrace"
 	"github.com/sirupsen/logrus"
-	"time"
 )
 
 const (
 	//TODO: pass this parameter
 	enclaveManagerUIPort                        = 9711
 	enclaveManagerAPIPort                       = 8081
+	restAPIPort                                 = 9779 //TODO: pass this parameter
```
leoporoli, line 28: will you make this change in the next PR?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -134,6 +136,16 @@ func CreateEngine(
 		)
 	}
 
+	restAPIPortSpec, err := port_spec.NewPortSpec(uint16(restAPIPort), consts.EngineTransportProtocol, consts.HttpApplicationProtocol, defaultWait)
+	if err != nil {
+		return nil, stacktrace.Propagate(
+			err,
+			"An error occurred creating the Enclave Manager UI's http port spec object using number '%v' and protocol '%v'",
```
leoporoli: is this the enclave manager or the engine's rest api?

engine/server/engine/main.go
```diff
@@ -60,6 +69,8 @@ const (
 	indexPath                   = "index.html"
 
 	shouldFlushMetricsClientQueueOnEachEvent = false
+
+	restAPIPortAddr uint16 = 9779
```
leoporoli: it should be received on the serverArgs

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	enclave_manager *enclave_manager.EnclaveManager,
+	perWeekLogsDatabaseClient centralized_logs.LogsDatabaseClient,
+	perFileLogsDatabaseClient centralized_logs.LogsDatabaseClient,
+	logFileManager *log_file_manager.LogFileManager,
+	metricsClient metrics_client.MetricsClient,
+) {
+	logrus.Info("Running REST API server...")
+
+	// This is how you set up a basic Echo router
+	e := echo.New()
```
leoporoli: I suggest using semantic var names, we try to not use one char var name.

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	// Log all requests
+	e.Use(echomiddleware.Logger())
+
+	// ============================== Engine Management API ======================================
+	// OpenAPI schema.
+	swagger_engine, err := engineApi.GetSwagger()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error loading swagger spec\n: %s", err)
```
leoporoli: can we replace this with a Logrus function?

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	logrus.Info("Running REST API server...")
+
+	// This is how you set up a basic Echo router
+	e := echo.New()
+	// Log all requests
+	e.Use(echomiddleware.Logger())
+
+	// ============================== Engine Management API ======================================
+	// OpenAPI schema.
+	swagger_engine, err := engineApi.GetSwagger()
```
leoporoli: I suggest to change the var name to swaggerEngine

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error loading swagger spec\n: %s", err)
+		os.Exit(1)
```
leoporoli: what will happen in this case? will the other engine servers gracefully shut down?

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+		ImageVersionTag: serverArgs.ImageVersionTag,
+		EnclaveManager:  enclave_manager,
+		LogFileManager:  logFileManager,
+		MetricsClient:   metricsClient,
+	}
+	engineApi.RegisterHandlers(e, engineApi.NewStrictHandler(engineRuntime, nil))
+
+	// ============================== Logging API ======================================
+
+	swagger_websocket, err := loggingApi.GetSwagger()
```
leoporoli: I suggest to change the var name to swaggerWebsocket

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error loading swagger spec\n: %s", err)
```
leoporoli: same comments here for using logrus

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	e.Use(echomiddleware.Logger())
+
+	// ============================== Engine Management API ======================================
+	// OpenAPI schema.
+	swagger_engine, err := engineApi.GetSwagger()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error loading swagger spec\n: %s", err)
+		os.Exit(1)
+	}
+	server.ServeSwaggerUI(e, "/api/specs/engine", server.NewSwaggerUIConfig(swagger_engine))
```
leoporoli: suggest to use a constant for the second argument

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	engineApi.RegisterHandlers(e, engineApi.NewStrictHandler(engineRuntime, nil))
+
+	// ============================== Logging API ======================================
+
+	swagger_websocket, err := loggingApi.GetSwagger()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error loading swagger spec\n: %s", err)
+		os.Exit(1)
+	}
+	server.ServeSwaggerUI(e, "/api/specs/websocket", server.NewSwaggerUIConfig(swagger_websocket))
```
leoporoli: suggest to use a constant for the second argument

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+		ImageVersionTag:             serverArgs.ImageVersionTag,
+		EnclaveManager:              enclave_manager,
+		MetricsUserID:               serverArgs.MetricsUserID,
+		DidUserAcceptSendingMetrics: serverArgs.DidUserAcceptSendingMetrics,
+		PerWeekLogsDatabaseClient:   perWeekLogsDatabaseClient,
+		PerFileLogsDatabaseClient:   perFileLogsDatabaseClient,
+		LogFileManager:              logFileManager,
+		MetricsClient:               metricsClient,
+	}
+	// TODO(edgar) add logging Close()
```
leoporoli: is this TODO for the next PR's? just calling in case you forgot.

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	// defer webSocketRuntime.ShutDown()
```
leoporoli: it should be activated, right?

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	loggingApi.RegisterHandlers(e, webSocketRuntime)
+
+	// ============================== Engine Management API ======================================
+	// OpenAPI schema.
+	swagger_enclave, err := enclaveApi.GetSwagger()
```
leoporoli: I suggest to change the var name to swaggerEnclave

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error loading swagger spec\n: %s", err)
```
leoporoli: same here for Logrus

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+		os.Exit(1)
+	}
+	server.ServeSwaggerUI(e, "/api/specs/enclave", server.NewSwaggerUIConfig(swagger_enclave))
```
leoporoli: suggest to use a constant for the second argument

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+
+	// Use our validation middleware to check all requests against the
+	// e.Use(middleware.OapiRequestValidator(swagger_enclave))
+	enclaveRuntime, err := restApi.NewEnclaveRuntime(ctx, *enclave_manager, false)
+	if err != nil {
+		// TODO(edgar) fix error handling
```
leoporoli: missing propagate I think

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	}
+	// TODO(edgar) add enclave Close()
+	// defer enclaveRuntime.ShutDown()
```
leoporoli: it should be activated, right? And the enclave close maybe?

engine/server/engine/main.go
```diff
@@ -338,3 +359,87 @@ func formatFilenameFunctionForLogs(filename string, functionName string) string
...
+	enclaveApi.RegisterHandlers(e, enclaveApi.NewStrictHandler(enclaveRuntime, nil))
+
+	// ============================== Start Server ======================================
+	e.Logger.Fatal(e.Start(net.JoinHostPort("0.0.0.0", fmt.Sprint(restAPIPortAddr))))
```
leoporoli: suggest using a constant for "0.0.0.0"

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+
+type enclaveRuntime struct {
+	enclaveManager           enclave_manager.EnclaveManager
+	remoteApiContainerClient map[string]rpc_api.ApiContainerServiceClient
+	connectOnHostMachine     bool
+	ctx                      context.Context
+	lock                     sync.Mutex
+}
+
+func (runtime enclaveRuntime) refreshEnclaveConnections() error {
```
leoporoli: nit: we usually put the private functions or methods after the public ones

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+	runtime.lock.Lock()
+	defer runtime.lock.Unlock()
+
+	enclaves, err := runtime.enclaveManager.GetEnclaves(runtime.ctx)
+	if err != nil {
+		return err
```
leoporoli: missing propagate

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+	}
+
+	// Clean up removed enclaves
+	for uuid := range runtime.remoteApiContainerClient {
```
leoporoli: what happens if this map contains a client for an enclave that was not returned by the enclave manager? shouldn't it be also deleted or not?

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+		_, found := enclaves[uuid]
+		if !found {
+			delete(runtime.remoteApiContainerClient, uuid)
+		}
+	}
+
+	// Add new enclaves - assuming enclaves properties (API container connection) are immutable
+	for uuid, info := range enclaves {
+		_, found := runtime.remoteApiContainerClient[uuid]
+		if !found && (info != nil) {
```
leoporoli:
```suggestion
		if !found && info != nil {
```

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+			conn, err := getGrpcClientConn(*info, runtime.connectOnHostMachine)
+			if err != nil {
+				logrus.Errorf("Failed to establish gRPC connection with enclave manager service on enclave %s", uuid)
+				return err
```
leoporoli: missing propagate

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+	runtime := enclaveRuntime{
+		enclaveManager:           manager,
+		remoteApiContainerClient: map[string]rpc_api.ApiContainerServiceClient{},
+		connectOnHostMachine:     connectOnHostMachine,
+		ctx:                      ctx,
+	}
+
+	err := runtime.refreshEnclaveConnections()
+	if err != nil {
+		return nil, err
```
leoporoli: missing propagate

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+func (manager *enclaveRuntime) GetEnclavesEnclaveIdentifierArtifacts(ctx context.Context, request api.GetEnclavesEnclaveIdentifierArtifactsRequestObject) (api.GetEnclavesEnclaveIdentifierArtifactsResponseObject, error) {
+	enclave_identifier := request.EnclaveIdentifier
+	apiContainerClient, responseErr := manager.GetApiClientOrResponseError(enclave_identifier)
+	if responseErr != nil {
+		return api.GetEnclavesEnclaveIdentifierArtifactsdefaultJSONResponse{Body: *responseErr, StatusCode: int(responseErr.Code)}, nil
+	}
+
+	artifacts, err := (*apiContainerClient).ListFilesArtifactNamesAndUuids(ctx, &emptypb.Empty{})
+	if err != nil {
+		return nil, err
```
leoporoli: missing propagation and I suggest to review it throughout the file

engine/server/engine/utils/list.go
```diff
@@ -0,0 +1,21 @@
...
+
+func MapList[T, U any](data []T, f func(T) U) []U {
+	res := make([]U, 0, len(data))
+	for _, e := range data {
+		res = append(res, f(e))
+	}
+	return res
+}
+
+func MapListStopOnError[T, U any](data []T, f func(T) (U, error)) ([]U, error) {
```
leoporoli: is not used

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+				Uuid: x.FileUuid,
+			}
+		})
+
+	return api.GetEnclavesEnclaveIdentifierArtifacts200JSONResponse(results), nil
+}
+
+// (POST /enclaves/{enclave_identifier}/artifacts/local-file)
+func (manager *enclaveRuntime) PostEnclavesEnclaveIdentifierArtifactsLocalFile(ctx context.Context, request api.PostEnclavesEnclaveIdentifierArtifactsLocalFileRequestObject) (api.PostEnclavesEnclaveIdentifierArtifactsLocalFileResponseObject, error) {
+	enclave_identifier := request.EnclaveIdentifier
```
leoporoli: use golang var name convention, I suggest to review it throughout the file

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+		}
+		filename := part.FileName()
+
+		client, err := (*apiContainerClient).UploadFilesArtifact(ctx)
+		if err != nil {
+			return nil, stacktrace.Propagate(err, "Can't start file upload gRPC call with enclave %s", enclave_identifier)
+		}
+		clientStream := grpc_file_streaming.NewClientStream[rpc_api.StreamedDataChunk, rpc_api.UploadFilesArtifactResponse](client)
+
+		response, err := clientStream.SendData(
```
leoporoli: missing err check

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+			filename,
+			part,
+			0, // Length unknown head of time
```
leoporoli: magic number

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+					PreviousChunkHash: previousChunkHash,
+					Metadata: &rpc_api.DataChunkMetadata{
+						Name: filename,
+					},
+				}, nil
+			},
+		)
+
+		// The response is nil when a file artifact with the same has already been uploaded
+		// TODO (edgar) Is this the expected behavior? If so, we should be explicit about it.
```
leoporoli: I think we should be explicit about it, probably a 400 status code

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+			return dataChunk.Data, dataChunk.PreviousChunkHash, nil
+		},
+	)
+
+	response := api.GetEnclavesEnclaveIdentifierArtifactsArtifactIdentifierDownload200ApplicationoctetStreamResponse{
+		Body:          pipeReader,
+		ContentLength: 0, // No file size is provided since we are streaming it directly
+	}
+
+	return api.GetEnclavesEnclaveIdentifierArtifactsArtifactIdentifierDownload200ApplicationoctetStreamResponse(response), nil
```
leoporoli: is not necessary to cast the var here because it's already this type

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+	pipeReader := clientStream.PipeReader(
+		artifact_identifier,
+		func(dataChunk *rpc_api.StreamedDataChunk) ([]byte, string, error) {
+			return dataChunk.Data, dataChunk.PreviousChunkHash, nil
+		},
+	)
+
+	response := api.GetEnclavesEnclaveIdentifierArtifactsArtifactIdentifierDownload200ApplicationoctetStreamResponse{
+		Body:          pipeReader,
+		ContentLength: 0, // No file size is provided since we are streaming it directly
```
leoporoli: magic number

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+	services, err := (*apiContainerClient).GetServices(ctx, &getServicesArgs)
+	if err != nil {
+		logrus.Errorf("Can't list services using gRPC call with enclave %s, error: %s", enclave_identifier, err)
+		return nil, stacktrace.NewError("Can't  list services using gRPC call with enclave %s", enclave_identifier)
+	}
+
+	mapped_services := utils.MapMapValues(services.ServiceInfo, toHttpServiceInfo)
+	selected_service, found := mapped_services[service_identifier]
+	if !found {
+		// TODO(edgar) add 404 return
```
leoporoli: just calling if we forgot this on this PR

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+	}
+	logrus.Infof("Run Starlark package on enclave %s", enclave_identifier)
+
+	package_id := request.PackageId
+	flags := utils.MapList(utils.DerefWith(request.Body.ExperimentalFeatures, []api_type.KurtosisFeatureFlag{}), toGrpcFeatureFlag)
+	// The gRPC always expect a JSON object even though it's marked as optional, so we need to default to `{}``
+	jsonParams := utils.DerefWith(request.Body.Params, map[string]interface{}{})
+	jsonBlob, err := json.Marshal(jsonParams)
+	if err != nil {
+		panic("Failed to serialize parameters")
```
leoporoli: shouldn't be better to return a request error or server error instead of panicking here? panic will stop the server I think

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+	client, err := (*apiContainerClient).RunStarlarkPackage(ctx, &runStarlarkPackageArgs)
+	if err != nil {
+		logrus.Errorf("Can't run Starlark package using gRPC call with enclave %s, error: %s", enclave_identifier, err)
+		return nil, stacktrace.NewError("Can't run Starlark package using gRPC call with enclave %s", enclave_identifier)
+	}
+	clientStream := grpc_file_streaming.NewClientStream[rpc_api.StreamedDataChunk, []byte](client)
+
+	// TODO: Why the gRPC is not sending the previous hash??? and we need to calc it manually
+	hasher := sha1.New()
+	previousBlockHash := ""
```
leoporoli: you can replace these with constants

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+		ImageDownloadMode:    utils.MapPointer(request.Body.ImageDownloadMode, toGrpcImageDownloadMode),
+	}
+	client, err := (*apiContainerClient).RunStarlarkScript(ctx, &runStarlarkScriptArgs)
+	if err != nil {
+		logrus.Errorf("Can't run Starlark script using gRPC call with enclave %s, error: %s", enclave_identifier, err)
+		return nil, stacktrace.NewError("Can't run Starlark script using gRPC call with enclave %s", enclave_identifier)
+	}
+	clientStream := grpc_file_streaming.NewClientStream[rpc_api.StreamedDataChunk, []byte](client)
+	pipeReader := clientStream.PipeReader(
+		"__RunStarlarkScript__",
```
leoporoli: I suggest to use a constant here

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+		func(dataChunk *rpc_api.StreamedDataChunk) ([]byte, string, error) {
+			return dataChunk.Data, dataChunk.PreviousChunkHash, nil
+		},
+	)
+	response := api.PostEnclavesEnclaveIdentifierStarlarkScripts200ApplicationoctetStreamResponse{
+		Body:          pipeReader,
+		ContentLength: 0, // No file size is provided since we are streaming it directly
```
leoporoli: magic numebr

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+
+func (manager enclaveRuntime) GetGrpcClientForEnclaveUUID(enclave_uuid string) (*rpc_api.ApiContainerServiceClient, error) {
+	err := manager.refreshEnclaveConnections()
+	if err != nil {
+		return nil, err
+	}
+
+	client, found := manager.remoteApiContainerClient[enclave_uuid]
+	if !found {
+		return nil, nil
```
leoporoli, line 771: I think it should return an error here or the caller should check for nil values

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+}
+
+func toGrpcConnect(conn api_type.Connect) rpc_api.Connect {
+	switch conn {
+	case api_type.CONNECT:
+		return rpc_api.Connect_CONNECT
+	case api_type.NOCONNECT:
+		return rpc_api.Connect_NO_CONNECT
+	default:
+		panic(fmt.Sprintf("Missing conversion of Connect Enum value: %s", conn))
```
leoporoli: I suggest returning an error instead of panicking here

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+func toHttpServiceStatus(status rpc_api.ServiceStatus) api_type.ServiceStatus {
+	switch status {
+	case rpc_api.ServiceStatus_RUNNING:
+		return api_type.ServiceStatusRUNNING
+	case rpc_api.ServiceStatus_STOPPED:
+		return api_type.ServiceStatusSTOPPED
+	case rpc_api.ServiceStatus_UNKNOWN:
+		return api_type.ServiceStatusUNKNOWN
+	default:
+		panic(fmt.Sprintf("Missing conversion of Service Status Enum value: %s", status))
```
leoporoli: same here

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+func toHttpContainerStatus(status rpc_api.Container_Status) api_type.ContainerStatus {
+	switch status {
+	case rpc_api.Container_RUNNING:
+		return api_type.ContainerStatusRUNNING
+	case rpc_api.Container_STOPPED:
+		return api_type.ContainerStatusSTOPPED
+	case rpc_api.Container_UNKNOWN:
+		return api_type.ContainerStatusUNKNOWN
+	default:
+		panic(fmt.Sprintf("Missing conversion of Container Status Enum value: %s", status))
```
leoporoli: same here

engine/server/engine/server/enclave_rest_api_handler.go
```diff
@@ -0,0 +1,857 @@
...
+func toHttpTransportProtocol(protocol rpc_api.Port_TransportProtocol) api_type.TransportProtocol {
+	switch protocol {
+	case rpc_api.Port_TCP:
+		return api_type.TCP
+	case rpc_api.Port_UDP:
+		return api_type.UDP
+	case rpc_api.Port_SCTP:
+		return api_type.SCTP
+	default:
+		panic(fmt.Sprintf("Missing conversion of Transport Protocol Enum value: %s", protocol))
```
leoporoli: same here

engine/server/engine/server/engine_connect_server_service.go
```diff
@@ -75,6 +79,94 @@ func NewEngineConnectServerService(
...
+func toGrpcEnclaveContainersStatus(status types.EnclaveContainersStatus) kurtosis_engine_rpc_api_bindings.EnclaveContainersStatus {
+	switch status {
+	case types.EnclaveContainersStatus_EMPTY:
+		return kurtosis_engine_rpc_api_bindings.EnclaveContainersStatus_EnclaveContainersStatus_EMPTY
+	case types.EnclaveContainersStatus_STOPPED:
+		return kurtosis_engine_rpc_api_bindings.EnclaveContainersStatus_EnclaveContainersStatus_STOPPED
+	case types.EnclaveContainersStatus_RUNNING:
+		return kurtosis_engine_rpc_api_bindings.EnclaveContainersStatus_EnclaveContainersStatus_RUNNING
+	default:
+		panic(fmt.Sprintf("Undefined mapping of value: %s", status))
```
leoporoli, line 79: propagate an error instead of panicking

engine/server/engine/server/engine_connect_server_service.go
```diff
@@ -75,6 +79,94 @@ func NewEngineConnectServerService(
...
+func toGrpcContainerStatus(status types.ContainerStatus) kurtosis_engine_rpc_api_bindings.EnclaveAPIContainerStatus {
+	switch status {
+	case types.ContainerStatus_NONEXISTENT:
+		return kurtosis_engine_rpc_api_bindings.EnclaveAPIContainerStatus_EnclaveAPIContainerStatus_NONEXISTENT
+	case types.ContainerStatus_STOPPED:
+		return kurtosis_engine_rpc_api_bindings.EnclaveAPIContainerStatus_EnclaveAPIContainerStatus_STOPPED
+	case types.ContainerStatus_RUNNING:
+		return kurtosis_engine_rpc_api_bindings.EnclaveAPIContainerStatus_EnclaveAPIContainerStatus_RUNNING
+	default:
+		panic(fmt.Sprintf("Undefined mapping of value: %s", status))
```
leoporoli, line 92: same here

engine/server/engine/server/engine_connect_server_service.go
```diff
@@ -75,6 +79,94 @@ func NewEngineConnectServerService(
...
+}
+
+func toGrpcEnclaveMode(mode types.EnclaveMode) kurtosis_engine_rpc_api_bindings.EnclaveMode {
+	switch mode {
+	case types.EnclaveMode_PRODUCTION:
+		return kurtosis_engine_rpc_api_bindings.EnclaveMode_PRODUCTION
+	case types.EnclaveMode_TEST:
+		return kurtosis_engine_rpc_api_bindings.EnclaveMode_TEST
+	default:
+		panic(fmt.Sprintf("Undefined mapping of value: %s", mode))
```
leoporoli, line 123: same here

engine/server/engine/server/engine_rest_api_handler.go
```diff
@@ -0,0 +1,279 @@
...
+// (GET /enclaves/{enclave_identifier})
+func (engine EngineRuntime) GetEnclavesEnclaveIdentifier(ctx context.Context, request api.GetEnclavesEnclaveIdentifierRequestObject) (api.GetEnclavesEnclaveIdentifierResponseObject, error) {
+	infoForEnclaves, err := engine.EnclaveManager.GetEnclaves(ctx)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred getting info for enclaves")
+	}
+	info, found := infoForEnclaves[request.EnclaveIdentifier]
+	if !found {
+		notFoundErr := stacktrace.NewError("Enclave '%s' not found.", request.EnclaveIdentifier)
+		return nil, notFoundErr
```
leoporoli: will it return a 404 status code? shouldn't it be a "ResponseInfo" type?

leoporoli: I see that the enclave handler is returning the "ResponseInfo" which specifies the status code but this handler isn't. shouldn't it return it?

engine/server/engine/server/engine_rest_api_handler.go
```diff
@@ -0,0 +1,279 @@
...
+func toHttpApiEnclaveContainersStatus(status types.EnclaveContainersStatus) api_type.EnclaveContainersStatus {
+	switch status {
+	case types.EnclaveContainersStatus_EMPTY:
+		return api_type.EnclaveContainersStatusEMPTY
+	case types.EnclaveContainersStatus_STOPPED:
+		return api_type.EnclaveContainersStatusSTOPPED
+	case types.EnclaveContainersStatus_RUNNING:
+		return api_type.EnclaveContainersStatusRUNNING
+	default:
+		panic(fmt.Sprintf("Undefined mapping of value: %s", status))
```
leoporoli: propagate error instead of panicking here

engine/server/engine/server/engine_rest_api_handler.go
```diff
@@ -0,0 +1,279 @@
...
+func toHttpApiContainerStatus(status types.ContainerStatus) api_type.ApiContainerStatus {
+	switch status {
+	case types.ContainerStatus_NONEXISTENT:
+		return api_type.ApiContainerStatusNONEXISTENT
+	case types.ContainerStatus_STOPPED:
+		return api_type.ApiContainerStatusSTOPPED
+	case types.ContainerStatus_RUNNING:
+		return api_type.ApiContainerStatusRUNNING
+	default:
+		panic(fmt.Sprintf("Undefined mapping of value: %s", status))
```
leoporoli: same here

engine/server/engine/server/engine_rest_api_handler.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func toHttpApiEnclaveMode(mode types.EnclaveMode) api_type.EnclaveMode {
+	switch mode {
+	case types.EnclaveMode_PRODUCTION:
+		return api_type.PRODUCTION
+	case types.EnclaveMode_TEST:
+		return api_type.TEST
+	default:
+		panic(fmt.Sprintf("Undefined mapping of value: %s", mode))
```
leoporoli: same here

engine/server/engine/server/engine_rest_api_handler.go
```diff
@@ -0,0 +1,279 @@
...
+func toHttpEnclaveContainersStatus(status types.EnclaveContainersStatus) api_type.EnclaveContainersStatus {
+	switch status {
+	case types.EnclaveContainersStatus_STOPPED:
+		return api_type.EnclaveContainersStatusSTOPPED
+	case types.EnclaveContainersStatus_RUNNING:
+		return api_type.EnclaveContainersStatusRUNNING
+	case types.EnclaveContainersStatus_EMPTY:
+		return api_type.EnclaveContainersStatusEMPTY
+	default:
+		panic(fmt.Sprintf("Undefined mapping of value: %s", status))
```
leoporoli: same here

engine/server/engine/server/websocket_api_handler.go
```diff
@@ -0,0 +1,386 @@
...
+	// per file pulls logs from enclaves created pre log retention feature
+	// TODO: remove once users are fully migrated to log retention/new log schema
+	PerFileLogsDatabaseClient centralized_logs.LogsDatabaseClient
+
+	LogFileManager *log_file_manager.LogFileManager
+
+	MetricsClient metrics_client.MetricsClient
+}
+
+func sendErrorCode(ctx echo.Context, code int, message string) error {
```
leoporoli: we usually put private functions after public ones

engine/server/engine/server/websocket_api_handler.go
```diff
@@ -0,0 +1,386 @@
...
+		ctx,
+		enclaveIdentifier,
+		utils.MapList(params.ServiceUuidSet, func(x string) user_service.ServiceUUID { return user_service.ServiceUUID(x) }),
+		params.FollowLogs,
+		params.ReturnAllLogs,
+		utils.MapPointer(params.NumLogLines, func(x int) uint32 { return uint32(x) }),
+		params.ConjunctiveFilters,
+	)
+	if err != nil {
+		logrus.Error(err)
```
leoporoli: I think we could add more context in this error, "Failed to get log streamer with error:" for instance

engine/server/engine/server/websocket_api_handler.go
```diff
@@ -0,0 +1,386 @@
...
+	serviceUuidList []user_service.ServiceUUID,
+	maybeShouldFollowLogs *bool,
+	maybeShouldReturnAllLogs *bool,
+	maybeNumLogLines *uint32,
+	maybeFilters *[]api_type.LogLineFilter,
+) (*LogStreamer, error) {
+	enclaveUuid, err := engine.EnclaveManager.GetEnclaveUuidForEnclaveIdentifier(context.Background(), enclaveIdentifier)
+	if err != nil {
+		logrus.Errorf("An error occurred while fetching uuid for enclave '%v'. This could happen if the enclave has been deleted. Treating it as UUID", enclaveIdentifier)
+		return nil, err
```
leoporoli: missing propagate

engine/server/engine/server/websocket_api_handler.go
```diff
@@ -0,0 +1,386 @@
...
+
+	var (
+		serviceLogsByServiceUuidChan chan map[user_service.ServiceUUID][]logline.LogLine
+		errChan                      chan error
+		cancelCtxFunc                func()
+	)
+
+	notFoundServiceUuids, err := engine.reportAnyMissingUuidsAndGetNotFoundUuidsListHttp(context, enclaveUuid, requestedServiceUuids)
+	if err != nil {
+		return nil, err
```
leoporoli: missing propagate, I suggest to check them  throughout the file

engine/server/engine/server/websocket_api_handler.go
```diff
@@ -0,0 +1,386 @@
...
+				}
+				logrus.Debug("Exiting the stream loop after receiving a close signal from the error chan")
+				return
+			}
+		}
+	}).ServeHTTP(ctx.Response(), ctx.Request())
+	return nil
+}
+
+func (streamer LogStreamer) streamHTTP(ctx echo.Context) error {
```
leoporoli: nice!!

leoporoli (commented):
approved with some comments

leoporoli (approved):

engine/server/engine/main.go
adschwartz: Agreed, they won't shut down. I've mentioned this before, but we should really not be running our webservers in the same containers as the engine itself. It's messy at best to handle the loss of a process inside the container. The best practice is to launch them as separate containers and use a healthcheck to detect the issue.
@lostbean

api/openapi/engine/engine_service.yaml
lostbean: Thanks for the heads up on these new fields but since they are additives we can also add them later without any break changes.

api/golang/core/kurtosis_core_rest_api_bindings/api_container_server.gen.go
lostbean: Cool, I'll skip it for now but we should decide to drop it or migrate it before the code clean after we are planning for later.

api/openapi/core/types.cfg.yaml
lostbean: done

api/openapi/engine/engine_service.yaml
lostbean: We should use https://semver.org/ here. So adding a 500 response would be a PATCH bump

api/openapi/specs/core_service.yaml
lostbean: Ended up merge all specs into a single file. But the definition is now [here](https://github.com/kurtosis-tech/kurtosis/blob/f3cb0c0d363bed2402603420b7e120d146980482/api/openapi/specs/kurtosis_api.yaml#L597)

api/openapi/specs/core_service.yaml
lostbean: Yes, they are specified in the `in:` field ([example](https://github.com/kurtosis-tech/kurtosis/blob/f3cb0c0d363bed2402603420b7e120d146980482/api/openapi/specs/kurtosis_api.yaml#L665))

api/openapi/specs/core_service.yaml
lostbean: Fair point! I think the limit is 2048 but most of these fields are optional and we don't expect them to any large since those are for status checking endpoints. From the RESTful perspective, and GET makes more sense since they have no side effect in the backend. Let's leave as a GET for now, and we can reconsider it if we actually start runinig into issues.

api/openapi/specs/engine_service.yaml
lostbean: Not an array but is returning dictionary map[string]enclaveInfo

api/openapi/specs/engine_service.yaml
lostbean: yes!

api/openapi/specs/engine_service.yaml
lostbean: +1

api/openapi/specs/websocket_service.yaml
lostbean: yes, the ones that are not in the path

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
lostbean: yes!

engine/server/engine/main.go
lostbean: I haven't put a lot of thought on this part of the code but you are both correct. I've improved and mitigated it for now (skipping errors from swagger since it's now a critical part of the engine). The swagger spec could move together with the Web Server (static asset server which btw is also [running from the engine](https://github.com/kurtosis-tech/kurtosis/blob/0c231668df9357a02b209118a5aa7f1eb85c1f24/engine/server/engine/main.go#L158)). I've put a  TODO to move it once we enter the clean-up and split the web server into its own service.

engine/server/engine/main.go
lostbean: implementation has changed but it's done

engine/server/engine/main.go
lostbean: implementation has changed but it's done

engine/server/engine/main.go
lostbean: implementation has changed but it's done

engine/server/engine/server/enclave_rest_api_handler.go
lostbean: Yes, that is what is going to happen. It basically removes the enclaves that are not listed in the GetEnclaves after a refresh. Added more info in the comment.

engine/server/engine/server/enclave_rest_api_handler.go
lostbean: +1, I added a per-file response using a 409

engine/server/engine/server/enclave_rest_api_handler.go
lostbean: IMO panicking is fine if you really don't expect that to fail, which is the case here. We are panicking here because we are serializing a deserialized JSON parameter and that never should fail. Also panicking here doesn't bring the server down, it only brings crashes the go routine handling that HTTP request. That said, I don't see any big issue on adding a bit of more code to pass the error back to the handler.

engine/server/engine/server/enclave_rest_api_handler.go
lostbean: Yes, and the client [does check for nil values](https://github.com/kurtosis-tech/kurtosis/blob/07c7f8e47f0dfab5a5fabea5b64b8cde1bd21ec7/engine/server/engine/server/enclave_rest_api_handler.go#L719)

engine/server/engine/server/enclave_rest_api_handler.go
lostbean: this is one of the panicking could make sense but we don't expect a failure under normal operations. This is an on-to-one mapping between enums and the only failure mode is if one adds a new value to the enum. I'm fine adding code to handle it throughout the codebase (as we need to keep passing it through the function stack calls) but it would be great if Go would have a native concept of enum exhaustion in switch statements (maybe we can work to add smtg like https://pkg.go.dev/github.com/nishanths/exhaustive to enforce that check).

engine/server/engine/server/engine_rest_api_handler.go
lostbean: Yes, I've also added them to engine api

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
lostbean: Ok, I decided indeed to move it to the clean-up phase, since of this ports/args might go away.

engine/server/engine/main.go
lostbean: Same here, I decided indeed to move it to the clean-up phase, since of this ports/args might go away.

engine/server/engine/server/enclave_rest_api_handler.go
lostbean: this piece of code is now gone

engine/server/engine/server/enclave_rest_api_handler.go
lostbean: I've added the [exhaustive](https://pkg.go.dev/github.com/nishanths/exhaustive) to our golangci-lint and the default case should never be reached.

Issue #1768, opened 2023-11-13
Unable to create new enclaves in em ui after upgrading engine


Dartoxian:
### What's your CLI version?

0.85.18

### Description & steps to reproduce

I am unable to create new enclaves in the em ui after I upgrade my local engine. These are the steps that I take:

1. `sudo apt update && sudo apt install --only-upgrade kurtosis-cli`
2. `kurtosis enclave ls` (just to trigger starting the engine with the new version and check it works)
3. Open the web ui, try to create a new enclave. The `run` button will spin, and eventually just stop spinning. No error is shown, but no enclave is created either.

I can work around this issue by manually creating an enclave on the cli (`kurtosis run github.com/kurtosis-tech/postgres-package`) - this seems to wake something up in my local instance and after this the web ui can be used to create enclaves.

I run kurtosis on Ubuntu 22.

This issue was also seen on 0.85.10.

### Desired behavior

After upgrading my engine I should be able to launch enclaves in the emui.

Additionally, the emui should show an error (rather than no message) when it does not create an enclave.

### What is the severity of this bug?

Papercut; this bug is frustrating, but I have a workaround.

### What area of the product does this pertain to?

Frontend: the Graphical User Interface, including the Enclave Manager

leoporoli:
This is the error shown in the UI: `Could not create an enclave, got: ConnectError: [unknown] NetworkError when attempting to fetch resource.`

This is the request that didn't receive a response from the EM server 
```
curl 'http://localhost:8081/kurtosis_enclave_manager.KurtosisEnclaveManagerServer/CreateEnclave' -X POST -H 'User-Agent: connect-es/0.13.2' -H 'Accept: */*' -H 'Accept-Language: en-GB,en;q=0.5' -H 'Accept-Encoding: gzip, deflate, br' -H 'Referer: http://localhost:9711/' -H 'connect-protocol-version: 1' -H 'content-type: application/json' -H 'Origin: http://localhost:9711/' -H 'Connection: keep-alive' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-site' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache' --data-raw '{"enclaveName":"","apiContainerVersionTag":"","apiContainerLogLevel":"info","mode":"PRODUCTION"}'
```

The engine server starts creating the enclave but for some reason, it doesn't finish.

[...]

PR #1829, opened 2023-11-17
refactor: inject Logrus logger in the Connect Go HTTP server and separated context while pulling Docker images

leoporoli:
## Description:
inject Logrus logger in the Connect Go HTTP server and separated context while pulling Docker images

NO

This is related to this ticket: https://github.com/kurtosis-tech/kurtosis/issues/1768


enclave-manager/server/server.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	listenPort                = 8081
+	listenPort                = 9081
```
adschwartz: @leoporoli this will break a good bit of stuff. It needs to be 8081

adschwartz (changes requested):
@leoporoli Port change is required, but otherwise good to go.

enclave-manager/server/server.go
leoporoli: oh! yes sorry, I changed it for testing purposes but I forgot to revert it

adschwartz (approved):

PR #1830, opened 2023-11-18
feat: search in service logs

### LINKED ISSUE #1791: Cloud: unable to find all occurrences of searched item in logs
### What's your CLI version?

N/A

When I try to find a certain log entry in https://cloud.kurtosis.com/enclave-manager/enclave/66a897a354f1/service/151acad62d58/logs I can't seem to find all occurrences of searched object with cmd+f. 

### Desired behavior

Would be good to either include a search box in the logs tab, or enable me to search the whole logs file with cmd+f. 

adschwartz:
## Description:
Adding a search bar to the service logs window
image

YES
Closes #1791


tedim52 (approved):

enclave-manager/web/src/components/enclaves/logs/LogLine.tsx
```diff
@@ -15,9 +18,21 @@ export type LogLineProps = {
...
+
+export type LogLineInput = {
+  logLineProps: LogLineProps;
+  logLineSearch?: LogLineSearch;
+  selected: boolean | undefined;
+};
+
 const logFontFamily = "Menlo, Monaco, Inconsolata, Consolas, Courier, monospace";
 
-export const LogLine = ({ timestamp, message, status }: LogLineProps) => {
```
Dartoxian, line 20: Here `LogLineProps` should have just become:
```
export type LogLineProps = {
      timestamp?: DateTime;
      message?: string;
      status?: LogStatus;
      selected?: boolean;
      logLineSearch?: LogLineSearch;
}
```

enclave-manager/web/src/components/enclaves/logs/LogLine.tsx
```diff
@@ -29,16 +44,58 @@ export const LogLine = ({ timestamp, message, status }: LogLineProps) => {
...
+
+  const HighlightPattern = ({
+    text,
+    regex,
+    selected,
+  }: {
+    text: string;
+    regex: RegExp;
+    selected: boolean | undefined;
+  }) => {
```
Dartoxian, line 69: I don't think this component (because that's what it is) should be inlined with `LogLineSearch` - this means it's redefined on every render _and_ defined for every `LogLine`. I think it's LogLine specific enough to just be another non exported component in this file.

enclave-manager/web/src/components/enclaves/logs/LogLine.tsx
```diff
@@ -29,16 +44,58 @@ export const LogLine = ({ timestamp, message, status }: LogLineProps) => {
     }
   };
 
-  const processText = (message: string) => {
-    if (hasAnsi(message)) {
-      return parse(convert.toHtml(message));
+  const processText = (text: string, selected: boolean | undefined) => {
```
Dartoxian, line 47: this is also actually another react component - it should be moved out of the `LogLine` component and given a better name as a (non exported) component like `Message`

enclave-manager/web/src/components/enclaves/logs/LogLine.tsx
```diff
@@ -4,6 +4,9 @@ import { DateTime } from "luxon";
 import { isDefined } from "../../../utils";
 // @ts-ignore
```
Dartoxian, line 5: Just noticed this - lets not use `has-ansi` (or its dependency `ansi-regex`) at all and just inline the code in `src/utils/index.ts` (and combine it with `stripAnsi`):

```
export function ansiRegex({ onlyFirst = false } = {}) {
  const pattern = [
    "[\\u001B\\u009B][[\\]()#;?]*(?:(?:(?:(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]+)*|[a-zA-Z\\d]+(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]*)*)?\\u0007)",
    "(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PR-TZcf-nq-uy=><~]))",
  ].join("|");

  return new RegExp(pattern, onlyFirst ? undefined : "g");
}

const regex = ansiRegex({ onlyFirst: true });

export function hasAnsi(text: string) {
  return regex.test(text);
}
```

enclave-manager/web/src/components/enclaves/logs/LogLine.tsx
```diff
@@ -4,6 +4,9 @@ import { DateTime } from "luxon";
...
 import hasAnsi from "has-ansi";
+import { ReactElement } from "react";
+import { normalizeLogText } from "./LogViewer";
```
Dartoxian, line 8: This is a circular dependency now between `LogLine` and `LogViewer` - this should be avoided, and this method should be moved to a utils file.

enclave-manager/web/src/components/enclaves/logs/LogViewer.tsx
```diff
@@ -24,9 +47,43 @@ export const LogViewer = ({
...
+        if (element === document.activeElement) {
+          e.preventDefault();
+          setSearch(undefined);
+          setRawSearchTerm("");
+          setSearchMatchesIndices([]);
+          setCurrentSearchIndex(undefined);
+        }
+      }
+    });
+  }, []);
```
Dartoxian, line 85: This effect should have an unsubscribe return so that the listener is not hanging around when this component is unmounted

enclave-manager/web/src/components/enclaves/logs/LogViewer.tsx
```diff
@@ -54,9 +111,158 @@ export const LogViewer = ({
       .join("\n");
   };
 
+  useEffect(() => {
+    if (search) findMatches(search);
+  }, [search?.searchTerm, logLines]);
```
Dartoxian, line 116: `findMatches` is a dependency, but `logLines` isn't. Also, I try to avoid `useEffect` that I am expecting to be triggered by a callback (which this is). Instead the callback should just include or call this functionality. If we want this callback to be called on first load then a useEffect with an empty dependency array is fine.

Also I generally try to structure components in roughly this order:
* `use` hooks (state and context)
* `useMemo`
* `useCallback` and functions
* `useEffect`
* jsx (and possibly conditional switches for different content`

I think this makes the code more readable, and less surprising (ie reading top to bottom at this point where did `findMatches` come from).

enclave-manager/web/src/components/enclaves/logs/LogViewer.tsx
```diff
@@ -54,9 +111,158 @@ export const LogViewer = ({
...
+    setCurrentSearchIndex(undefined);
+    const searchTerm = normalizeLogText(rawText);
+    const logLineSearch: LogLineSearch = {
+      searchTerm: searchTerm,
+      pattern: new RegExp(searchTerm, "gi"), // `i` is invariant case
+    };
+    setSearch(logLineSearch);
+  };
+  const debouncedUpdateSearchTerm = debounce(updateSearchTerm, 100);
+  const debouncedUpdateSearchTermCallback = useCallback(debouncedUpdateSearchTerm, []);
```
Dartoxian, line 128: `debouncedUpdateSearchTerm` is a dependency of `debouncedUpdateSearchTermCallback` so should be in the dependency array, but it's not. If it were, this would break `debouncedUpdateSearchTermCallback` as `debouncedUpdateSearchTerm` is changed on every render, instead it should just be defined in a single `useMemo` - like:

```
  const debouncedUpdateSearchTermCallback = useMemo(
    () =>
      debounce((rawText: string) => {
        setCurrentSearchIndex(undefined);
        const searchTerm = normalizeLogText(rawText);
        const logLineSearch: LogLineSearch = {
          searchTerm: searchTerm,
          pattern: new RegExp(searchTerm, "gi"), // `i` is invariant case
        };
        setSearch(logLineSearch);
      }, 100),
    [],
  );
```

enclave-manager/web/src/components/enclaves/logs/LogViewer.tsx
```diff
@@ -54,9 +111,158 @@ export const LogViewer = ({
...
+  useEffect(() => {
+    if (search) findMatches(search);
+  }, [search?.searchTerm, logLines]);
+
+  const updateSearchTerm = (rawText: string) => {
+    setCurrentSearchIndex(undefined);
+    const searchTerm = normalizeLogText(rawText);
+    const logLineSearch: LogLineSearch = {
+      searchTerm: searchTerm,
+      pattern: new RegExp(searchTerm, "gi"), // `i` is invariant case
```
Dartoxian, line 123: `i` is case insensitive

enclave-manager/web/src/components/enclaves/logs/LogViewer.tsx
```diff
@@ -54,9 +111,158 @@ export const LogViewer = ({
...
+    };
+    setSearch(logLineSearch);
+  };
+  const debouncedUpdateSearchTerm = debounce(updateSearchTerm, 100);
+  const debouncedUpdateSearchTermCallback = useCallback(debouncedUpdateSearchTerm, []);
+
+  const hasSearchTerm = () => {
+    if (!search) return false;
+    return isDefined(search.searchTerm) && isNotEmpty(search.searchTerm);
```
Dartoxian, line 132: (nit):
```
return isDefined(search?.searchTerm) && isNotEmpty(search.searchTerm);
```

enclave-manager/web/src/components/enclaves/logs/LogViewer.tsx
```diff
@@ -54,9 +111,158 @@ export const LogViewer = ({
...
+      setSearchMatchesIndices(matches);
+    }
+  };
+
+  const handleOnChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setRawSearchTerm(e.target.value);
+    debouncedUpdateSearchTermCallback(e.target.value);
+  };
+
+  const priorMatch = () => {
```
Dartoxian, line 154: (very minor) wasn't immediately obvious to me this is a function - should be named something like `handlePriorMatchClick`

enclave-manager/web/src/components/enclaves/logs/LogViewer.tsx
```diff
@@ -54,9 +111,158 @@ export const LogViewer = ({
...
+    }
+    setCurrentSearchIndex(newIndex);
+    return newIndex;
+  };
+
+  useEffect(() => {
+    if (virtuosoRef?.current && currentSearchIndex !== undefined && currentSearchIndex >= 0) {
+      virtuosoRef.current.scrollToIndex(searchMatchesIndices[currentSearchIndex]);
+    }
+  }, [currentSearchIndex]);
```
Dartoxian, line 183: this should not be a `useEffect` and should probably be either in a callback or called direct from the navigation/find matches code.

enclave-manager/web/src/components/enclaves/logs/LogViewer.tsx
```diff
@@ -54,9 +111,158 @@ export const LogViewer = ({
...
+                        <>
+                          <Editable
+                            p={0}
+                            m={0}
+                            size={"sm"}
+                            value={`${currentSearchIndex + 1}`}
+                            onChange={(inputString) =>
+                              updateSearchIndexBounded(parseMatchIndexRequest(inputString) - 1)
+                            }
+                          >
```
Dartoxian, line 249: I'm not familiar with this component, but wondering if we can apply numeric constraint to it.

Dartoxian (commented):
Nice job getting this out - I expect a decent amount of the feedback I've left here you're already aware of. The main piece of feedback I have is a strong preference to avoid chains of `useEffect` in response to callbacks - they make code harder to read and can eventually introduce bugs that are hard to trace. React has some notes on how to identify and remove avoidable effects here https://react.dev/learn/you-might-not-need-an-effect

PR #1847, opened 2023-11-24
feat: emui enhancements

Dartoxian:
## Description:
This PR includes a number or emui enhancements:
* Artifacts view
* fix to dict/list input sizes
* Update logging ui to match designs
* Update general app layout to match designs (headers)

Additionally the feedback left on https://github.com/kurtosis-tech/kurtosis/pull/1830 is implemented.

### Demo of the artifacts view

https://github.com/kurtosis-tech/kurtosis/assets/4419574/0849e46e-c410-4a07-a65f-bcc74297c53b

YES


enclave-manager/web/src/components/FileTree.tsx
```diff
@@ -0,0 +1,123 @@
...
+          }
+          onFileSelected={onFileSelected}
+        />
+      ))}
+    </Flex>
+  );
+};
+
+type FileTreeNodeComponentProps = {
+  node: FileTreeNode;
```
adschwartz, line 48: Would making this optional/`?` alleviate some of the checking and typecasting in the `FileTreeNodeComponent` and `DirectoryNode`? If not, would love to learn why this pattern is better than having an optional member of the `FileTreeNodeComponentProps`

enclave-manager/web/src/components/FileTree.tsx
```diff
@@ -0,0 +1,123 @@
...
+          selectedFilePath={childSelectedFilePath}
+          _isChildNode
+        />
+      )}
+    </>
+  );
+};
+
+const FileNode = ({ node, selectedFilePath, onFileSelected }: FileTreeNodeComponentProps) => {
+  const isSelected = isDefined(selectedFilePath) && selectedFilePath.length === 1 && selectedFilePath[0] === node.name;
```
adschwartz, line 108: maybe make this the same function as on line 70 (avoiding duplication)?

adschwartz (approved):

enclave-manager/web/src/components/FileTree.tsx
Dartoxian: Ah thanks for asking. Here I've chosen to keep `node` mandatory because this component is rendering a `FileTreeNode` - specifically it's deciding whether to render a directory node or a file node. It is never the case that the `node` is optional.

I agree the casting in `FileTreeNodeComponent` is not ideal - it's needed because typescript is not able to infer parent types (ie the type of `props`) based on the type of child types (ie the type of `childNodes` tells us the type of `node`). https://github.com/microsoft/TypeScript/issues/42384. An alternative would be rather than using our generic `isDefined` typeguard (which is a generally useful utility) we could implement a typeguard specificly to assert `FileTreeNodeComponentProps & { node: { childNodes: FileTreeNode[] } }`

enclave-manager/web/src/components/FileTree.tsx
Dartoxian: yep great flag

Dartoxian: sorry just looked at this again - these are actually checking different things, in this case it's checking `is this node selected`, in the `DirectoryNode` case it's checking `is a child file selected`

PR #1850, opened 2023-11-27
fix: tasks remove containers after they are done

### LINKED ISSUE #1846: Task containers are left running after the task completes
### What's your CLI version?

0.85.33

Pretty sure this is just a holdover from tasks v0 being implemented quickly. The actual cause is that we use `tail -f /dev/null` for the container command (I'm not actually sure why).

### Desired behavior

1. We don't use `tail -f /dev/null` (which is also swallowing the task logs
2. We actually run the container using what the user passed in, so we get logs the way we expect
3. Task containers stop as you'd expect

I suspect that the `tail -f /dev/null` way was done due to a misunderstanding of the way the `store` command needs to work, and we're currently:
1. staring the container with `tail -f /dev/null`
1. exec'ing the user's command on the container
2. running a `tar` to capture files artifacts defined in `store`

What I believe should actually happen:
1. we examine the `store` specs before starting anything, and preemptively create volumes at the `store` mountpoints before the container even starts
2. we run the task with the volumes mounted into the appropriate place
3. we run a container afterwards that mounts the same volumes, packages them up as files artifacts, and sends them to the APIC
4. optionally, we can also choose to manipulate how the files in the volume get tar'd up (e.g. maybe the user wants to grab only a single file from the entire mounted directory)

h4ck3rk3y:
## Description:
Closes #1846 

YES

This removes task containers after the task is "done"

### DIFF
--- a/core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_python.go
+++ b/core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_python.go
@@ -274,12 +274,10 @@ func (builtin *RunPythonCapabilities) Validate(_ *builtin_argument.ArgumentValue
 	return validateTasksCommon(validatorEnvironment, builtin.storeSpecList, serviceDirpathsToArtifactIdentifiers, builtin.serviceConfig.GetContainerImageName())
 }
 
-// Execute This is just v0 for run_python task - we can later improve on it.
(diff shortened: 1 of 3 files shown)


h4ck3rk3y:
before/after

image
image

leoporoli (approved):

PR #1851, opened 2023-11-27
fix: make run_python not create additional files artifact

h4ck3rk3y:
## Description:
we'd create a files artifact which would

1. make interpret depend on service network - an anti pattern
2. bloat enclave inspect

This fixes both; tested against etheruem-package


h4ck3rk3y:
after/before

image
image

leoporoli (approved):

PR #1852, opened 2023-11-27
feat: add string slice arguments to our parsed_flags library

omar711:
## Description:
Adds slice strings to our parsed_flags library.  This will let us specify flags for the upcoming static ports work, e.g.:

```
kurtosis run --connect web.http --connect db.postgresql
# or
kurtosis run --connect web.http,db.postgresql
```

NO


omar711:
sorry, this isn't fininshed, I'll reopen once set

cli/cli/command_framework/lowlevel/flags/flag_type_processors.go
```diff
@@ -45,6 +46,27 @@ func processStringFlag(
...
 
+func processStringSliceFlag(
+	flagKey string,
+	shorthand string,
+	defaultValueStr string,
+	usage string,
+	cobraFlagSet *flag.FlagSet,
+) error {
+
+	if defaultValueStr != "" {
```
omar711, line 57: There's an assumption in the library that defaults will always derive from strings.  

We could look to parse string representations of arrays, or add more genral support e.g. for `[]string` here.  

However, my thinking is defaults don't make much sense for a slice flag, so just make it clear it won't work, where somebody does specify a default.

barnabasbusa:

PR #1853, opened 2023-11-27
fix: deprecate ConnectServices, remove calls to it

omar711:
## Description:
We won't end up using `ConnectServices` so I'm removing the call to it first of all.

Once this calling code is no longer running out there, we can drop the protobuf definition and linked code.  Doing it all at once will probably break things post-upgrade, where people have pre-existing APICs running. 

NO

### DIFF
--- a/api/golang/core/lib/enclaves/enclave_context.go
+++ b/api/golang/core/lib/enclaves/enclave_context.go
@@ -429,6 +429,7 @@ func (enclaveCtx *EnclaveContext) GetAllFilesArtifactNamesAndUuids(ctx context.C
 }
 
 // Docs available at https://docs.kurtosis.com/sdk#connectservices
+// Deprecated: currently a no-op that we're removing
 func (enclaveCtx *EnclaveContext) ConnectServices(ctx context.Context, connect kurtosis_core_rpc_api_bindings.Connect) error {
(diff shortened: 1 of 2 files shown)


api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -429,6 +429,7 @@ func (enclaveCtx *EnclaveContext) GetAllFilesArtifactNamesAndUuids(ctx context.C
 }
 
 // Docs available at https://docs.kurtosis.com/sdk#connectservices
+// Deprecated: currently a no-op that we're removing
```
laurentluce, line 432: It is a no-op for Docker Cloud but not for the remote K8S use case where this call is relied on in the CLI Gateway.

omar711:
Closing this for now, will wait for Gateway's removal first.  

The reason we can't do this right away is https://github.com/kurtosis-tech/kurtosis/blob/75b8c844243a55bdf5ddbd653e229e64ac555e0b/cli/cli/kurtosis_gateway/server/api_container_gateway/api_container_gateway_service_server.go#L119 , which is where a flag is set that tells Gateway to restore connections and clean up obsolete ones, whenever calls are made to `GetServices`.

PR #1863, opened 2023-11-28
fix: service name collision error message

h4ck3rk3y:
The previous message was confusing and left @barnabasbusa bewildered. This is much more clearer and comes with a recommendation


barnabasbusa:
Approved ✅

leoporoli (approved):

PR #1864, opened 2023-11-28
fix: support logs db for k8s

tedim52:
## Description:
`service logs` over kubernetes was mistakenly turned off when centralized logging over docker was implemented. This change turns it back on.

YES (users will be able to do `service logs` over k8s)

### DIFF
--- a/engine/server/engine/main.go
+++ b/engine/server/engine/main.go
@@ -322,6 +322,7 @@ func getKurtosisBackend(ctx context.Context, kurtosisBackendType args.KurtosisBa
 	return kurtosisBackend, nil
 }
 
+// if cluster is docker, return logs client for centralized logging, otherwise use logs db of kurtosis backend which uses k8s logs under the hood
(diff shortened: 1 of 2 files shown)


internal_testsuites/golang/testsuite/stream_logs_test/stream_logs_test.go
```diff
@@ -1,8 +1,3 @@
-//go:build !kubernetes
```
h4ck3rk3y: can we also add a test in the circle ci yaml to docker / k8s basic CLI tests

h4ck3rk3y (approved):

internal_testsuites/golang/testsuite/stream_logs_test/stream_logs_test.go
tedim52: yep

tedim52: seems like k8s basic cli tests are turned off right now due to inconsistencies like this onehttps://github.com/kurtosis-tech/kurtosis/issues/1866. I'm going to merge this to get the fix out ASAP and make a note to resolve the CI dicrepancy

PR #1865, opened 2023-11-28
feat: emui catalog overview

Dartoxian:
## Description:
This PR implements the first pass of the emui catalog. Additionally:
* It has a minor tweak to log viewer button colours
* Refactors the KurtosisBreadcrumbs to no longer be dependent on being on an `enclaves` route
* Fixes the application max width to be 1440px, not the 1320px I had incorrectly set it to.

### Demo

https://github.com/kurtosis-tech/kurtosis/assets/4419574/dfc5a489-3c6f-44d0-b27f-ca4d7d61786f

YES

* Figma


enclave-manager/web/src/emui/catalog/Catalog.tsx
```diff
@@ -1,33 +1,150 @@
...
+  const isSearching = searchTerm.length > 0;
+  const filteredCatalog = useMemo(
+    () => catalog.packages.filter((kurtosisPackage) => kurtosisPackage.name.toLowerCase().indexOf(searchTerm) > 0),
+    [searchTerm, catalog],
+  );
+
+  useEffect(() => {
+    const listener = function (e: KeyboardEvent) {
+      const element = searchRef?.current;
+      if ((e.ctrlKey && e.keyCode === 75) || (e.metaKey && e.keyCode === 75)) {
```
adschwartz, line 59: @Dartoxian wondering if we can centralize this code so we have one place where we define how to hit ctrl/command F?

enclave-manager/web/src/emui/catalog/storage.ts
```diff
@@ -0,0 +1,26 @@
...
+import { isDefined, stringifyError } from "../../utils";
+
+const SAVED_PACKAGES_LOCAL_STORAGE_KEY = "kurtosis-saved-packages";
+
+export const storeSavedPackages = (kurtosisPackages: KurtosisPackage[]) => {
+  localStorage.setItem(
+    SAVED_PACKAGES_LOCAL_STORAGE_KEY,
+    JSON.stringify(kurtosisPackages.map((kurtosisPackage) => kurtosisPackage.name)),
+  );
+};
```
adschwartz, line 11: @Dartoxian what do you think about adding version number to this storage object. Like `"version":1` or something? Just so that we can later on migrate people if we make a breaking change to this data?

adschwartz (approved):

enclave-manager/web/src/emui/catalog/Catalog.tsx
Dartoxian: yep good point - we just want a `useCtrlKeyListener` type of hook. I'll add

enclave-manager/web/src/emui/catalog/storage.ts
Dartoxian: If this is the long term storage layer I'd agree - though I think from this https://sageroomworkspace.slack.com/archives/C05DZ4UABDZ/p1701104638634789?thread_ts=1701103387.278829&cid=C05DZ4UABDZ the plan is to move this storage to the backend anyway.

PR #1870, opened 2023-11-29
fix: move log collector creation logic

### LINKED ISSUE #1833: Resource-leaked `logs-collector-XXXX` containers
### What's your CLI version?

0.85.29

My local Docker-backed Kurtosis is empty:
```
kurtosis enclave ls
UUID   Name   Status   Creation Time
```

`kurtosis clean -a` doesn't clean anything, as expected.

However, there are two zombie logs-collector containers hanging around that don't go away:

Screen Shot 2023-11-19 at 13 27 21

These do not correspond to any enclave networks, as there are none (as expected):

```
$ docker network ls
NETWORK ID     NAME      DRIVER    SCOPE
8662e507c238   bridge    bridge    local
deab0a0cb95e   host      host      local
a8c59496b516   none      null      local
```

More details in #1832 

### Desired behavior

Kurtosis should _always_ fully clean up after itself.

The original `clean` logic was written like so:
1. Use the presence/absence of `-a` to decide which enclaves ought to _remain_ (if `-a` is specified, no resources should remain)
2. After finding the enclaves-to-remain, find and kill all Kurtosis resources that aren't in the enclave-to-remain set in the proper order (containers -> volumes -> networks). We DON'T say "look for enclave networks, and then remove resources associated with them" because if the enclave network is removed without all the resources being removed for any reason (e.g. maybe a failure happened), then we permanently leak resources.

In other words, we first grab _ALL_ Kurtosis resources, default them all to "destroy this resource", and then remove the ones that should be preserved. This guarantees that we process every single resource Kurtosis can create.

This algorithm is very important, because it guarantees that even if an enclave is in an inconsistent state (e.g. an error happened, or maybe the user messed with Docker), all the resources for the enclave can still get fully cleaned up.

I suspect that the `logs-collector` cleanup has not been written this way, and is instead written with, "first try to remove the enclave network and all the containers in it, then try to remove the logs-collector". This means that if a failure happens (e.g. because of Ctrl-C), then the enclave network is lost but the logs-collector hasn't been cleaned up.

tedim52:
## Description:
Fixes: https://github.com/kurtosis-tech/kurtosis/issues/1833

The resource leak was being caused by a state where if a failure occurred while creating an enclave (eg. ctrl+C while running `kurtosis enclave add`) at a point where the log collector container was created BEFORE it had been connected to the network AND before the necessary `defer undo`s to clean up the log collector had been queued THEN the network would get cleaned up by the `defer undo` from `CreateNetwork`, but the log collector container would remain.

Any attempt to do a `kurtosis clean -a` would fail to clean log collector container because the network(and thus enclave) the container was created for had already been cleaned up.

After digging - I realized the log collector was being created in `container-engine-lib` as opposed to in the `engine` Moving the logic to the engine AFTER the enclave is created fixes the issue. If there is an error at ANY point in the creation of the log collector container (even if the log collectors `defer undo` hasn't been queued), the `defer undo` from the `CreateEnclave` call will clean the log collector because it has a label with the `enclaveUUID`.

NO

## References:
https://github.com/kurtosis-tech/kurtosis/issues/1833

### DIFF
--- a/engine/server/engine/enclave_manager/enclave_creator.go
+++ b/engine/server/engine/enclave_manager/enclave_creator.go
@@ -83,13 +83,12 @@ func (creator *EnclaveCreator) CreateEnclave(
 	}()
 
 	// only create log collector for backend as
-	var shouldDeleteLogsCollector bool
+	shouldDeleteLogsCollector := true
 	if kurtosisBackendType == args.KurtosisBackendType_Docker {
 		// TODO the logs collector has a random private ip address in the enclave network that must be tracked
 		if _, err := creator.kurtosisBackend.CreateLogsCollectorForEnclave(setupCtx, enclaveUuid, defaultTcpLogsCollectorPortNum, defaultHttpLogsCollectorPortNum); err != nil {
 			return nil, stacktrace.Propagate(err, "An error occurred creating the logs collector with TCP port number '%v' and HTTP port number '%v'", defaultTcpLogsCollectorPortNum, defaultHttpLogsCollectorPortNum)
 		}
-		shouldDeleteLogsCollector = true
 		defer func() {
 			if shouldDeleteLogsCollector {
 				err = creator.kurtosisBackend.DestroyLogsCollectorForEnclave(teardownCtx, enclaveUuid)
@@ -113,7 +112,6 @@ func (creator *EnclaveCreator) CreateEnclave(
 		cloudUserID,
 		cloudInstanceID,
 	)
-
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "An error occurred launching the API container")
 	}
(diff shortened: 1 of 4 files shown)


engine/server/engine/enclave_manager/enclave_creator.go
```diff
@@ -75,6 +82,24 @@ func (creator *EnclaveCreator) CreateEnclave(
 		}
 	}()
 
+	// only create log collector for backend as
+	var shouldDeleteLogsCollector bool
```
adschwartz: would it make sense to set `shouldDeleteLogsCollector` to `true` already here? Wondering if it optentially opens up to bugs to only set it later? I don't see any bugs now, but waiting to set it could potentially leave it default to `false`

adschwartz (approved):

engine/server/engine/enclave_manager/enclave_creator.go
tedim52: good idea

Issue #1843, opened 2023-11-22
Ensure no logs are truncated/dropped when services are restarted


tedim52:
### What's your CLI version?

0.85.34

### Description & steps to reproduce

We observed that the first few logs of the Hyperlane relayer were being truncated when it was misconfigured misconfigured and being restarted constantly by Docker.

### Desired behavior

1. The logs stream is to the UI is continuous in the case of failure/restarting.
2. No logs are being dropped as the service goes through failure/restarting.

### What is the severity of this bug?

Critical; I am blocked and Kurtosis is unusable for me because of this bug.

### What area of the product does this pertain to?

Frontend: the Graphical User Interface, including the Enclave Manager

tedim52:
Changing the tail package and changing download logs button to pull all logs from engine (as opposed to logs in the UI) fixes this issue.

PR #1874, opened 2023-11-30
fix: change tail package

tedim52:
## Description:
When logs were being produced at a rate of greater than 200-300 logs per second, the `nxadm/tail` would be returning incomplete lines, causing parsing errors on tailed logs. Switching to using `hpcloud/tail` fixes this completely when I ran several tests with >1000 logs per second. (the apis are identically and `hpcloud/tail` seems to be more widely used) In case this package also fails, I've changd the logic return err immediately.

NO

## References
https://github.com/kurtosis-tech/kurtosis/issues/1843


engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
```diff
@@ -67,12 +67,12 @@ func (manager *LogFileManager) StartLogFileManagement(ctx context.Context) {
 
 		logrus.Debugf("Scheduling log file path creation every '%v' minutes...", volume_consts.CreateLogsWaitMinutes)
 		for range logFileCreatorTicker.C {
-			logrus.Debug("Creating log file paths...")
+			logrus.Trace("Creating log file paths...")
```
tedim52, line 70: made this change bc these logs were polluting engine logs

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -372,6 +371,14 @@ func (strategy *PerWeekStreamLogsStrategy) followLogs(
 	}
 
 	for logLine := range logTail.Lines {
+		if logLine.Err != nil {
+			return stacktrace.Propagate(logLine.Err, "hpcloud/tail encountered an error with the following log line: %v", logLine.Text)
+		}
+		if !isValidJsonString(logLine.Text) {
+			// if tail package fails to parse a valid new line, fail fast
+			return stacktrace.NewError("hpcloud/tail returned the following incomplete line: '%v' that was not valid json.\nThis is a bug in tailing package!", logLine.Text)
```
adschwartz: is it possible there are other causes though?
"This is a bug" => "This is potentially a bug"

adschwartz: Optional change ^

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
```diff
@@ -380,6 +387,11 @@ func (strategy *PerWeekStreamLogsStrategy) followLogs(
 	return nil
 }
 
+func isValidJsonString(line string) bool {
+	var jsonLog JsonLog
+	return json.Unmarshal([]byte(line), &jsonLog) == nil
+}
```
adschwartz, line 398: Is it going to be expensive to do this test on every line that we process (thinking about the memory and cpu resources that are spent)?

adschwartz: Will this function be called for each line that is logged instantly or does it only run rarely?

adschwartz (changes requested):
@tedim52 overall ok, but I have a question regarding the cost of testing the JSON string

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
tedim52: done

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/per_week_stream_logs_strategy.go
tedim52: Hm good point - we have to do this operation anyways to process the message (extract the log message field from the log line). I could refactor so that this call is only being made once (as opposed to 2x per line)

tedim52: If we find that making this call even once per log line is inefficient, in the future can optimize to unmarshal several json log lines at once.

adschwartz (approved):

PR #1878, opened 2023-12-01
docs: remove cloud early access signup form

leeederek:
## Description:
Cloud is real now and works amazingly. This PR replaces mentions of an early sign-up form for Beta access to Cloud with an actual link to cloud.kurtosis.com. 

The old form was made and put in our docs to collect early feedback and design partners in the absence of an actual product for people to try. We now have a real cloud offering for people to try.

YES

https://kurtosistech.slack.com/archives/C051H93B6DV/p1701435264644419

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Check out an introductory demo video here:
 
 How do I get going?
 ===================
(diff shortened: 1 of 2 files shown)


chunha-park (approved):
Approved!

PR #1889, opened 2023-12-04
docs: add starlark writing tips to more places in the docs

leeederek:
## Description:
We have fielded questions from many users about tips and tricks to make writing and reading Starlark easier. This PR adds our top 3 Starlark tips (VSCode plugin, Vim syntax highlighting and Neovim syntax highlighting) to various parts of our docs. Namely:
- Best Practices section
- Get started section
- Expanded the tip box in our "write your own package" quickstart
- Starlark introduction page

YES

https://discord.com/channels/783719264308953108/1131048810861314169/1180239655376015360

### DIFF
--- a/docs/docs/get-started/get-started.md
+++ b/docs/docs/get-started/get-started.md
@@ -80,24 +80,3 @@ kurtosis run github.com/kurtosis-tech/basic-service-package \
 ![service-c-partying.png](/img/home/service-c-k8s.png)
  
 </details>
-
-Writing and reading Starlark
-----------------------------
-The Starlark programming language is used to manipulate the contents of environments spun up with Kurtosis. Read more about Starlark and why we chose to use it [here](../advanced-concepts/why-kurtosis-starlark.md).
-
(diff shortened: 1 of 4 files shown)


docs/docs/get-started/get-started.md
```diff
@@ -79,4 +79,25 @@ kurtosis run github.com/kurtosis-tech/basic-service-package \
 
 ![service-c-partying.png](/img/home/service-c-k8s.png)
  
-</details>
\ No newline at end of file
+</details>
```
galenmarchetti: @leeederek  I don't think this should go in the `get-started` section. its really important that we keep this as minimal as possible - strong opinion that we should keep this particular file the same, but include it in other places so its findable on the search bar.

galenmarchetti (changes requested):

docs/docs/get-started/get-started.md
leeederek: 👍🏻 ok, makes sense. I've removed it.

galenmarchetti (approved):

PR #1891, opened 2023-12-05
feat: emui handle json/yaml input interchangably

Dartoxian:
## Description:
This PR adds support for using `yaml` in the `json` code editor input boxes. Demo below

https://github.com/kurtosis-tech/kurtosis/assets/4419574/405e72fe-363e-4cb2-8cf7-039d9e7c0c6b

YES

* Discussed on slack.


enclave-manager/web/src/components/CodeEditor.tsx
```diff
@@ -57,37 +62,65 @@ export const CodeEditor = forwardRef<CodeEditorImperativeAttributes, CodeEditorP
...
+              const formatAction = editor.getAction("editor.action.formatDocument");
+              assertDefined(formatAction, `Format action is not defined`);
+              await formatAction.run();
+            }
+          };
+
+          if (isReadOnly) {
+            return new Promise((resolve) => {
+              const listenerDisposer = editor.onDidChangeConfiguration((event) => {
+                if (event.hasChanged(89 /* ID of the readonly option */)) {
```
adschwartz: @Dartoxian heads up: I believe this number changes/has changed over time with new versions of monaco. In the old EM UI I made this a constant (fwiw we are big on extracting our magic numbers) and I think we should extract this out as a setting so it can more easily be found and has some documentation for us to quickly identify.

adschwartz (approved):
Approved but with one comment to look at

enclave-manager/web/src/components/CodeEditor.tsx
Dartoxian: thanks @adschwartz - will do. 

What's weird is that according to the definition in the enum types this isn't even the right option - in my local types the `EditorOption` with value `89` is actually `quickSuggestionsDelay`!

adschwartz: @Dartoxian Yep, I noticed the same discrepancy as I looked at the source code. I believe I came to the conclusion that the actual version of monaco is not the version we see in the project source code. As I dug deeper I realized that these enum numbers have changed over time (which by itself seems like a terrible practice and leads to lots of pain for users of that API)

PR #1895, opened 2023-12-05
feat: service logs full download

Dartoxian:
## Description:
This PR implements the full streamed download of service logs. This is based on some work already done by @tedim52  in tedi/copydownload.

Changes:
* The copy logs button is no longer available on the Service Logs page
* The download logs button sends the log stream to a stream writer to download them as a file.

Concerns:
* For services that log excessively, like the log spammer, it seems downloading the logs is incredibly slow - <1kb/s. I'm not sure what could improve this, but if the backend streamed more than one record per message (as seems to be possible by the api spec) that seems like it would help (and would help improve performance when rendering service logs in the ui).

### Demo

https://github.com/kurtosis-tech/kurtosis/assets/4419574/61716034-fb6e-47c1-b776-e8c262b40799

YES


enclave-manager/web/src/components/DownloadButton.tsx
```diff
@@ -1,10 +1,11 @@
 import { Button, ButtonProps, IconButton, IconButtonProps } from "@chakra-ui/react";
 import { FiDownload } from "react-icons/fi";
-import { isDefined } from "../utils";
+import streamsaver from "streamsaver";
```
adschwartz, line 3: @Dartoxian do you need to add this guy to `package.json`?

Dartoxian: no it's already there.

enclave-manager/web/src/components/DownloadButton.tsx
```diff
@@ -17,9 +18,20 @@ export const DownloadButton = <IsIconButton extends boolean>({
   isIconButton,
   ...buttonProps
 }: DownloadButtonProps<IsIconButton>) => {
-  const handleDownloadClick = () => {
+  const handleDownloadClick = async () => {
     if (isDefined(valueToDownload)) {
       const v = typeof valueToDownload === "string" ? valueToDownload : valueToDownload();
+
+      if (isAsyncIterable(v)) {
+        const writableStream = streamsaver.createWriteStream(fileName);
```
adschwartz, line 26: @Dartoxian I think there's a good chance this may work fine, but I wonder if you've tried downloading from your cloud instance? It's possible this might not work going over the proxies (without some tweaking) but it's possible it will.

adschwartz (approved):

enclave-manager/web/src/components/DownloadButton.tsx
Dartoxian: Yep I've verified that the `streamsaver` library works in the cloud as it is also used to download artifact tgz files in the cloud.

PR #1897, opened 2023-12-05
fix: emui small design tweaks based on team feedback

Dartoxian:
## Description:
This PR includes a number of the easier emui tweaks from the december burndown:
* Don't show copy button next to the package source
image
* Make the delete button have a hover
image
* ellipsis long enclave names on the enclaves list
image
* Don't include entrypoint and returns on the catalog package page
* Dynamic package card layout
image
* support navigating logs with enter and shift+enter when searching
* fix an issue with log search that was causing excessive renders, which I think was causing the `ok.` bug mentioned on notion.

YES

- December burndown notion


enclave-manager/web/src/components/enclaves/logs/LogViewer.tsx
```diff
@@ -247,7 +249,7 @@ const SearchControls = ({ searchState, onChangeSearchState, logLines }: SearchCo
     [logLines, onChangeSearchState],
   );
 
-  const debouncedUpdateMatches = useMemo(() => debounce(updateMatches, 100), [updateMatches]);
+  const debouncedUpdateMatches = useMemo(() => debounce(updateMatches, 300), [updateMatches]);
```
adschwartz: @Dartoxian I would prefer a quicker update, but wondering if you hit something that made 100ms infeasible. Let me know. Else I think faster updating, even as you type is a better experience

adschwartz (approved):

enclave-manager/web/src/components/enclaves/logs/LogViewer.tsx
Dartoxian: yep I was finding that with the 100ms speed I was encountering a set state limit error with react. Maybe what would be more appropriate here, and would feel more responsive would be a `throttle` rather than a `debounce`. I'll implement that on Thursday.

adschwartz: ok makes sense. Thanks!

Dartoxian: `throttle` is working a lot better than `debounce` in this situation

PR #1871, opened 2023-11-29
feat: User service ports Traefik Docker labels

laurentluce:
## Description:
As part of the authenticated public http ports feature, we are bringing Traefik inside the Docker cluster to route HTTP traffic to the user service HTTP ports.  This PR sets the required Docker labels.

This was tested with Traefik running inside the engine enclave with the following static config and the user service enclave network added to the Traefik container list of networks.

```
version: '3'

services:
  reverse-proxy:
    image: traefik:v2.10
    # Enables the web UI and tells Traefik to listen to docker
    command:
      - --accesslog=true
      - --api.debug=true
      - --api.insecure=true
      - --api.dashboard=true
      - --api.disabledashboardad=true
      - --providers.docker
      - --entrypoints.web.address=:8000
      - --providers.docker.network=bridge
      - --providers.docker.exposedByDefault=false
      - --log.level=DEBUG
    ports:
      # The HTTP port
      - "8000:8000"
      # The Web UI (enabled by --api.insecure=true)
      - "8080:8080"
    volumes:
      # So that Traefik can listen to the Docker events
      - /var/run/docker.sock:/var/run/docker.sock
```

User service "nginx" port labels:
Enclave short UUID: 65d2fb6d6732
Service short UUID: 3771c85af16a
HTTP Port number: 80
```
"traefik.enable": "true"
"traefik.http.routers.65d2fb6d6732-3771c85af16a-80.rule": "Host(`80-3771c85af16a-65d2fb6d6732`)"
"traefik.http.routers.65d2fb6d6732-3771c85af16a-80.service": "65d2fb6d6732-3771c85af16a-80"
"traefik.http.services.65d2fb6d6732-3771c85af16a-80.loadbalancer.server.port": "80"
```

```
$ curl -I http://localhost:8000 -H "Host: 80-3771c85af16a-65d2fb6d6732"
HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Length: 615
Content-Type: text/html
Date: Wed, 29 Nov 2023 21:32:51 GMT
Etag: "6537cac7-267"
Last-Modified: Tue, 24 Oct 2023 13:46:47 GMT
Server: nginx/1.25.
```

## Is this change user facing?
NO

## References (if applicable):
https://www.notion.so/kurtosistech/Public-user-service-HTTP-ports-bdf1107b0d1c4ca990c346fd87473174

PR #1906, opened 2023-12-06
feat: Reverse proxy lifecycle management and connectivity on Docker

laurentluce:
## Description:
In #1871 we added the Traefik labels to the user services so Traefik can discover them and route to them.  This PR implements a reverse proxy using Traefik on Docker.  The engine starts and stops Traefik.  The Traefik container is automatically connected to the enclave networks.  The implementation leverages what was done for the logs aggregator.

YES

#1871


.circleci/config.yml
```diff
@@ -80,14 +80,19 @@ steps_prepare_testing_k8s_k3s: &steps_prepare_testing_k8s_k3s
         command: "${KURTOSIS_BINPATH} gateway"
         background: true
 
-# Steps to prepare a job for Docker testing
```
laurentluce, line 83: Refactoring to share those steps a bit easier.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
```diff
@@ -238,11 +239,21 @@ func getDockerKurtosisBackend(
...
+		if !found {
+			return nil, stacktrace.NewError("An error occured while getting the reverse proxy enclave network IP address for enclave '%v', This is a bug in Kurtosis", enclaveUuid)
+		}
+
 		alreadyTakenIps := map[string]bool{
 			networkIp.String():      true,
 			network.GetGatewayIp():  true,
 			apiContainerIp.String(): true,
 			logsCollectorObj.GetEnclaveNetworkIpAddress().String(): true,
+			reverseProxyEnclaveNetworkIpAddress.String(): true,
```
laurentluce: Add the reverse proxy enclave network private ip to the list of already taken ips so the user service private ips don't conflict.

.circleci/config.yml
```diff
@@ -639,6 +658,8 @@ jobs:
   test_old_enclave_continuity:
     executor: ubuntu_vm
     steps:
+      - <<: *abort_job
```
laurentluce, line 662: Disable this test until this change is released since both previously released and being tested CLI need to be aware of the reverse proxy network.

container-engine-lib/lib/backend_interface/objects/reverse_proxy/reverse_proxy_config.go
laurentluce, line 1: Add as much as possible to the backend interface so it can be re-used in the K8S implementation.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_enclave_functions.go
```diff
@@ -357,15 +372,33 @@ func (backend *DockerKurtosisBackend) DestroyEnclaves(
 		erroredEnclaveUuids[enclaveUuid] = volumeRemovalErr
 	}
 
+	// Disconnect the containers from the enclave networks being removed
```
laurentluce: We keep the disconnect networks logic generic so other types of containers (not just reverse proxy containers) can leverage it.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -471,6 +474,57 @@ func (backend *DockerKurtosisBackend) DestroyLogsCollectorForEnclave(ctx context
...
+		backend.objAttrsProvider,
+	)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred creating the reverse proxy using the reverse proxy container '%+v'.", reverseProxyContainer)
+	}
+	return reverseProxy, nil
+}
+
+func (backend *DockerKurtosisBackend) GetReverseProxy(ctx context.Context) (*reverse_proxy.ReverseProxy, error) {
+	maybeReverseProxy, err := reverse_proxy_functions.GetReverseProxy(
```
tedim52, line 494: is it possible that `maybeReverseProxy` is `nil`?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
```diff
@@ -73,14 +74,24 @@ func (backend *DockerKurtosisBackend) CreateAPIContainer(
...
 	if err != nil {
-		return nil, stacktrace.Propagate(err, "An error occurred while getting the logs collector for enclave '%v; This is a bug in Kurtosis'", enclaveUuid)
+		return nil, stacktrace.Propagate(err, "An error occurred while getting the logs collector for enclave '%v'; This is a bug in Kurtosis", enclaveUuid)
+	}
+
+	reverseProxy, err := backend.GetReverseProxy(ctx)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred while getting the reverse proxy, This is a bug in Kurtosis")
+	}
+	reverseProxyEnclaveNetworkIpAddress, found := reverseProxy.GetEnclaveNetworksIpAddress()[enclaveNetwork.GetId()]
```
tedim52, line 87: if `reverseProxy` is `nil` this call will fail , could bet hat your making a check later in code but haven't got there yet

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_enclave_functions.go
```diff
@@ -357,15 +372,33 @@ func (backend *DockerKurtosisBackend) DestroyEnclaves(
...
+	networksToDisconnect := map[enclave.EnclaveUUID]string{}
+	for enclaveUuid := range successfulVolumeRemovalEnclaveUuids {
+		networkInfo, found := matchingNetworkInfo[enclaveUuid]
+		if !found {
+			return nil, nil, stacktrace.NewError("Would have attempted to disconnect enclave network '%v' that didn't match the filters", enclaveUuid)
```
tedim52: If this error is hit, likely indicates a bug is in Kurtosis, can we add that? eg. `Attempt was made to disconnect enclave that did not match filters. This is likely a bug in Kurtosis.`

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_enclave_functions.go
tedim52: This makes sense, but for right now the only one that's being disconnected is the reverse proxy container right? Also - specifically this is a special type of containers - it containers that weren't in the enclave, but were connected to the enclaves network right?

If I'm right, I wonder if there's a way to be a bit more explicit about that fact bc this logic can be a bit misleading as it goes `destroyContainersInEnclaves` before `disconnectContainersFromEnclaveNetworks`.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
tedim52: I jus realized this is the same way we do it in logs aggregator logic

tedim52 (approved):

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_enclave_functions.go
laurentluce: I did some renaming to indicate we are disconnecting external containers from the enclave networks.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
laurentluce: Added a check.

PR #1928, opened 2023-12-12
fix: add ability to pass down storage class to apic/engine

h4ck3rk3y:
## Description:
We'd ignore storage class options now we have it available in the kubernets_manager

This will be used in upcoming PV/PVC prs

YES


h4ck3rk3y:
Closing in favor of #1943

PR #1941, opened 2023-12-13
feat: User service K8S ingresses for reverse proxy routing

laurentluce:
## Description:
Similar to how we are adding Traefik labels for user services on Docker, we are adding a K8S ingress for each user service with a path per HTTP port so HTTP traffic can be routed to them.

NO

https://github.com/kurtosis-tech/kurtosis/pull/1871


container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/shared_helpers/shared_helpers.go
```diff
@@ -361,13 +366,52 @@ func GetUserServiceKubernetesResourcesMatchingGuids(
...
+		serviceUuid := service.ServiceUUID(serviceGuidStr)
+
+		numIngressesForGuid := len(kubernetesIngressForGuid)
+		if numIngressesForGuid == 0 {
+			// This would indicate a bug in our ingress retrieval logic because we shouldn't even have a map entry if there's nothing matching it
+			return nil, stacktrace.NewError("Got entry of result ingresses for service GUID '%v', but no Kubernetes ingress were returned; this is a bug in Kurtosis", serviceUuid)
+		}
+		if numIngressesForGuid > 1 {
+			return nil, stacktrace.NewError("Found %v Kubernetes ingresses associated with service GUID '%v'; this is a bug in Kurtosis", numIngressesForGuid, serviceUuid)
+		}
```
lostbean: nit: we could combined both check:
```suggestion
		if numIngressesForGuid != 1 {
			return nil, stacktrace.NewError("Found %v Kubernetes ingresses associated with service GUID '%v', but number of ingresses should be exactly 1", numIngressesForGuid, serviceUuid)
		}
```

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/start_user_services.go
```diff
@@ -423,6 +433,43 @@ func createStartServiceOperation(
...
+				return nil, stacktrace.Propagate(err, "An error occurred creating ingress for service with UUID '%v'", serviceUuid)
+			}
+			shouldDestroyIngress = true
+			defer func() {
+				if !shouldDestroyIngress {
+					return
+				}
+				if err := kubernetesManager.RemoveIngress(ctx, createdIngress); err != nil {
+					logrus.Errorf("Starting service didn't complete successfully so we tried to remove the ingress we created but doing so threw an error:\n%v", err)
+					logrus.Errorf("ACTION REQUIRED: You'll need to remove ingress '%v' in '%v' manually!!!", ingressName, namespaceName)
```
lostbean, line 468: What does happen if no action is taken? Do we need to escalate this to the user and/or platform admin? Otherwise this might get buried in the logs

lostbean (approved):
Just one small comment and one question but LGTM

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/start_user_services.go
laurentluce: Those messages are displayed back to the user.

PR #1943, opened 2023-12-13
feat: persistent volumes work on multi node k8s clusters

h4ck3rk3y:
feat: use storage-class in the kurtosis-config.yaml; users should make sure that they have the right storage class configured in their `kurtosis-config.yaml`


h4ck3rk3y:
TODO

- [x] deletion

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/persistent_directories.go
```diff
@@ -12,14 +12,12 @@ import (
 )
 
 type kubernetesVolumeWithClaim struct {
```
h4ck3rk3y, line 14: Change Default size to 1g as thats minimum for. the dynamic provisioning drivers

leoporoli: good

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/persistent_directories.go
```diff
@@ -96,16 +93,7 @@ func preparePersistentDirectoriesResources(
...
-		if persistentVolume, err = kubernetesManager.GetPersistentVolume(ctx, volumeName); err != nil {
-			persistentVolume, err = kubernetesManager.CreatePersistentVolume(ctx, namespace, volumeName, volumeLabelsStrs, persistentVolumeSize)
-			if err != nil {
-				return nil, stacktrace.Propagate(err, "An error occurred creating the persistent volume for '%s'", persistentDirectory.PersistentKey)
-			}
-			volumesCreated[persistentVolume.Name] = persistentVolume
-		}
-
-		// For now, we have a 1:1 mapping between volume and volume claims, so it's fine giving it the same name
+		// This claim works with a dynamic driver - it will spin up its own volume - the volume will get deleted when said claims is deleted
```
leoporoli, line 96: just out of curiosity, Did you test this with Minikube? Does Minikube create a hostPath volume?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/persistent_directories.go
```diff
@@ -136,17 +123,6 @@ func preparePersistentDirectoriesResources(
 				logrus.Warnf("You'll need to clean up volume claim '%v' manually!", volumeClaimNameStr)
 			}
 		}
-		for volumeNameStr := range volumesCreated {
-			// Background context so we still run this even if the input context was cancelled
-			if err := kubernetesManager.RemovePersistentVolumeClaim(context.Background(), namespace, volumeNameStr); err != nil {
```
leoporoli, line 141: it was wrong right? because it was deleting a PVC instead of a PV, well it doesn't matter because now it is not necessary

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
```diff
@@ -327,119 +321,6 @@ func (manager *KubernetesManager) GetIngressesByLabels(ctx context.Context, name
...
-	}
-
-	volume, err := volumesClient.Create(ctx, &persistentVolumeDefinition, globalCreateOptions)
-	if err != nil {
-		return nil, stacktrace.Propagate(err, "Failed to create volume '%s'", volumeName)
-	}
-	return volume, err
-}
-
 func (manager *KubernetesManager) RemovePersistentVolume(
```
leoporoli: shouldn't this one also be removed? if the creation is removed

leoporoli: same for `GetPersistentVolume`

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
```diff
@@ -561,13 +442,7 @@ func (manager *KubernetesManager) CreatePersistentVolumeClaim(
 		return nil, stacktrace.Propagate(err, "Failed to create volume claim '%s'", volumeClaimName)
 	}
 
-	// Wait for the PVC to become bound and return that object (which will have VolumeName filled out)
-	boundVolumeClaim, err := manager.waitForPersistentVolumeClaimBinding(ctx, namespace, volumeClaim.Name)
```
leoporoli, line 565: why you are removing this? is it, not usable to determine that the volume is created an can be used?

leoporoli (approved):
approved with some comments. The most important is the one related to `waiting for pv binding` is it not necessary to know that the cluster provider creates the PV? or how Kurtosis detects that the cluster provider (DO, for instance) creates the PV?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/persistent_directories.go
h4ck3rk3y: Yes, it was wrong!

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
h4ck3rk3y: @leoporoli We are using dynamic provisioning

So I create PVC
I start pod with the PVC mount;

This is when the PV will automatically be created by k8s; so the claim will never be bound; as this validation would happen before the service is created

h4ck3rk3y: The service will fail to start if the PV doesn't get created

leoporoli:
> Note this enforces usage of `storage-class`. Is this a breaking change?

@h4ck3rk3y I think if users run `kurtosis engine restart` and it returns an error because they don't have the storage class defined it will be a breaking change, but if it's well described in the changelogs I think we could release it before the BCH date because the affectation is low, at least this is my perspective.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
h4ck3rk3y: removed them and a bunch of other code(pre Guillaume!!)

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/persistent_directories.go
h4ck3rk3y: Minikube uses `standard` not `local-path` I have fixed the doc I mistakenly editted. It should support dynamic provisioning https://platform9.com/blog/tutorial-dynamic-provisioning-of-persistent-storage-in-kubernetes-with-minikube/

PR #1950, opened 2023-12-14
feat: Web UI migration to REST API 

lostbean:
## Description:
This PR migrates the web client to use the new REST API.

NO


enclave-manager/web/src/client/enclaveManager/KurtosisClient.ts
```diff
@@ -74,9 +91,47 @@ export abstract class KurtosisClient {
...
 
+  async *getServiceLogsWS(
+    abortController: AbortController,
+    enclave: RemoveFunctions<EnclaveFullInfo>,
+    serviceUUID: string,
+    followLogs?: boolean,
+    numLogLines?: number,
+    returnAllLogs?: boolean,
+    conjunctiveFilters?: LogLineFilter[]
+  ): AsyncGenerator<components["schemas"]["ServiceLogs"]> {
```
Dartoxian: It was dropped when my commit was rolled back, but having a type like `type KurtosisRestTypes = components["schemas"]` would be slightly better to use throughout the code. Importing a type called `components` everywhere we need them is misleading as:
* The reader probably expects it to contain React components
* Lowercase name for a type is not consistent with style
* Unclear what `components` the thing is referring to

enclave-manager/web/src/client/enclaveManager/websocketClient/WebSocketClient.d.ts
```diff
@@ -0,0 +1,43 @@
...
+        url: P,
+        ...init: HasRequiredKeys<
+            RequestOptions<FilterKeys<Paths[P], "get">>
+        > extends never
+            ? [(RequestOptions<FilterKeys<Paths[P], "get">> | undefined)?]
+            : [RequestOptions<FilterKeys<Paths[P], "get">>]
+    ): AsyncGenerator<FetchResponse<"get" extends infer T ? T extends "get" ? T extends keyof Paths[P] ? Paths[P][T] : unknown : never : never>>;
+};
+
+export { };
```
Dartoxian: We should avoid using type definition files alongside JS as much as possible imo:
* There is a higher developer maintenance burden because two files must now be maintained for a single piece of functionality.
* Inconsistencies between (complex) types and implementation are generally only going to be discovered at runtime.

This client could be written as a single `.ts` file right?

enclave-manager/web/src/client/enclaveManager/KurtosisClientContext.tsx
```diff
@@ -7,6 +7,8 @@ import { KURTOSIS_CLOUD_EM_PAGE, KURTOSIS_CLOUD_EM_URL } from "../constants";
 import { AuthenticatedKurtosisClient } from "./AuthenticatedKurtosisClient";
 import { KurtosisClient } from "./KurtosisClient";
 import { LocalKurtosisClient } from "./LocalKurtosisClient";
+import React from "react";
+import { EnclaveInfo } from "enclave-manager-sdk/build/engine_service_pb";
```
Dartoxian: Don't think anything has changed in this file to warrant this. In this part of the codebase we use a code formatter called `prettier` - you can either have it automatically run on save by your ide, or run `yarn prettier:fix` to run across all of the code.

enclave-manager/web/src/client/enclaveManager/websocketClient/WebSocketClient.d.ts
adschwartz: Agreed

enclave-manager/web/src/client/enclaveManager/KurtosisClient.ts
lostbean: Yes, using a type alias for this will be much clear.

enclave-manager/web/src/client/enclaveManager/websocketClient/WebSocketClient.d.ts
lostbean: Agreed! Not sure why `openapi-fetch` went with JS + type interface but I don't a strong reason make into a pure TS  implementation and have all the benefits of the type system.

enclave-manager/web/src/client/enclaveManager/KurtosisClientContext.tsx
lostbean: nice tip! just added it to my IDE

enclave-manager/web/src/client/enclaveManager/websocketClient/WebSocketClient.d.ts
lostbean: done :D

barnabasbusa:

PR #1962, opened 2023-12-15
fix: Set the user service K8S ingress labels so it can be found

laurentluce:
## Description:
The user service K8S ingress labels need to be set properly so the ingress can be found during retrieval and deletion operations.

NO

#1941

### DIFF
--- a/container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/label_value_consts/label_value_consts.go
+++ b/container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/label_value_consts/label_value_consts.go
@@ -15,9 +15,9 @@ const (
 	engineKurtosisResourceTypeLabelValueStr = "kurtosis-engine"
 	// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! DO NOT CHANGE THESE VALUES !!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-	enclaveKurtosisResourceTypeLabelValueStr            = "enclave"
-	apiContainerKurtosisResourceTypeLabelValueStr       = "api-container"
-	userServiceKurtosisResourceTypeLabelValueStr        = "user-service"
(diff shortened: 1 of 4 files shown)


container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/label_value_consts/label_value_consts.go
```diff
@@ -15,10 +15,9 @@ const (
 	engineKurtosisResourceTypeLabelValueStr = "kurtosis-engine"
 	// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! DO NOT CHANGE THESE VALUES !!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-	enclaveKurtosisResourceTypeLabelValueStr            = "enclave"
-	apiContainerKurtosisResourceTypeLabelValueStr       = "api-container"
-	userServiceKurtosisResourceTypeLabelValueStr        = "user-service"
-	userServiceIngressKurtosisResourceTypeLabelValueStr = "user-service-ingress"
```
laurentluce, line 21: The ingress like other user service resources uses the "user-service" resource type label.

h4ck3rk3y: Fair enough, this will then follow the rules of retrieval/deletion of a user services as well? And are we okay with it!

h4ck3rk3y (approved):

h4ck3rk3y (approved):

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/label_value_consts/label_value_consts.go
laurentluce: Yes because we want to collect all the user service resources.  We do the same for the K8S service and pod.

PR #1963, opened 2023-12-15
feat: add production mode to k8s

h4ck3rk3y:
## Description:
1. Services restart on failure like Docker
2. /kurtosis-data/ is in a PV on k8s so APIC is restartable

YES


container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/kubernetes_kurtosis_backend_api_container_functions.go
```diff
@@ -411,13 +413,44 @@ func (backend *KubernetesKurtosisBackend) CreateAPIContainer(
...
+
 	apiContainerContainers, apiContainerVolumes, err := getApiContainerContainersAndVolumes(image, containerPorts, envVarsWithOwnIp, enclaveDataVolumeDirpath)
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "An error occurred getting API containers and volumes")
 	}
 
 	apiContainerInitContainers := []apiv1.Container{}
 
+	// Data is always persistent we can always restart like Docker
+	apiContainerRestartPolicy := apiv1.RestartPolicyOnFailure
```
h4ck3rk3y, line 453: @leoporoli Questions here -

This is very similar to what Docker does; and docker always has persistent volume for `/kurtosis-data` but on Kubernetes that adds time on DO/AWS; on k3s it seems like no time is added

leoporoli: we talked about this on Slack and I agree with adding this for all enclaves so far.

leoporoli (approved):

PR #1965, opened 2023-12-15
fix: ignore the current status of the service during a start/stop

h4ck3rk3y:


core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -463,9 +463,6 @@ func (network *DefaultServiceNetwork) StartServices(
 		if err != nil {
 			return nil, nil, stacktrace.Propagate(err, "An error occurred while getting service registration for identifier '%v'", serviceIdentifier)
 		}
-		if serviceRegistration.GetStatus() == service.ServiceStatus_Started {
```
h4ck3rk3y, line 466: @laurentluce I was demoing Kurtosis to Barnabas and we tried a workflow where he deleted a pod and the pod was attached to a persistent volume ( my current work)

we wanted to bring the pod back alive; so i tried start but the complain was that the service is started as the service is started according to the Kurtosis DB; so I had to issue a stop and then a start.

Talking to Kevin he said maybe the service start command should be idempotent. Tagging you on this PR as you have most knowledge about this; think this is okay? I tried the same flow that I ran with Barnabas and that works; still waiting for our gauntlet of tests to run!

laurentluce: We had a discussion about that with @gbouv when designing the start and stop service feature.  We wanted to be consistent with the other operations like add and remove service.  The Docker API client also errors out when you try to start an already started container.  Discussion: https://www.notion.so/Starting-and-Stopping-a-Service-b601dc1f60c847fca9cfaf832d424046?d=48ed233b181a44319d4234dd7a616f27&pvs=4#f0b843e7af4146aa91f207515587a0bf

h4ck3rk3y: Hmm, this is a shift towards "declarative Kurtosis" as Kevin puts it; the problem with our approach is that the Status of the Service is different from the Status of the Container(dead/non existent). Our service status is the desired service status but I do see your point.

Do you think this would pose a problem?

laurentluce: Yes the service status is not retrieved from the backend which is something @tedim52 looked into.  I don't see an issue with making it idempotent.

laurentluce (approved):

PR #1970, opened 2023-12-17
feat: Engine K8S ingress for REST API reverse proxy routing

laurentluce:
## Description:
Create engine K8S ingress for the REST API port so we can interact with the API via Traefik.  The `Host` header is set to `engine`.

Tested with k3d locally:
```
curl -i http://localhost:9730/engine/info -H "Host: engine"
HTTP/1.1 200 OK
...
{"engine_version":"ae7579"}
```

NO

#1941


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -27,7 +27,6 @@ const (
 	//TODO: pass this parameter
 	enclaveManagerUIPort                        = 9711
 	enclaveManagerAPIPort                       = 8081
-	restAPIPort                                 = 9779 //TODO: pass this parameter
```
laurentluce, line 30: Move some of those constants to the backend interface engine object so they can be used by both backends.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/consts/consts.go
```diff
@@ -14,11 +15,16 @@ const (
 	//  be stored in the port spec label
 	KurtosisInternalContainerGrpcPortSpecId = "grpc"
 
-	// The ID of the GRPC proxy port for Kurtosis-internal containers. This is necessary because
```
laurentluce, line 17: GRPC proxy port is not used anymore.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -165,13 +179,34 @@ func CreateEngine(
...
+		ctx,
+		namespaceName,
+		engineAttributesProvider,
+		privateRESTAPIPortSpec,
+		kubernetesManager,
+	)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred creating the engine ingress")
+	}
+	var shouldRemoveIngress = true
```
lostbean, line 192: why this is a `var` and why this always `true`? It seems I'm missing something in the behavior of of this code

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -165,13 +179,34 @@ func CreateEngine(
...
+	defer func() {
```
lostbean, line 193: My reading on why we have this async (goroutine) part is this is a non-critical code path that if fails there is no big deal. So we can spawn off and try it.  But the `ERROR` log and `ACTION REQUIRED` in case of failure gives a conflicting reading.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -490,11 +530,83 @@ func createEngineService(
...
+	engineIngressName string,
+	privateRESTAPIPortSpec *port_spec.PortSpec,
+) ([]netv1.IngressRule, error) {
+	var ingressRules []netv1.IngressRule
+	ingressRule := netv1.IngressRule{
+		Host: restAPIPortHost,
+		IngressRuleValue: netv1.IngressRuleValue{
+			HTTP: &netv1.HTTPIngressRuleValue{
+				Paths: []netv1.HTTPIngressPath{
+					{
```
lostbean, line 591: nit: maybe we could break down this data structure tree into smaller bits to facilitate the reading (e.g. `ingressPath := {....}` and then `Paths: []netv1.HTTPIngressPath{ingressPath}`).

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/engine_functions/shared_helpers.go
```diff
@@ -297,9 +302,38 @@ func getMatchingEngineKubernetesResources(
...
+		)
+		if err != nil {
+			return nil, stacktrace.Propagate(err, "An error occurred getting ingresses matching engine GUID '%v' in namespace '%v'", engineGuid, namespaceName)
+		}
+		var ingress *netv1.Ingress
+		if ingressesForId, found := ingresses[engineGuidStr]; found {
+			if len(ingressesForId) > 1 {
+				return nil, stacktrace.NewError(
+					"Expected at most one engine ingress in namespace '%v' for engine with GUID '%v' "+
+						"but found '%v'",
```
lostbean: nit: awkward line break. Is it the formatter?

lostbean (approved):
left few comments but LGTM

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/engine_functions/create_engine.go
laurentluce: It is set to true here so the code tries to remove the ingress if the `CreateEngine` function returns early with an error.  This pattern of cleanup in case of error is used a lot at Kurtosis.  This variable is set to false at the end of the `CreateEngine` function if everything went well.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/engine_functions/create_engine.go
laurentluce: Those are not asyncs but defers: cleanup in case of an early exit due to an error.

laurentluce:
> left few comments but LGTM

Thanks @lostbean for the review!

PR #1976, opened 2023-12-18
feat: update api path and keep alive

lostbean:
## Description:
Update REST API path with `/api` prefix and add websocket ping/pong handler to keep connection alive.

NO


engine/server/engine/main.go
```diff
@@ -411,14 +413,15 @@ func restApiServer(
 
 	// This is how you set up a basic Echo router
 	echoRouter := echo.New()
-	echoRouter.Use(echomiddleware.Logger())
+	echoApiRouter := echoRouter.Group("/api")
```
leoporoli: I think you can use this constant here `pathToApiGroup`

engine/server/engine/main.go
```diff
@@ -411,14 +413,15 @@ func restApiServer(
...
+	echoApiRouter.Use(echomiddleware.Logger())
```
leoporoli, line 417: nice!

engine/server/engine/server/websocket_api_handler.go
```diff
@@ -245,32 +274,43 @@ func streamStarlarkLogsWithHTTP[T any](ctx echo.Context, streamerPool streaming.
...
-		err := streamer.Consume(func(logline *api_type.ServiceLogs) error {
-			err := websocket.JSON.Send(ws, logline)
-			if err != nil {
-				return err
-			}
-			return nil
+func streamServiceLogsWithWebsocket(ctx echo.Context, cors cors.Cors, streamer streaming.ServiceLogStreamer) {
+	wsPump, err := NewWebsocketPump[api_type.ServiceLogs](ctx, cors)
+	if err != nil {
+		logrus.WithError(err).Error("Failed to start websocket connection")
```
leoporoli, line 264: I think you could create a helper function here to dry a bit, it's the same used in `streamStarlarkLogsWithWebsocket`

engine/server/engine/server/websocket_api_handler.go
```diff
@@ -245,32 +274,43 @@ func streamStarlarkLogsWithHTTP[T any](ctx echo.Context, streamerPool streaming.
...
+	defer wsPump.Close()
+	go wsPump.startPumping()
+
+	err = streamer.Consume(func(logline *api_type.ServiceLogs) error {
+		wsPump.pumpMessage(logline)
+		return nil
+	})
+
+	if err != nil {
+		logrus.WithError(err).WithFields(logrus.Fields{
```
leoporoli, line 281: same for this error

engine/server/engine/server/websocket_api_handler.go
```diff
@@ -300,3 +340,95 @@ func streamServiceLogsWithHTTP(ctx echo.Context, streamer streaming.ServiceLogSt
 		}
 	}
 }
+
+type WebsocketPump[T interface{}] struct {
```
leoporoli: small suggestion, I think this object and methods could be on its file

engine/server/engine/server/websocket_api_handler.go
```diff
@@ -300,3 +340,95 @@ func streamServiceLogsWithHTTP(ctx echo.Context, streamer streaming.ServiceLogSt
...
+
+func NewWebsocketPump[T interface{}](ctx echo.Context, cors cors.Cors) (*WebsocketPump[T], error) {
+	// nolint: exhaustruct
+	upgrader := websocket.Upgrader{
+		ReadBufferSize:  wsReadBufferSize,
+		WriteBufferSize: wsWriteBufferSize,
+		CheckOrigin:     cors.OriginAllowed,
+	}
+
+	conn, err := upgrader.Upgrade(ctx.Response(), ctx.Request(), nil)
```
leoporoli: nice!

engine/server/engine/main.go
lostbean: fixed. nice catch!

engine/server/engine/server/websocket_api_handler.go
```diff
@@ -300,3 +340,95 @@ func streamServiceLogsWithHTTP(ctx echo.Context, streamer streaming.ServiceLogSt
...
+func (pump *WebsocketPump[T]) pumpResponseInfo(msg *api_type.ResponseInfo) {
+	pump.infoChan <- msg
+}
+
+func (pump *WebsocketPump[T]) pumpMessage(msg *T) {
+	pump.inputChan <- msg
+}
+
+func (pump *WebsocketPump[T]) Close() {
+	pump.cancelFunc()
```
leoporoli: I think you can close the web socket here too and the channels, or is it necessary to let them open?

leoporoli: and the ticker?

engine/server/engine/server/websocket_api_handler.go
lostbean: Here it's different, we are returning the error over the websocket channel. I can't use the same abstraction.

engine/server/engine/streaming/websocket_pump.go
```diff
@@ -0,0 +1,119 @@
...
+
+	logrus.WithFields(logrus.Fields{
+		"pongWait":       pongWait,
+		"pingPeriod":     pingPeriod,
+		"maxMessageSize": maxMessageSize,
+	}).Debug("Started keep alive process for websocket connection.")
+
+	pump.websocket.SetReadLimit(maxMessageSize)
+	if err := pump.websocket.SetReadDeadline(time.Now().Add(pongWait)); err != nil {
+		logrus.WithError(err).Error("Failed to set Pong wait time")
```
leoporoli, line 78: will the client receive and error if this happens?

engine/server/engine/server/websocket_api_handler.go
lostbean: If I do it here, we'll be closing it twice and it'll throw an error. The goroutine `StartPumping()` will always close it on exit (`defer`).

engine/server/engine/streaming/websocket_pump.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (pump WebsocketPump[T]) StartPumping() {
+	ticker := time.NewTicker(pingPeriod)
+	defer func() {
+		ticker.Stop()
+		pump.websocket.Close()
+	}()
+
+	logrus.WithFields(logrus.Fields{
+		"pongWait":       pongWait,
```
leoporoli, line 71: nit: you could use constants for these keys

engine/server/engine/streaming/websocket_pump.go
lostbean: no, this is logged only the server side. if this fails, it's very likely the connection is broken.

engine/server/engine/streaming/websocket_pump.go
lostbean: what's the benefity in this case? it does seems an overkill, no?

leoporoli (approved):

engine/server/engine/server/websocket_api_handler.go
leoporoli: @lostbean makes sense, so the close channels could be included in the defer on exit? it is suggested to always close the channels when the writer ends the task

engine/server/engine/streaming/websocket_pump.go
leoporoli: probably is an over-engineering suggestion, I use to do it for me or someone else copy the code later both pieces have the same key and then if it is edited, will be edited in only one place, but.
Not necessary to do the change.

engine/server/engine/server/websocket_api_handler.go
lostbean: yes, I forgot to close the go channels!

lostbean: done!

PR #1996, opened 2023-12-20
fix: change restart policy to always

h4ck3rk3y:
DIVA node kills itself after an API call and needs to restart

### DIFF
--- a/container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
+++ b/container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
@@ -229,7 +229,7 @@ func (backend *DockerKurtosisBackend) StartRegisteredUserServices(ctx context.Co
 
 	var restartPolicy docker_manager.RestartPolicy = docker_manager.NoRestart
 	if backend.productionMode {
-		restartPolicy = docker_manager.RestartOnFailure
+		restartPolicy = docker_manager.RestartAlways
 	}
 
 	successfullyStartedService, failedService, err := user_service_functions.StartRegisteredUserServices(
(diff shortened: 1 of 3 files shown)


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -229,7 +229,7 @@ func (backend *DockerKurtosisBackend) StartRegisteredUserServices(ctx context.Co
 
 	var restartPolicy docker_manager.RestartPolicy = docker_manager.NoRestart
 	if backend.productionMode {
-		restartPolicy = docker_manager.RestartOnFailure
+		restartPolicy = docker_manager.RestartAlways
```
leoporoli, line 232: @h4ck3rk3y Do you know why RestartOnFailure did not restart the container in the DIVA use case?

h4ck3rk3y: Yes, it didn't restart for DIVA. I built this image and restarted and DIVA was restartign!

leoporoli: shouldn't they refactor their code to exit >0 if their service dies?

h4ck3rk3y: So the service dying is the expected behavior in this case! There's an API call that you make and that API call is meant to restart the service but I empathize

leoporoli: but probably yes, it's better to always restart a service in production mode

leoporoli (approved):

PR #2001, opened 2023-12-22
feat: docker compose integration [pt. 1]

tedim52:
## Description:
This PR is the pt. 1 of enabling the use case:
```
kurtosis run docker-compose.yaml
```
This PR contains only the logic for transpiling docker compose into starlark. The next PR will integrate it with APIC.

NO (next one will be).


core/server/api_container/server/docker_compose_transpiler/docker_compose_transpiler.go
```diff
@@ -0,0 +1,582 @@
...
+	envVarsFilepath := path.Join(packageAbsDirpath, envVarsFilename)
+	var envVars map[string]string
+	envVarsInFile, err := godotenv.Read(envVarsFilepath)
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return "", stacktrace.Propagate(err, "An %v file was found in the package, but an error occurred reading it.", envVarsFilename)
+		}
+		envVarsInFile = map[string]string{}
+	}
+	envVars = envVarsInFile
```
lostbean: do we really need `envVars` here? Why not just use `envVarsInFile`

core/server/api_container/server/docker_compose_transpiler/docker_compose_transpiler.go
```diff
@@ -0,0 +1,582 @@
...
+		Environment: envVars,
+	}
+	setOptionsFunc := func(options *loader.Options) {
+		options.SetProjectName(composeProjectName, shouldOverrideComposeYamlKeyProjectName)
+
+		// Don't resolve paths as they should be resolved by package content provider
+		options.ResolvePaths = false
+
+		// Don't convert Windows paths to Linux paths as APIC runs on Linux
+		options.ConvertWindowsPaths = true
```
lostbean: Wondering if this kind of configuration should go in the `const` section at the top.

core/server/api_container/server/docker_compose_transpiler/docker_compose_transpiler.go
```diff
@@ -0,0 +1,582 @@
...
+		); err != nil {
+			return nil, stacktrace.Propagate(err, "An error occurred setting key '%s' in environment variables Starlark dict.", key)
+		}
+	}
+	return enVarsSLDict, nil
+}
+
+// The 'volumes:' compose key supports named volumes and bind mounts https://docs.docker.com/storage/volumes/
+// bind mount semantics for starlark:
+// <rel path on host>:<path on container> := upload a files artifacts of <rel path on host>, mount the files artifacts on the container at <path on container>
```
lostbean, line 434: These kind of encoding seems a bit trick and can potentially lead to some confusion with user. Wondering if we should simplify the conversion to a safer transcription and let the user adjust the richer options in Starlark later.

core/server/api_container/server/docker_compose_transpiler/docker_compose_transpiler.go
```diff
@@ -0,0 +1,582 @@
...
+	tuple := []starlark.Value{
+		starlark.String(argName),
+		argValue,
+	}
+	return append(kwargs, tuple)
+}
+
+// Returns list of service names in an order that respects dependencies by performing a topological sort
+// Returns error if cyclical dependency is detected
+// o(n^2) but simpler variation of Kahns algorithm https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm
```
lostbean, line 538: Nice :)

lostbean (approved):
Left some small comments but looks good

core/server/api_container/server/docker_compose_transpiler/docker_compose_transpiler.go
tedim52: Definitely. Particularly with volume's, it's not a 1:1 match in terms of semantics, but rather it requires translating Compose's volume concept into our concepts which definitely needs to be transparent to users. 

Going to put some thought into this, but do you have any ideas for what a simpler transcription might be?

core/server/api_container/server/docker_compose_transpiler/docker_compose_transpiler.go
tedim52: facts

core/server/api_container/server/docker_compose_transpiler/docker_compose_transpiler.go
tedim52: facts

core/server/api_container/server/docker_compose_transpiler/docker_compose_transpiler.go
tedim52: After putting some thought into it, going to merge this for now and return to this when writing docs on the translation between files artifact and volume.

PR #2004, opened 2023-12-28
fix: Make the reverse proxy connect and disconnect to and from the enclave network idempotent

laurentluce:
## Description:
This PR makes the reverse proxy network calls idempotent so the engine can be started when networking is already set up.  This can happen when you restart the docker daemon.

In addition, this PR appends the engine UUID to the reverse proxy container name to make it unique across runs.  This is an approach similar to the engine naming scheme which is cleaner and safer.

NO

#2005


container-engine-lib/lib/backend_impls/docker/object_attributes_provider/object_attributes_provider.go
```diff
@@ -137,14 +138,33 @@ func (provider *dockerObjectAttributesProviderImpl) ForLogsStorageVolume() (Dock
...
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred creating a Docker object name object from string '%v'", nameStr)
+	}
+
+	idLabelValue, err := docker_label_value.CreateNewDockerLabelValue(string(engineGuid))
 	if err != nil {
-		return nil, stacktrace.Propagate(err, "An error occurred creating a Docker object name object from string '%v'", reverseProxyName)
+		return nil, stacktrace.Propagate(err, "An error occurred creating the reverse proxy GUID Docker label from string '%v'", engineGuid)
+	}
+	guidLabelValue, err := docker_label_value.CreateNewDockerLabelValue(string(engineGuid))
```
tedim52, line 159: I'm not too familiar/don't remember what the implementation of `CreateNewDockerLabelValue` is but what's the difference between the `guid` and the `id` label value? Or are they meant to be the same.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/reverse_proxy_functions/network_reverse_proxy.go
```diff
@@ -20,13 +20,18 @@ var (
...
 func ConnectReverseProxyToNetwork(ctx context.Context, dockerManager *docker_manager.DockerManager, networkId string) error {
-	_, maybeReverseProxyContainerId, err := getReverseProxyObjectAndContainerId(ctx, dockerManager)
+	maybeReverseProxyObject, maybeReverseProxyContainerId, err := getReverseProxyObjectAndContainerId(ctx, dockerManager)
 	if err != nil {
 		logrus.Warnf("Attempted to connect reverse proxy to a network but no reverse proxy container was found.")
 		return nil
 	}
 
-	if maybeReverseProxyContainerId == "" {
+	if maybeReverseProxyObject == nil {
```
tedim52, line 27: Can we add a comment explaining what this check means semantically? I think it means "If theres no reverse proxy, we can't connect it to the network so don't do anything" but initially I thought it should've returned an error because the reverse proxy didn't exist.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/reverse_proxy_functions/network_reverse_proxy.go
```diff
@@ -48,6 +53,11 @@ func DisconnectReverseProxyFromNetwork(ctx context.Context, dockerManager *docke
 		return nil
 	}
 
+	_, found := maybeReverseProxyObject.GetEnclaveNetworksIpAddress()[networkId]
```
tedim52, line 54: Does there need to be a  `maybeReverseProxyObject != nil` check?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/reverse_proxy_functions/network_reverse_proxy.go
tedim52: Also, out of curiosity, what happens if no reverse proxy is connected to the enclave network?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/reverse_proxy_functions/network_reverse_proxy.go
```diff
@@ -20,13 +20,18 @@ var (
...
 		logrus.Warnf("Attempted to connect reverse proxy to a network but no reverse proxy container was found.")
 		return nil
 	}
 
-	if maybeReverseProxyContainerId == "" {
+	if maybeReverseProxyObject == nil {
+		return nil
+	}
+
+	_, found := maybeReverseProxyObject.GetEnclaveNetworksIpAddress()[networkId]
```
tedim52, line 31: Maybeee a comment here too?

tedim52 (approved):
LGTM! I think one or two comments explaining the idempotent behavior would be helpful

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/object_attributes_provider.go
laurentluce: For some objects like services, the id is the name and the guid is the uuid.  For others like the engine, both id and guid labels are set to the uuid.  I did the same for the reverse proxy for consistency.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/reverse_proxy_functions/network_reverse_proxy.go
laurentluce: I agree the current logic is flawed, we are tolerating invalid configurations.  I updated both the connect and disconnect functions to error if the reverse proxy does not exist since it is mandatory.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/reverse_proxy_functions/network_reverse_proxy.go
laurentluce: Cleaned that up too.

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/object_attributes_provider.go
tedim52: ah yes I remember this now makes sense

PR #2008, opened 2024-01-02
fix: log file path formatting for week

tedim52:
## Description:
log file paths were being written with week `01` and being read as week `1`. This PR fix this by making it so all weeks have at least two digits and adds a leading zero if single digit.

NO

### DIFF
--- a/engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
+++ b/engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
@@ -206,25 +206,27 @@ func (manager *LogFileManager) createSymlinkLogFile(targetLogFilePath, symlinkLo
 	return nil
 }
 
-// TODO: Implement a FilePath Builder to remove exploding constructors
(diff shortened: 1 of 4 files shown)


engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
```diff
@@ -206,25 +206,27 @@ func (manager *LogFileManager) createSymlinkLogFile(targetLogFilePath, symlinkLo
 	return nil
 }
 
-// TODO: Implement a FilePath Builder to remove exploding constructors
+// TODO: Implement a FilePath Builder to centralize log file path creation across the entire module
```
tedim52, line 209: log filepath code is in need of a refactor+tests for log file manager to centralize all log file path creation. wanted to get this fix out but going to do this refactor in my next available cycle

tedim52: ideally i would've only had to make this change once but i had to in several places to fix formatting which is bad

lostbean (approved):

PR #2019, opened 2024-01-04
feat: Engine Traefik Docker labels for REST API reverse proxy routing

laurentluce:
## Description:
Create engine Traefik Docker labels for the REST API port so we can interact with the API via Traefik. The Host header is set to "engine".

Tested with Docker locally:

```
$ curl -i http://localhost:9730/api/engine/info -H "Host: engine"
HTTP/1.1 200 OK
...
{"engine_version":"70d65c"}
```

NO

#1970


container-engine-lib/lib/backend_impls/docker/object_attributes_provider/object_attributes_provider.go
```diff
@@ -173,3 +188,67 @@ func (provider *dockerObjectAttributesProviderImpl) ForReverseProxy(engineGuid e
...
 }
+
+// Return Traefik labels
+// Including the labels required to route traffic to the engine rest api port if the Host header is set to "engine".
+//
+//	"traefik.enable": "true",
+//	"traefik.http.routers.engine-rest-api.rule": "Host(`engine`)",
+//	"traefik.http.routers.engine-rest-api.service": "engine-rest-api",
+//	"traefik.http.services.engine-rest-api.loadbalancer.server.port": "<engine rest api port number>"
+func (provider *dockerObjectAttributesProviderImpl) getTraefikLabelsForEngine(restAPIPortSpec *port_spec.PortSpec) (map[*docker_label_key.DockerLabelKey]*docker_label_value.DockerLabelValue, error) {
```
lostbean, line 199: just a small improvement idea (probably coming from my functional oriented side :p) to deduce the code a bit. Basically create a map with the raw string pairs and then mapping over to the docker lables:
```
func (provider *dockerObjectAttributesProviderImpl) getTraefikLabelsForEngine(restAPIPortSpec *port_spec.PortSpec) (map[*docker_label_key.DockerLabelKey]*docker_label_value.DockerLabelValue, error) {
	var labelStringPairs map[string]string
	labelStringPairs = map[string]string{
		fmt.Sprintf("some %s", "key"): fmt.Sprintf("some %s", "value"),
		fmt.Sprintf("some %s", "key"): fmt.Sprintf("some %s", "value"),
		fmt.Sprintf("some %s", "key"): fmt.Sprintf("some %s", "value"),
	}

	labels := map[*docker_label_key.DockerLabelKey]*docker_label_value.DockerLabelValue{}
	for key, value := range labelStringPairs {
		labelKey, err := docker_label_key.CreateNewDockerTraefikLabelKey(key)
		if err != nil {
			return nil, stacktrace.Propagate(err, "An error occurred getting the traefik  label key with suffix '%v'", key)
		}
		labelValue, err := docker_label_value.CreateNewDockerLabelValue(value)
		if err != nil {
			return nil, stacktrace.Propagate(err, "An error occurred creating the traefik  label value with value '%v'", value)
		}
		labels[labelKey] = labelValue
	}

	return labels, nil
}
```

lostbean (approved):
LGTM, just left a comment for a possible code simplification.

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/object_attributes_provider.go
laurentluce: I changed that and also for the user service labels.

PR #2025, opened 2024-01-08
feat!: allow to mount multiple artifacts to the same folder in a service. Users will need to replace the `Directory.artifac_name` field key with `Directory.artifac_names`

### LINKED ISSUE #1519: Allow to mount multiple artifacts to the same folder in a service
### Background & motivation

When creating a new service it is not possible to add multiple artifacts to the same path in the `files` dictionary. For example, in this code there is no direct way to put both files inside the `/files` directory in the service:

```python
def run(plan, args={}):
    artifact_name_1 = plan.upload_files("./folder-1/file-1.txt")
    artifact_name_2 = plan.upload_files("./folder-2/file-2.txt")

    service = plan.add_service(
        "service-example",
        config=ServiceConfig(
            image="alpine:latest",
            command=["ls", "/files"],
            files={
                "/files": artifact_name_1,
                # "/files": artifact_name_2,
            },
        ),
    )
```

**Workaround:**

Currently, this is possible with the use of [plan.run_sh](https://docs.kurtosis.com/starlark-reference/plan#run_sh), mounting artifacts to different directories, combining them with the shell command(`cp`, `mv`, or similar), and later extracting a new artifact using the `store` parameter. This new artifact can be mounted in the others services.

### Desired behaviour

It should be possible to allow the use of the same path for multiple artifacts that aren't loaded together. Also, allowing to combine artifacts in an easier way could also solve this issue.

leoporoli:
## Description:
allow to mount multiple artifacts to the same folder in a service

YES

Fix https://github.com/kurtosis-tech/kurtosis/issues/1519


core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/test_engine/service_config_multiple_files_in_same_folder_test.go
```diff
@@ -0,0 +1,64 @@
+package test_engine
+
+import (
```
h4ck3rk3y, line 3: thanks for being so thorough!

docs/docs/api-reference/starlark-reference/directory.md
```diff
@@ -5,24 +5,32 @@ sidebar_label: Directory
...
 the [ServiceConfig][service-config] object).
-Directory object can be either a files artifact or a persistent directory, depending on the arguments passed to the 
+Directory object can be either one or many files artifacts or a persistent directory, depending on the arguments passed to the 
 constructor.
 
 ```python
+# composed with only one file artifact:
 file_artifact_directory = Directory(
-    artifact_name=files_artifact_1,
+    artifact_names=files_artifact_1,
```
h4ck3rk3y: will have to edit this when the syntax gets changed

internal_testsuites/golang/testsuite/startosis_directory_test/startosis_mutiple_artifacts_directory_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	multipleArtifactsDirectoryPackage = "../../../starlark/service-directory/multiple-artifacts-directory"
+)
+
+func (suite *StartosisDirectoryTestSuite) TestAddServiceWithMultipleArtifactsDirectory() {
+	ctx := context.Background()
+	rulResult, err := suite.RunPackage(ctx, multipleArtifactsDirectoryPackage)
```
h4ck3rk3y: should this be `runResult`?

h4ck3rk3y (approved):

docs/docs/api-reference/starlark-reference/directory.md
leoporoli: updated, thanks for the reminder

PR #2030, opened 2024-01-08
fix: treat size value as megabytes

h4ck3rk3y:
## Description:
We expect(in DOCS) the user to enter value in megabytes but were never multiplying the value with the megaByteToByte multiplier


barnabasbusa:
Could we actually use Gigabytes over megabytes?

barnabasbusa (approved):

h4ck3rk3y:
@barnabasbusa discussed with you, this fix is fine for now but i will support 5GB, 2.5TB or other values, please create a ticket!

leoporoli (approved):

PR #2042, opened 2024-01-11
chore: ete emui cypress

Dartoxian:
## Description:
This PR adds ETE cypress testing for the new emui.

NO


enclave-manager/web/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
```
adschwartz: @Dartoxian do we need to remove this file?

adschwartz (approved):
Great work. May want to remove the example.json but otherwise gtg

enclave-manager/web/cypress/fixtures/example.json
Dartoxian: ah thanks for the flag, yep I'll tidy this up.

PR #2043, opened 2024-01-11
feat: docker compose integration pt. 2

tedim52:
## Description:
This feature allows users to run basic docker compose's with Kurtosis. Now, directories containing a `kurtosis.yml` OR a `docker-compose.yaml` are treated as starlark packages. If a compose package is detected, `APIC` converts the compose to Starlark, and executes the Starlark within enclaves.

eg. 
`kurtosis run .`
`kurtosis run github.com/awesome-compose/django`

YES

## References:
Part 1 (Docker Compose Transpiler) https://github.com/kurtosis-tech/kurtosis/pull/2001


api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -47,8 +47,22 @@ const (
...
 
 	dotRelativePathIndicatorString = "."
+
+	// required to get around "only Github URLs" validation
+	composePackageIdPlaceholder = "github.com/NOTIONAL_USER/USER_UPLOADED_COMPOSE_PACKAGE"
 )
 
+// TODO Remove this once package ID is detected ONLY the APIC side (i.e. the CLI doesn't need to tell the APIC what package ID it's using)
+// Doing so requires that we upload completely anonymous packages to the APIC, and it figures things out from there
+var supportedDockerComposeYmlFilenames = []string{
```
tedim52, line 57: this list exists twice in  `api` and once in `core` but idk a great place to put a `const` for this so it's defined only once, maybe create a `kurtosis_constants` library?

api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -143,12 +157,22 @@ func (enclaveCtx *EnclaveContext) RunStarlarkPackage(
 
 	starlarkResponseLineChan := make(chan *kurtosis_core_rpc_api_bindings.StarlarkRunResponseLine)
 
-	kurtosisYml, err := getKurtosisYaml(packageRootPath)
+	packageName, packageReplaceOptions, err := getPackageNameAndReplaceOptions(packageRootPath)
 	if err != nil {
-		return nil, nil, stacktrace.Propagate(err, "An error occurred getting Kurtosis yaml file from path '%s'", packageRootPath)
```
lostbean, line 148: nit: why removing the propagation here?

core/server/api_container/server/docker_compose_transpiler/docker_compose_transpiler.go
```diff
@@ -448,6 +449,14 @@ func getStarlarkFilesArtifacts(composeVolumes []types.ServiceVolumeConfig, servi
 		var shouldPersist bool
 		switch volumeType {
 		case types.VolumeTypeBind:
+			// Handle case where home path is reference
```
lostbean, line 453: The `~` home folder is a strong indication the volume bind might be outside the project folder. But a lot of other paths can also be outside. Wondering is there a better way of detecting it e.g. resolving volume paths against the project path and pattern matching it?  Maybe we could just block absolute paths (`~` includude) or relative paths going upstream `../something`

lostbean (approved):

api/golang/core/lib/enclaves/enclave_context.go
tedim52: good point will add the propagation!

core/server/api_container/server/docker_compose_transpiler/docker_compose_transpiler.go
tedim52: Good point, right now, we are treating absolute paths as persistent directories, but that doesn't account for home path and didn't think about the case where relative paths that go upstream as they seem to be less common since `docker-compose.yaml` are usually at the root of the project

tedim52: For now, I'm going to keep the absolute path behavior (these become persistent behavior), but I'm going to account for the case where relative paths goes upstream eg. `..` and block that here as well.

tedim52: I agree this can be handled better, but want to leave that for the future if this feature requires iteration from usage : )

PR #2047, opened 2024-01-15
fix: Make sure that the tags are fetched before testing with git describe

Dartoxian:
## Description:
This fixes the issue seen in https://app.circleci.com/pipelines/github/kurtosis-tech/kurtosis/10677/workflows/66565433-39e0-4397-b0d8-4c3c81ad366b/jobs/152621. 

When ci checks out code in a step it doesn't fetch tags, so `git describe --exact-match` will always fail - this change fixes this by fetching tags. 

NO

- https://app.circleci.com/pipelines/github/kurtosis-tech/kurtosis/10677/workflows/66565433-39e0-4397-b0d8-4c3c81ad366b/jobs/152621
-https://github.com/kurtosis-tech/kurtosis/pull/2041


.circleci/config.yml
```diff
@@ -435,6 +435,7 @@ jobs:
 
       - run: |
           # If the current commit is not a tag and if there is no change (compared to main) of anything but docs, then skip this
+          git fetch origin --tags
```
leoporoli, line 438: you can add extra guard rails with

```
if ! git fetch origin --tags; then
  touch /tmp/testsuite-failed
fi
```

leoporoli (approved):

.circleci/config.yml
Dartoxian: But if the command has a non zero exit code the test suite will fail anyway.

PR #2052, opened 2024-01-16
fix: emui enable exact package name searching

### LINKED ISSUE #2051: Going back and forth between Form and YAML has a quoting issue so that both get progressively longer
### What's your CLI version?

0.86.6

Go to https://cloud.kurtosis.com/enclave-manager?package-id=github.com%2Fkurtosis-tech%2Fethereum-package&package-args=eyJlbmNsYXZlTmFtZSI6IiIsInJlc3RhcnRTZXJ2aWNlcyI6dHJ1ZSwiYXJncyI6eyJhcmdzIjoie30ifX0%3D and go back and forth between Form and YAML repeatedly

It'll start to look like this:
Screen Shot 2024-01-16 at 11 40 20

### Desired behavior

Args don't grow

Dartoxian:
## Description:
This PR implements exact package name searching in the EMUI. This allows indexing errors to be surfaced in the UI, for example:

[exact-match.webm](https://github.com/kurtosis-tech/kurtosis/assets/4419574/8cf29524-59de-44fc-834b-c07852406a12)

This PR also fixes #2051 by treating arguments with no argtype (as is the case for the `args` argument) as a `JSON` argument. This behaviour for Form->kurtosis format is now consistent with the existing kurtosis format->Form transformer.

YES

- Fixes https://github.com/kurtosis-tech/kurtosis/issues/2051


adschwartz (approved):
I'm ok with this change, although I wonder if showing as an explicit error when writing the name field is a little much? But will let you decide

PR #2102, opened 2024-01-31
feat: image load

lostbean:
## Description:
Load container images directly into Kurtosis. 

NO

### DIFF
--- a/container-engine-lib/lib/backend_interface/objects/service/service_config.go
+++ b/container-engine-lib/lib/backend_interface/objects/service/service_config.go
@@ -2,7 +2,9 @@ package service
 
 import (
 	"encoding/json"
+
 	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/image_build_spec"
+	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/image_load"
 	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/image_registry_spec"
 	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/port_spec"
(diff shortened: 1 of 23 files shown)


container-engine-lib/lib/backend_interface/objects/service/service_config.go
```diff
@@ -28,7 +30,11 @@ type privateServiceConfig struct {
 	// Configuration for container engine to pull an in a private registry behind authentication
 	// If nil, we will use the ContainerImageName and not use any auth
 	// Mutually exclusive from ImageBuildSpec, ContainerImageName
+
 	ImagerRegistrySpec *image_registry_spec.ImageRegistrySpec
+	// Reference to a image file that can be loaded into a backend image
+	// This is one of the options to reference a image and this value is optional (i.e. nil)
+	ImageLoad *image_load.ImageLoad
```
mieubrisse, line 37: Is there a usecase for this? I'd imagine users would be either a) specifying an existing Docker image, or b) getting an image through indirect means (either `ImageBuildSpec` or `NixBuildSpec`)

core/server/api_container/server/startosis_engine/kurtosis_types/service_config/image_load_spec.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/kurtosis-tech/kurtosis/core/server/api_container/server/startosis_engine/startosis_packages"
+	"go.starlark.net/starlark"
+)
+
+const (
+	ImageLoadTypeName = "ImageLoad"
+	LoadImageAttr     = "image_file_path"
+)
+
+func NewImageLoadType() *kurtosis_type_constructor.KurtosisTypeConstructor {
```
mieubrisse, line 21: Same question as above - makes a lot of sense to me that a KurtosisBackend needs the ability to load images, but do we expect users to be writing Starlark that references an image .tgz?

lostbean:
deprecated in favor of Nix builds: #2132

PR #2117, opened 2024-02-05
feat: Enclave connect

laurentluce:
## Description:
We want the user to be able to connect from their local machine to an enclave with a start from the UI.  A connect button is added to the enclave page with the CLI commands to paste.  A CLI enclave connect command is added.

YES

(https://www.notion.so/kurtosistech/Connect-local-CLI-to-running-enclave-workflow-for-cloud-based-enclaves-via-frontend-636d41ca937c4e49b54161aa498704b1?pvs=4)


cli/cli/commands/cloud/load/load.go
```diff
@@ -50,6 +51,18 @@ func run(ctx context.Context, _ *flags.ParsedFlags, args *args.ParsedArgs) error
...
+	contextsConfigStore := store.GetContextsConfigStore()
+	currentContext, err := contextsConfigStore.GetCurrentContext()
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred while retrieving the current context")
+	}
+
+	if currentContext.Uuid.Value == instanceID {
+		logrus.Infof("Cloud instance %s already loaded", instanceID)
+		return nil
+	}
```
laurentluce, line 64: Make the cloud load command idempotent.

api/typescript/src/core/lib/enclaves/enclave_context.ts
```diff
@@ -304,6 +304,8 @@ export class EnclaveContext {
         return ok(serviceContext);
     }
 
+    // TODO: Add getServiceContexts
```
adschwartz, line 307: Todo for the future?

laurentluce: Yes i want to push the current changes asap so they can be used during user interviews.

cli/cli/commands/enclave/connect/connect.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred creating Kurtosis Context from local engine")
+	}
+
+	enclaveCtx, err := kurtosisCtx.GetEnclaveContext(ctx, enclaveIdentifier)
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred getting an enclave context from enclave info for enclave '%v'", enclaveIdentifier)
+	}
+	allServicesMap := map[string]bool{}
+	serviceContexts, err := enclaveCtx.GetServiceContexts(allServicesMap)
```
adschwartz, line 103: while the err check below should handle errors I guess its still possible for someone to make this function return a nil for `serviceContexts` even if there's no `err`. We should check for that here

api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -353,26 +353,35 @@ func (enclaveCtx *EnclaveContext) GetServiceContext(serviceIdentifier string) (*
...
-		serviceInfo.GetMaybePublicIpAddr(),
-		serviceCtxPublicPorts,
-	)
+	serviceContexts := make(map[services.ServiceName]*services.ServiceContext, len(response.GetServiceInfo()))
+	for _, serviceInfo := range response.GetServiceInfo() {
+		serviceContext, err := enclaveCtx.convertServiceInfoToServiceContext(serviceInfo)
+		if err != nil {
+			return nil, stacktrace.Propagate(err, "An error occurred converting the service info to a service context.")
+		}
+		serviceContexts[serviceContext.GetServiceName()] = serviceContext
```
adschwartz, line 381: `serviceContexts` could be nil

cli/cli/commands/cloud/load/load.go
adschwartz: nice!

adschwartz: Is it possible you might need to reload the config? If you do I guess you have to jus switch to default and back to this remote context?

adschwartz (approved):
Looks good, just a few comments to check out

cli/cli/commands/cloud/load/load.go
laurentluce: Yes and you cannot reload today, it fails on the remove context operation saying: "you cannot remove the current context."  More work would be required to reload.

Issue #2121, opened 2024-02-05
Unable to reference the same Directory by different services within the same run


barnabasbusa:
### What's your CLI version?

0.86.12

### Description & steps to reproduce

This issue persist on docker and kubernetes.

I'd like to repopulate a persistent directory using the following service:
```py
                shadowfork_data = plan.add_service(
                    name="shadowfork-{0}".format(el_service_name),
                    config=ServiceConfig(
                        image="rclone/rclone:1.55.1",
                        cmd=[
                            "rclone copy -P mys3:ethpandaops-ethereum-node-snapshots/"
                            + base_network
                            + "/"
                            + el_client_type
                            + "/latest /data/"
                            + el_client_type
                            + "/execution-data"
                        ],
                        entrypoint=["/bin/sh", "-c"],
                        files={
                            "/data/"
                            + el_client_type
                            + "/execution-data": Directory(
                                persistent_key="data-{0}".format(el_service_name),
                                size=constants.VOLUME_SIZE[base_network][
                                    el_client_type + "_volume_size"
                                ],
                            ),
                        },
                        env_vars={
                            "RCLONE_CONFIG_MYS3_TYPE": "s3",
                            "RCLONE_CONFIG_MYS3_PROVIDER": "DigitalOcean",
                            "RCLONE_CONFIG_MYS3_ENDPOINT": "https://ams3.digitaloceanspaces.com",
                        },
                    ),
                )
```

later I want to use this same persistent directory to start up a EL service. 

Error message: 
```
[...]

PR #2123, opened 2024-02-06
fix: allow persistent directory to be reused across services

h4ck3rk3y:
There are workflows where people want to use prepopulated persistent volumes in different services. #2121 is an example of this workflow

This PR
- removes the Service UUID label from the PVCs / Persistent Volumes
- Allows PV to be re-used

Note in Kubernetes this will break for multi node clusters and this reuse of PVs will have to be in conjunction with node affinity selectors


tedim52:
This change is pretty interesting from a product perspective - makes the enclave behave more and more like a filesystem.

tedim52 (approved):

PR #2128, opened 2024-02-07
ci: faster build times

h4ck3rk3y:


internal_testsuites/golang/scripts/test.sh
```diff
@@ -17,8 +17,12 @@ KUBERNETES_TIMEOUT="510s" # K8S takes longer than docker
 TESTSUITE_CLUSTER_BACKEND_DOCKER="docker"
```
adschwartz, line 17: With this change we can potentailly take `KUBERNETES_TIMEOUT` back to `8ms`? Or maybe same time as docker even?

adschwartz (approved):

internal_testsuites/golang/scripts/test.sh
h4ck3rk3y: With the run_sh changes; we can reduce this definitely! Will bring it back down :-)

Issue #2140, opened 2024-02-09
FR: Allow me to not wait for pod termination in k8s


barnabasbusa:
### Background & motivation

For each task, a new pod is spun up. When this tasks completes (within a couple of seconds), the pod needs to wait for termination, which can take up to 30 seconds. 

While I understand that this feature is there to ensure that each run will stay deterministic, I'd prefer if there would be a way I could pass an extra flag during runtime where I can specify not to wait for termination.

Note: This is only relevant with k8s backend.

### Desired behaviour

`kurtosis run . --fast-mode...` would enable pods to be created, and proceed to the next plan item before waiting for pod termination. 

### How important is this to you?

Painful; the lack of this feature makes using Kurtosis frictionful.

### What area of the product does this pertain to?

CLI: the Command Line Interface

galenmarchetti:
thanks for submitting a suggestion for design here, as well - i like it. ensures deterministic runs at the level of package definition while giving the end user a way to break the assumptions for a faster run, understanding you might get breakage running it 👍 checking how hard this would be

tedim52:
Hey @barnabasbusa here's a PR for that: https://github.com/kurtosis-tech/kurtosis/pull/2153

h4ck3rk3y:
Closed by @tedim52

PR #2153, opened 2024-02-13
feat: kurtosis run `--non-blocking-tasks`

tedim52:
## Description:
Adds a flag `--non-blocking-tasks`. If set, task services from `run_sh` and `run_python` are not removed before continuing with execution. The effects is small over Docker, but this significantly reduces run times over k8s because of the overhead of removing a pod (up to 30s).

For the `ethereum-package` over k8s with images cached, running with the flag goes from 4min to 1min. However, task services will have to be removed manually after execution or via a `kurtosis clean`.

YES

## References:
https://github.com/kurtosis-tech/kurtosis/issues/2140


core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
```diff
@@ -297,7 +297,7 @@ func formatErrorMessage(errorMessage string, errorFromExec string) string {
 func removeService(ctx context.Context, serviceNetwork service_network.ServiceNetwork, serviceName string) error {
 	_, err := serviceNetwork.RemoveService(ctx, serviceName)
 	if err != nil {
-		return stacktrace.NewError("error occurred while removing task with name %v", serviceName)
+		return stacktrace.Propagate(err, "error occurred while removing task with name %v", serviceName)
```
leoporoli, line 300: Nice!

leoporoli (approved):

PR #2156, opened 2024-02-13
docs: add in icons and rearrange order

galenmarchetti:
## Description:
why kurtosis first, then helm and dcompose q

YES


chunha-park (approved):
SHIP

PR #2208, opened 2024-02-22
feat: allow users to validate their doc strings via linter

### LINKED ISSUE #2148: move doc string validation to Kurtosis
### Background & motivation

For now we validate doc strings inside the kurtosis-package-indexer / cloud-catalog but that is no good as the loop to fix a doc string is very long.

### Desired behaviour

If there's a doc string kurtosis run should validate that too

h4ck3rk3y:
Closes #2148


cli/cli/commands/lint/lint.go
```diff
@@ -40,6 +46,8 @@ const (
 	presentWorkingDirectory = "."
 	versionArg              = "version"
 
+	mainDotStarFilename = "main.star"
```
leoporoli, line 49: There is a flag `--main-file`  in the `kurtosis run` cmd that allow users to set a different file as the main file, I wonder if we should include this flag here too? 
At the same time I'm not really sure if the indexer has this distinction, as far as I remember the indexer always look for the main.star file to parse the docstring.
And if we [check the docs](https://docs.kurtosis.com/advanced-concepts/packages/#runnable-packages) the `main.star` file is mandatory for runnable packages
I'm a little confuse with this point.
WDYT?

cli/cli/commands/lint/lint.go
```diff
@@ -166,3 +193,50 @@ func getVolumeToMountAndPathToLint(pathOfFileOrDirToLint string) (string, string
...
+
+func parseDocString(mainStarFilepath string) error {
+	contents, err := os.ReadFile(mainStarFilepath)
+	if err != nil {
+		return stacktrace.Propagate(err, "")
+	}
+	contentsAsStr := string(contents)
+	currentOutput := logrus.StandardLogger().Out
+	defer logrus.SetOutput(currentOutput)
+	// we disable the output as the crawler pollutes the screen otherwise
```
leoporoli, line 236: nice!

cli/cli/commands/lint/lint.go
leoporoli: Let's keep it `main.star` I think it's valid and then we can discuss with the team about the `--main-file` flag, if we should rename it to avoid confusions.

leoporoli (approved):

cli/cli/commands/lint/lint.go
h4ck3rk3y: This makes sense! We can iteratively add it later.

PR #2217, opened 2024-02-26
docs: What's New 2024-02-26

mieubrisse:
## Description:
I want to be pushing "What's new in Kurtosis" every week. This is the first week.

YES


chunha-park (approved):
Looks great! Thanks for doing this :)

PR #2226, opened 2024-02-27
fix: remove package json

tedim52:
## Description:
These files aren't need at the base of the monorepo and we're mistakenly added in previous commits.

NO


go.work.sum
```diff
@@ -30,12 +30,15 @@ cloud.google.com/go/automl v1.13.1 h1:iP9iQurb0qbz+YOOMfKSEjhONA/WcoOIjt6/m+6pIg
 cloud.google.com/go/automl v1.13.1/go.mod h1:1aowgAHWYZU27MybSCFiukPO7xnyawv7pt3zK4bheQE=
```
h4ck3rk3y, line 30: why is this here?

h4ck3rk3y (approved):
LGTM, is the go work intentional?

go.work.sum
tedim52: the `go.work` always gets changed with every change and I always commit it bc Go instructs you to but I don't think it has any effect.

PR #2233, opened 2024-02-28
feat: added the get_service instruction

h4ck3rk3y:
allows you to get a service object by name; useful if you lose a reference to a service or are using package imports

https://www.notion.so/kurtosistech/get_service-and-port-url-b81ad4110c294ddc8851a7ec0ffe2932


core/server/api_container/server/startosis_engine/interpretation_time_value_store/service_value_repository.go
```diff
@@ -0,0 +1,113 @@
...
+		// check for existence
+		if serviceSerializedValue == nil {
+			return stacktrace.NewError("Service '%v' doesn't exist in the repository", name)
+		}
+
+		isEmptyValue := len(serviceSerializedValue) == len(emptyValue)
+
+		serviceSerializedValueStr := string(serviceSerializedValue)
+
+		// this will the case if the key was saved with an empty value
```
tedim52: `this will catch the case`?

core/server/api_container/server/startosis_engine/interpretation_time_value_store/service_value_repository.go
```diff
@@ -0,0 +1,113 @@
...
+
+	repository := &serviceInterpretationValueRepository{
+		enclaveDb:          enclaveDb,
+		starlarkValueSerde: starlarkValueSerde,
+	}
+
+	return repository, nil
+}
+
+func (repository *serviceInterpretationValueRepository) AddService(name service.ServiceName, service *kurtosis_types.Service) error {
```
tedim52: nit: suggest renaming `AddService` to `PutService` bc `AddService` means something else in the startosis engine (add in service network) so it's an overloaded concept

core/server/api_container/server/startosis_engine/interpretation_time_value_store/service_value_repository.go
tedim52: What cases would it be valid to store `service` with an empty value? If it's not valid, maybe we can prevent it from being stored as an empty value in `AddService`?

core/server/api_container/server/startosis_engine/interpretation_time_value_store/service_value_repository_test.go
```diff
@@ -0,0 +1,93 @@
...
+	err := repository.AddService(serviceName, expectedService)
+	require.Nil(t, err)
+
+	actualService, err := repository.GetService(serviceName)
+	require.Nil(t, err)
+	require.Equal(t, expectedService.AttrNames(), actualService.AttrNames())
+	require.Equal(t, expectedService.String(), actualService.String())
+}
+
+func getRecipeResultRepositoryForTest(t *testing.T) *serviceInterpretationValueRepository {
```
tedim52: is `Recipe` supposed to indicate a template of some sort?

core/server/api_container/server/startosis_engine/interpretation_time_value_store/service_value_repository_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+const (
+	starlarkThreadName     = "thread-for-db-test"
+	serviceName            = service.ServiceName("datastore-1")
+	serviceNameStarlarkStr = starlark.String(serviceName)
+	hostName               = serviceNameStarlarkStr
+	ipAddress              = starlark.String("172.23.34.44")
+)
+
+func TestAddAndGetTest(t *testing.T) {
```
tedim52: maybe add a test for the service doesn't exist case and the case where the service was stored as an empty value to illustrate/test those behaviors?

core/server/api_container/server/startosis_engine/interpretation_time_value_store/service_value_repository.go
h4ck3rk3y: Yeah, i had got this from recipe but now i am throwing an error!

tedim52 (approved):

PR #2222, opened 2024-02-27
docs: Basic Concepts cleanup

mieubrisse:
APPROVE #2221 FIRST!

## Description:
The "basic concepts" section was using the CLI, which is 5/10 interesting. Switched this to screenshots of the app, which is 7.5/10 interesting.

Also fixed the issue #2017 was trying to address.

## Is this change user facing?
YES

PR #2242, opened 2024-02-29
fix: moving the golangci-lint condition because it has matrix strategy

leoporoli:
## Description:
moving the golangci-lint condition because it has matrix strategy

We have to move the condition to the steps section otherwise these jobs won't be executed when a doc's PR is created.

NO


h4ck3rk3y:
@leoporoli i think it being on the top section is okay; we don't want the linter to run if its only a docs change

leoporoli:
> @leoporoli i think it being on the top section is okay; we don't want the linter to run if its only a docs change

@h4ck3rk3y it doesn't work if we don't move it, you can see this example PR> https://github.com/kurtosis-tech/kurtosis/pull/2222

It doesn't find all the golang-lint jobs because the matrix is not generated, and the branch protection rule is waiting for all of them, so we have to run them, but with the condition in all the steps they will run fast

h4ck3rk3y (approved):

PR #2254, opened 2024-03-04
fix: replace env var with runtime values in run_sh

tedim52:
## Description:
Env vars weren't being replaced with runtime value in `run_sh` eg. 
```
def run(plan, args):
    database = plan.add_service(name="database", config=ServiceConfig(image="postgres:latest"))

    plan.run_sh(
        run="echo $VAR_1",
        env_vars={
            "VAR_1": database.ip_address
        }
    )
 ```
 used to output
 ```
 Command returned with exit code '0' and the following output:
--------------------
{{kurtosis:b49259098950411ba1e409e8a749c562:ip_address.runtime_value}}
--------------------
```
but now we get:
```
 Command returned with exit code '0' and the following output:
--------------------
172.16.24.5
--------------------
```

YES


core/server/api_container/server/startosis_engine/kurtosis_instruction/add_service/add_service_shared.go
```diff
@@ -188,9 +188,6 @@ func replaceMagicStrings(
 	if serviceConfig.GetEnvVars() != nil {
 		envVars = make(map[string]string, len(serviceConfig.GetEnvVars()))
 		for envVarName, envVarValue := range serviceConfig.GetEnvVars() {
-			if err != nil {
```
tedim52, line 191: these lines of code weren't actually doing anything

h4ck3rk3y (approved):

Issue #2268, opened 2024-03-07
feat: HTTP traffic should be routable to the user services using more than one Host header format


laurentluce:
### Background & motivation

The request Host header reaching the user service should match the original request Host header so URLs generation and redirects work correctly.  This should happen locally and in the cloud.

### Desired behaviour

Traefik routes http traffic to the user services.  The user service Host header label should accept the following two formats:
- `<port number>.<service short uuid>.<enclave short uuid>`
- `<port number>.<service short uuid>.<enclave short uuid>.<cloud instance short uuid>.<subdomain>`

Method to update: https://github.com/kurtosis-tech/kurtosis/blob/c640c5eae57429de2a827e64de0175a111fc298d/container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go#L533

### How important is this to you?

Critical; Kurtosis is unusable for me without it.

### What area of the product does this pertain to?

Other: anything not covered by the above

leoporoli:
It was fixed with those PRs, and it supports retro-compatibility.

daramir:
@leoporoli was this tested?
I'm not sure if the Traefik Golang Regexp for Host matching 
```
HostRegexp(`{^name:%d-%s-%s-?%s$}`)
``` 

is correct?

Is the caret in  `^name` part of the  ["arbitrarily named variable followed by the colon-separated regular expression"](https://doc.traefik.io/traefik/v2.0/routing/routers/)? 

I have a kurtosis setup running on docker that I would like to use the reverse proxy feature for, but it's unusable... I'm unable to reach those hosts on the Traefik configured routes.

PR #2269, opened 2024-03-07
feat: updated Traefik config rule matcher in order to use a HostRegex matcher instead of Host

leoporoli:
## Description:
Created to solve this ticket https://github.com/kurtosis-tech/kurtosis/issues/2268
Still pending a config upgrade in the cloud gateway also

NO

### DIFF
--- a/container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go
+++ b/container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go
@@ -30,6 +30,8 @@ const (
 	logsCollectorFragment                  = "kurtosis-logs-collector"
 	// The collector is per enclave so this is a suffix
 	logsCollectorVolumeFragment = logsCollectorFragment + "-vol"
+
+	anyCharacterPrefixRegexToken = ".*"
 )
 
 type DockerEnclaveObjectAttributesProvider interface {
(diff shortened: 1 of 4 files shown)


container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go
```diff
@@ -544,7 +546,7 @@ func (provider *dockerEnclaveObjectAttributesProviderImpl) getTraefikLabelsForEn
 			shortServiceUuid := uuid_generator.ShortenedUUIDString(serviceUuid)
 			servicePortStr := fmt.Sprintf("%s-%s-%d", shortEnclaveUuid, shortServiceUuid, portSpec.GetNumber())
 
-			labelKeyValuePairs[fmt.Sprintf("http.routers.%s.rule", servicePortStr)] = fmt.Sprintf("Host(`%d-%s-%s`)", portSpec.GetNumber(), shortServiceUuid, shortEnclaveUuid)
+			labelKeyValuePairs[fmt.Sprintf("http.routers.%s.rule", servicePortStr)] = fmt.Sprintf("HostRegexp(`{name:%d-%s-%s%s}`)", portSpec.GetNumber(), shortServiceUuid, shortEnclaveUuid, anyCharacterPrefixRegexToken)
```
laurentluce: Do we need the group label `name`?

laurentluce: Should we add the beginning and ending characters `^` and `$`:
`^81-3771c85af16a-65d2fb6d6732.*$`

laurentluce (approved):

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go
laurentluce: Should we add `-?` before `.*` so the enclave Id is delimited:
`^81-3771c85af16a-65d2fb6d6732-?.*$`

leoporoli: @laurentluce yes, we need the group label name otherwise it will works only for the subdomain part by default.

Thanks for all the other suggestions I've incorporated all them in the matcher, I think the ^ and $ characters are not mandatory because the app add them if they are not present, but it's also works if we add them, so I'm adding them in order to have a better and cleaner regex.

PR #2271, opened 2024-03-08
fix: image download mode always applies to k8s as well

### LINKED ISSUE #2246: --image-download always doesn't work on k8s
### What's your CLI version?

0.87.3

If I use k8s as the backend for my kurtosis run, image-download always will be ignored, and the old images will be used.

### Desired behavior

If I set `--image-download always` flag, the pod image pull policy should be set to `Always` instead of `IfNotPresent`

h4ck3rk3y:
Earlier this flag was limited to Docker but now it works on k8s too

Closes #2246 

Earlier `latest` defaulted to `Always` but now it will depend on the flag sent 

https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2764-L2771


container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/start_user_services.go
```diff
@@ -3,6 +3,7 @@ package user_services_functions
 import (
 	"context"
 	"fmt"
+	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/image_download_mode"
```
leoporoli, line 6: will you add the docs in another PR?

leoporoli (approved):

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/start_user_services.go
h4ck3rk3y: Its already in docs :-)

PR #2282, opened 2024-03-13
feat: enable import module for docker compose packages

tedim52:
YES/NO


tedim52:
implemented in https://github.com/kurtosis-tech/kurtosis/pull/2327

PR #2299, opened 2024-03-14
fix: the package replace logic when using tag, branch or commit

### LINKED ISSUE #2279: fix bug in package replace that breaks versioned imports
### What's your CLI version?

0.88.3

I was on the chainlink node package in awesome kurtosis and i was trying to make it depend on 1.4.0 via Package replace but it failed with

https://app.circleci.com/pipelines/github/kurtosis-tech/awesome-kurtosis/1620/workflows/22629b4c-e7c5-419c-97e3-9726bca1f34b/jobs/4232?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

Here is the change I made - https://github.com/kurtosis-tech/awesome-kurtosis/pull/308/commits/f9638dd24e1a8b51f808985891d5af160c292cf5

It fails with tags too!

### Desired behavior

Replace with specific versions should work

leoporoli:
## Description
Fix the package replace logic when using tag, branch or commit

The current issue is the `package replace logic` fails when you add a tag, branch or commit in the replace value, like this example:

`replace:
            - github.com/kurtosis-tech/ethereum-package: github.com/kurtosis-tech/ethereum-package@da55be84861e93ce777076e545abee35ff2d51ce`

If we see the evaluation error 

`Evaluation error: An error occurred while loading the module 'github.com/kurtosis-tech/ethereum-package@da55be84861e93ce777076e545abee35ff2d51ce/src/package_io/input_parser.star'
        Caused by: '/input_parser.star' doesn't exist in the package 'kurtosis-tech/ethereum-package'
        at [github.com/kurtosis-tech/ethereum-package@da55be84861e93ce777076e545abee35ff2d51ce/main.star:1:29]: <toplevel>
        at [github.com/kurtosis-tech/awesome-kurtosis/chainlink-node/main.star:2:33]: <toplevel>`

we understand it's trying to import the file from this path `github.com/kurtosis-tech/ethereum-package@da55be84861e93ce777076e545abee35ff2d51ce` and it doesn't exist because the folder, inside the APIC file system, is created without the hash, it's `github.com/kurtosis-tech/ethereum-package` 

This is happening because an absolute locator like this `github.com/kurtosis-tech/ethereum-package/src/package_io/input_parser.star` does not incorporate the concept of hash, branch or commit.

The fix is incorporated the concept of `hash, branch or commit` in a new AbsoluteLocator object, so this way we can specify if the absolute locator is pointing to other place than the main branch.
            
With this change we can accept the `github.com/kurtosis-tech/ethereum-package@da55be84861e93ce777076e545abee35ff2d51ce` replace value and an absolute locator string like this `github.com/kurtosis-tech/ethereum-package@da55be84861e93ce777076e545abee35ff2d51ce/src/package_io/input_parser.star` will be represented with an AbsoluteLocator object with locator `github.com/kurtosis-tech/ethereum-package/src/package_io/input_parser.star` and and tagBranchOrCommit `da55be84861e93ce777076e545abee35ff2d51ce` 

We already had support for retrieving packages from different tag, branch or commit but only for `packageId` values and not for `absolute locators`

YES

## References (if applicable)
Fix #2279


core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
```diff
@@ -211,10 +236,37 @@ func TestGitPackageProvider_FailsForNonExistentPackage(t *testing.T) {
...
 
-	_, err = provider.GetModuleContents(nonExistentModulePath)
+	nonExistentModuleAbsoluteLocator := startosis_packages.NewPackageAbsoluteLocator(nonExistentModulePath, defaultMainBranch)
+
+	_, err = provider.GetModuleContents(nonExistentModuleAbsoluteLocator)
 	require.NotNil(t, err)
 }
 
+func TestGitPackageProvider_GetContentFromAbsoluteLocatorWithCommit(t *testing.T) {
+	oackageDir, err := os.MkdirTemp("", packagesDirRelPath)
```
tedim52: `package`

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
```diff
@@ -211,10 +236,37 @@ func TestGitPackageProvider_FailsForNonExistentPackage(t *testing.T) {
...
+	githubAuthTokenFilePath, err := os.CreateTemp(githubAuthDir, githubAuthTokenFilename)
+	require.Nil(t, err)
+	defer os.RemoveAll(githubAuthDir)
+
+	provider := NewGitPackageContentProvider(packageDir, packageTmpDir, githubAuthTokenFilePath.Name(), nil)
+
+	absoluteLocatorStr := "github.com/kurtosis-tech/ethereum-package/src/package_io/input_parser.star"
+	commitHash := "da55be84861e93ce777076e545abee35ff2d51ce"
+
+	absoluteLocatorWithCommit := startosis_packages.NewPackageAbsoluteLocator(absoluteLocatorStr, commitHash)
```
tedim52, line 263: It's a bit unclear to me what the behavior was prior to enabling this. Did we not handle retrieving git repos from off of branches, commits, and hashes? I thought we had that actually. 

Maybe we can make this more explicit by adjusting this test to show the behavior for an absolute locator with no specified tag, branch, or commit, and one with?

I'd imagine we could do this by comparing that the contents from each are different - and requiring they equal to what we'd expect. What do you think?

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
```diff
@@ -211,10 +236,37 @@ func TestGitPackageProvider_FailsForNonExistentPackage(t *testing.T) {
...
+	defer os.RemoveAll(packageTmpDir)
+	githubAuthDir, err := os.MkdirTemp("", githubAuthDirRelPath)
+	require.Nil(t, err)
+	githubAuthTokenFilePath, err := os.CreateTemp(githubAuthDir, githubAuthTokenFilename)
+	require.Nil(t, err)
+	defer os.RemoveAll(githubAuthDir)
+
+	provider := NewGitPackageContentProvider(packageDir, packageTmpDir, githubAuthTokenFilePath.Name(), nil)
+
+	absoluteLocatorStr := "github.com/kurtosis-tech/ethereum-package/src/package_io/input_parser.star"
```
tedim52, line 260: It's not great that we have tests that rely on external resources, I guess because the absolute locator reference to a pinned commit this is fine, but if you decide to update the test to account for my comment below, might have to reference the main branch which is susceptible to change. I wonder if we can mock git repos somehow or maybe refer to a package in this repo? but maybe thats for a separate change, I'll let you decide whats best here

tedim52 (approved):

tedim52:
approved! Is there any chance you can provide more context on what the specific bug around locators was and what the specific fix was for it was in the PR description?

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
leoporoli: We already had support for retrieving packages from different tag, branch or commit but only for `packageId` values and not for `absolute locators`

I'm adding that test, good.

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_content_provider_test.go
leoporoli: Yes it's not the best, now I'm using an archived repository to handle this problem, but it's not the final solution.

Doing the refactor is bigger than this change and probably out of the scope of this issue and I don't want to block it in order to unblock the user, so I created this ticket that we can prioritize later> https://github.com/kurtosis-tech/kurtosis/issues/2310

leoporoli:
> approved! Is there any chance you can provide more context on what the specific bug around locators was and what the specific fix was for it was in the PR description?

@tedim52 I added more context in the description, I hope this can put more light in the issue and the fix.

tedim52:
Sweet, thanks for adding context to the PR, LGTM!

PR #2316, opened 2024-03-20
feat: enclave manager has an endpoint to create a webhook

h4ck3rk3y:
This endpoint only works if a GitHub auth token is set and the enclave manager detects a `_prevenv_web_token`


h4ck3rk3y:
https://github.com/kurtosis-tech/kurtosis/assets/6670312/5ffef9e5-5932-49ff-ae29-2c198726d39d

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/consts/consts.go
```diff
@@ -33,6 +33,7 @@ const (
 	HttpApplicationProtocol                             = "http"
 
 	GitHubAuthStorageDirPath = "/kurtosis-data/github-auth/"
+	GithubAuthStorageToken   = "token.txt"
```
tedim52, line 36: `s/GithubAuthStorageToken/GithubAuthTokenFileName` 

seperate - out of scope of this PR and not related to your change but just calling out- at this point LOTS of the codebase relies on  the hardcoded`/kurtosis-data` directory path. I believe it's originally set as env var to the APIC, but somehow the directory path needs to be refactored and propagated to the elements that need it.

h4ck3rk3y: Agree with you! I just followed the current program to get something out but we should refactor ( when we have the time and have earned the right too)

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/github_auth_storage_creator/github_auth_storage_creator.go
```diff
@@ -95,6 +96,15 @@ func (creator *GitHubAuthStorageCreator) CreateGitHubAuthStorage(
 	return nil
 }
 
+// GetBestEffortGitHubAuthToken Returns empty string if no token found in [githubAuthTokenFile] or [githubAuthTokenFile] doesn't exist
+func GetBestEffortGitHubAuthToken() string {
```
tedim52: nit: maybe? `s/GetBestEffort/Get`, I think if it's documented in the function spec/comment that behavior is empty string when no token, we don't need to add `BestEffort`. will leave up to you on this

tedim52 (approved):

PR #2326, opened 2024-03-25
fix!: copy contents of directory instead of the directory if store is used for directories

### LINKED ISSUE #2227: plan/run_sh/store: providing a directory path doesn't copy the content of it
### What's your CLI version?

0.86.12

1. Try to run an bash plan command 
2. Provide a store path, so the content can be mounted for the plan.add_service part later
```
results = plan.run_sh(
        # image and others 
        store=[
            "/foo/bar/",
        ],
    )
 plan.print(results.files_artifacts)
 ```

### Desired behavior

Content under bar is stored in the context for mounting in later stages like plan.add_service()

Note: adding an asterisk to /foo/bar/* fixes this behaviour for now

SDK: the Software Development Kit libraries - Typescript, Go, etc.

h4ck3rk3y:
Closes #2227


leoporoli (approved):

barnabasbusa:
Before this gets merged we have to have a PR ready to go for all major kurtosis packages, and all CI jobs have to be tested with this new version, as I expect it to break a lot of things!

PR #2336, opened 2024-03-26
feat: change cloud protobuf to allow for whitelisting ports as public

h4ck3rk3y:


cloud/api/protobuf/kurtosis_backend_server_api.proto
```diff
@@ -139,3 +143,31 @@ message UpdateAddressArgs {
   string postal_code = 7;
   string country = 8;
 }
+
+message LockUnlockPortRequest {
+  uint32 port_number=1;
+  string instance_short_uuid=2;
+  string service_short_uuid=3;
+  string enclave_short_uuid=4;
```
laurentluce: Should we make those four attributes a message named `Port` and use that in some of the requests and responses?  It could also replace `PortAndService`.

cloud/api/protobuf/kurtosis_backend_server_api.proto
```diff
@@ -19,6 +19,10 @@ service KurtosisCloudBackendServer {
   rpc RefreshDefaultPaymentMethod (RefreshDefaultPaymentMethodArgs) returns (google.protobuf.Empty) {};
   rpc CancelPaymentSubscription (CancelPaymentSubscriptionArgs) returns (google.protobuf.Empty) {};
   rpc UpdateAddress(UpdateAddressArgs) returns (google.protobuf.Empty) {};
+  rpc CheckPortAuthorization(CheckIsPortAuthorizedArgs) returns(google.protobuf.Empty){};
```
laurentluce: Can we have the args name match the method name.

laurentluce (approved):

PR #2342, opened 2024-03-27
fix: enclave manager passing in deprecated args object

tedim52:
## Description
The enclave builder would fail to render packages that used `def run(plan, args)` where starlark in the package depended on values in `args` being set. This was because the enclave manager was passing args to interpretation using via an `args`. The use of `args` dict as sole input is deprecated.

While it's deprecated we still support it for backwards compatibility but everything in `args` is dropped - which caused the failure. This PR fixes it by unpacking the values in `args` so enclave builder passes them as unpacked `kwargs` to interpretation, as opposed to using the sole deprecated `args` dict.

I've also added a test to highlight this behavior.

```
def run(plan, args):
	all_arg_values = args["arg1"] + ":" + args["arg2"]
	return all_arg_values
```
will error when passing input with only `args`
`{"args": {"arg1": "arg1-value", "arg2": "arg2-value"}}`

NO


enclave-manager/web/packages/app/src/emui/enclaves/components/enclaveBuilder/nodes/KurtosisPackageNode.tsx
```diff
@@ -40,6 +40,14 @@ export const KurtosisPackageNode = memo(
             setMode({ type: "error", error: "APIC info missing from temporary enclave" });
             return;
           }
+
+          // If args only has one record, and its value is args, ASSUME user is passing args via a JSON or YAML into the args object of def run(plan, args)
+          // via Json or Yaml editor
+          // If only an `args` object is provided, kurtosis will not interpret the value in the args object as passing args via the args dictionary is (technically) deprecated even though it's still allowed
+          if (Object.keys(args).length === 1 && args.hasOwnProperty('args')) {
+            args = args['args'] as Record<string, string> // TODO(tedi): ideally we'd validate and handle this in transform args utils
```
tedim52: opted not to adjust the transformation logic (mainly in utils.tsx `transformFormArgsIntoKurtosisArgs` function as it's a bit complex and I'm not familiar with our frontend so didn't want to break other codepaths, but this case should probably be handled there

leoporoli (approved):

PR #2355, opened 2024-04-03
feat: enable setting git token per package run

tedim52:
## Description
Enables setting an github auth token per package.

NO


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/github_auth_storage_creator/github_auth_storage_creator.go
```diff
@@ -28,23 +28,18 @@ const (
 	authStorageCreationCmdMaxRetries     = 2
 	authStorageCreationCmdDelayInRetries = 200 * time.Millisecond
 
+	githubAuthTokenFilePath = "token.txt"
+
 	sleepSeconds = 1800
 )
 
-type GitHubAuthStorageCreator struct {
```
tedim52, line 34: remove the object here in favor of functions, as object wasn't providing any benefit

core/server/api_container/server/startosis_engine/startosis_packages/git_package_content_provider/git_package_auth_provider.go
```diff
@@ -0,0 +1,59 @@
...
+		return err
+	}
+	logrus.Infof("Successfully stored GitHub auth token for package: '%v'", packageId)
+	return nil
+}
+
+func (gitAuth *GitHubPackageAuthProvider) GetGitHubTokenForPackage(packageId string) string {
+	tokenBytes, err := os.ReadFile(path.Join(gitAuth.githubAuthStorageDirPath, getGitHubTokenFileName(packageId)))
+	if err != nil {
+		return ""
```
tedim52, line 41: opted to not return an error here to make the upstream err checking simpler but maybe I should? or at least log it?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/github_auth_storage_creator/github_auth_storage_creator.go
```diff
@@ -28,23 +28,18 @@ const (
 	authStorageCreationCmdMaxRetries     = 2
 	authStorageCreationCmdDelayInRetries = 200 * time.Millisecond
 
+	githubAuthTokenFilePath = "token.txt"
```
lostbean, line 31: Just wondering if storing the token as a file artifact is still need for the per-run-call configuration, since the token in meant to expire and the user should provide a new token every time. But I guess is meant the older engine provided token option that is still supported, correct?

lostbean (approved):
looks good, just left a comment

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/github_auth_storage_creator/github_auth_storage_creator.go
tedim52: `githubAuthTokenFilePath` is the path where the token provided to an engine is stored in the `github-auth-storage` volume - I wanted to keep this auth option available for those that don't want to auth `run` call. The per-run-run token is stored in a `package-id-token.txt` file in the `github-auth-storage` volume - this overrides the engine token if it exists. Does that answer your question? Maybe I can make this behavior more explicit?

Issue #2057, opened 2024-01-17
add ability to update the running image, parameters, and configuration of a given service


h4ck3rk3y:
### Background & motivation

https://github.com/kurtosis-tech/ethereum-package/issues/465
https://github.com/kurtosis-tech/ethereum-package/issues/458

There have been some requests from users of the Ethereum Package to change a service in the network when it comes to its underlying image or arguments so that they don't have to recreate the testnet

### Desired behaviour

An `update_service` api that allows you to change image or other arguments of a given service

### How important is this to you?

Nice to have; this feature would make using Kurtosis more enjoyable.

### What area of the product does this pertain to?

CLI: the Command Line Interface

h4ck3rk3y:
idempotent runs might solve for this!

mieubrisse:
Hey @h4ck3rk3y ! yep, I think enclave updating (formerly known as idempotent runs 😆 ) would solve for this - just re-throw an updated configuration at the enclave

h4ck3rk3y:
Yep! idempotent runs / updates do solve this. Everything but the affected instructions get skipped.

h4ck3rk3y:
Waiting for @barnabasbusa to confirm if this is enough for now; will close this if so

h4ck3rk3y:
Currently proposed workflow is arg tweaking and re-running against the same enclave

```
participants:
  - el_client_type: geth
    cl_client_type: lighthouse
  - el_client_type: geth
    cl_client_type: lighthouse
  - el_client_type: geth
    cl_client_type: lighthouse
    el_client_image: "ethereum/client-go:v1.13.10"
```

Changing the 10 to 11 and back to 10, seems to work on my machine

tedim52:
Enclave edits should address this issue but I've tried running an updated config against an existing enclave and enclave edits aren't working. Something triggers entire enclave to be rerun even when changing arg to just add an additional service. Unsure if it's something in the package or a bug in enclave edit logic. I've done a deep dive for a few hours and found some potential issues but haven't root caused yet. LMK if you're seeing otherwise though @h4ck3rk3y @barnabasbusa

PR #2372, opened 2024-04-11
feat: set service v0

tedim52:
## Description

Adds `set_service` instruction which can be used to override the service config of a service added earlier in the plan. This can be really useful for two cases:
1. A user wants to set a specific image (or other config) for a service in a package but the package author has not made the image configurable. This can be done by appending `set_service` to the plan.
ex.
```
postgres = import_module("github.com/kurtosis-tech/postgres-package/main.star")

def run(plan, args):
    postgres.run(plan)
    plan.set_service(name="postgres", config=ServiceConfig(image="postgres:bullseye"))
```
2. A user wants to swap out the image (or other config) of a service running in an enclave. This can be done by appending `set_service` to the plan and enclave editing.
ex. 
https://www.loom.com/share/f88d0f5555dd49b0aeabc55c5ed41d43

This instruction has some functionality missing such as overriding ports, env vars, and cmd/entrypoint as these require special handling of future references and handling consecutive set service instructions properly but functionality can be added iteratively in future PRs.

YES

## References:
https://github.com/kurtosis-tech/kurtosis/issues/2057


container-engine-lib/lib/backend_interface/objects/service/service_config.go
```diff
@@ -190,27 +217,51 @@ func (serviceConfig *ServiceConfig) GetMinCPUAllocationMillicpus() uint64 {
...
 
+func (serviceConfig *ServiceConfig) SetMinMemoryAllocationMegabytes(memoryAllocation uint64) {
+	serviceConfig.privateServiceConfig.MemoryAllocationMegabytes = memoryAllocation
+}
+
 func (serviceConfig *ServiceConfig) GetUser() *service_user.ServiceUser {
 	return serviceConfig.privateServiceConfig.User
 }
 
+func (serviceConfig *ServiceConfig) SetUser(user *service_user.ServiceUser) {
```
h4ck3rk3y, line 237: with so many setters i wonder whether can just have these variables as not private anymore;

core/server/api_container/server/startosis_engine/kurtosis_instruction/add_service/add_service.go
```diff
@@ -172,6 +172,8 @@ func (builtin *AddServiceCapabilities) Interpret(locatorOfModuleInWhichThisBuilt
 	if err != nil {
 		return nil, startosis_errors.WrapWithInterpretationError(err, "An error occurred while persisting return value for service '%v'", serviceName)
 	}
+	builtin.interpretationTimeValueStore.PutServiceConfig(builtin.serviceName, builtin.serviceConfig)
```
h4ck3rk3y, line 176: hmm i would think serialization could have failures

core/server/api_container/server/startosis_engine/kurtosis_instruction/add_service/add_service.go
```diff
@@ -308,7 +314,7 @@ func (builtin *AddServiceCapabilities) Description() string {
 	return builtin.description
 }
 
-func validateAndConvertConfigAndReadyCondition(
+func ValidateAndConvertConfigAndReadyCondition(
```
h4ck3rk3y: perhaps we can move this to a shared..

core/server/api_container/server/startosis_engine/kurtosis_instruction/set_service/set_service.go
```diff
@@ -0,0 +1,247 @@
...
+					IsOptional:        false,
+					ZeroValueProvider: builtin_argument.ZeroValueProvider[starlark.String],
+					Validator: func(value starlark.Value) *startosis_errors.InterpretationError {
+						return builtin_argument.NonEmptyString(value, ServiceNameArgName)
+					},
+				},
+				{
+					Name:              SetServiceConfigArgName,
+					IsOptional:        false,
+					ZeroValueProvider: builtin_argument.ZeroValueProvider[*service_config.ServiceConfig],
```
h4ck3rk3y, line 57: smart do we have this in add_service as well?

core/server/api_container/server/startosis_engine/kurtosis_instruction/set_service/set_service.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, startosis_errors.WrapWithInterpretationError(err, "Unable to extract value for '%s' argument", SetServiceConfigArgName)
+	}
+	rawImageVal, found, interpretationErr := kurtosis_type_constructor.ExtractAttrValue[starlark.Value](serviceConfigOverride.KurtosisValueTypeDefault, service_config.ImageAttr)
+	if interpretationErr != nil {
+		return nil, startosis_errors.WrapWithInterpretationError(err, "Unable to extract raw image attribute.")
+	}
+	if !found {
+		return nil, startosis_errors.NewInterpretationError("Unable to extract image attribute off of service config.")
+	}
+	builtin.imageVal = rawImageVal
```
h4ck3rk3y, line 127: just curious - how does this work with pulling of images; say amd64 on arch64, is the fetch logic the same?

core/server/api_container/server/startosis_engine/kurtosis_instruction/set_service/set_service.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func (builtin *SetServiceCapabilities) Validate(_ *builtin_argument.ArgumentValuesSet, validatorEnvironment *startosis_validator.ValidatorEnvironment) *startosis_errors.ValidationError {
+	if exists := validatorEnvironment.DoesServiceNameExist(builtin.serviceName); exists == startosis_validator.ComponentNotFound {
+		return startosis_errors.NewValidationError("Service '%v' required by '%v' instruction doesn't exist", builtin.serviceName, SetServiceBuiltinName)
+	}
+	return nil
+}
+
+func (builtin *SetServiceCapabilities) Execute(_ context.Context, _ *builtin_argument.ArgumentValuesSet) (string, error) {
```
h4ck3rk3y, line 165: do i need to do a restart service after this call for this to take effect?

core/server/api_container/server/startosis_engine/kurtosis_instruction/set_service/set_service.go
h4ck3rk3y: can we add some integration tests for this please?

h4ck3rk3y (approved):

core/server/api_container/server/startosis_engine/kurtosis_instruction/set_service/set_service.go
tedim52: yep, fetch logic is unaffected

core/server/api_container/server/startosis_engine/kurtosis_instruction/set_service/set_service.go
tedim52: yep, this validation function exists on `add_service`! and the the validator function gets tested in the `TestSetService` instruction framework test, but do you also suggest I create a separate test to test the failure case as well?

core/server/api_container/server/startosis_engine/kurtosis_instruction/add_service/add_service.go
tedim52: If you're referring to serde to `enclaveDb`, there's no serialization of `serviceConfig`s happening unlike with the `service` values. This logic works with the `ServiceConfig` objects are stored in memory.

container-engine-lib/lib/backend_interface/objects/service/service_config.go
tedim52: Def a good point. What do you think about doing this when we've enabled the ability to set all of the `ServiceConfig` values? I think there's value in keeping `Ports`, `EnvVars`, `Entrypoint`.. any ServiceConfig value affected by future references private until we have a way to handle setting them carefully.

core/server/api_container/server/startosis_engine/kurtosis_instruction/set_service/set_service.go
tedim52: Nope, similar to `get_service`, the execute is a no op. The effects of this instruction are seen during the execution stage when the `add_service` retrieves the service config from the `interpretationTimeValueStore`. 
This way, during an enclave edit, this will apply downstream effects that restarting the service may have, as opposed to just restarting the service.

core/server/api_container/server/startosis_engine/interpretation_time_value_store/interpretation_time_value_store.go
```diff
@@ -50,3 +56,32 @@ func (itvs *InterpretationTimeValueStore) RemoveService(name service.ServiceName
...
+func (itvs *InterpretationTimeValueStore) SetServiceConfig(name service.ServiceName, serviceConfig *service.ServiceConfig) {
+	itvs.setServiceConfigValues[name] = serviceConfig
+}
+
+func (itvs *InterpretationTimeValueStore) ExistsNewServiceConfigForService(name service.ServiceName) bool {
+	_, doesConfigFromSetServiceInstructionExists := itvs.setServiceConfigValues[name]
+	return doesConfigFromSetServiceInstructionExists
+}
+
+func (itvs *InterpretationTimeValueStore) GetNewServiceConfig(name service.ServiceName) (*service.ServiceConfig, error) {
```
tedim52, line 81: This API doesn't account for multiple `set_service` instructions properly and its a bit complex, gonna fix and add tests

core/server/api_container/server/startosis_engine/interpretation_time_value_store/interpretation_time_value_store.go
```diff
@@ -50,3 +56,32 @@ func (itvs *InterpretationTimeValueStore) RemoveService(name service.ServiceName
 	}
 	return nil
 }
+
+func (itvs *InterpretationTimeValueStore) PutServiceConfig(name service.ServiceName, serviceConfig *service.ServiceConfig) {
```
tedim52, line 60: This API  doesn't support a behavior where you have `add_service -> set_service -> set_service` and the config overrides are applied in order to the original config. Right now, only the latest `set_service` config overrides would be applied to the original config. I think that's okay for v0.

core/server/api_container/server/startosis_engine/startosis_interpreter_idempotent_test.go
```diff
@@ -479,6 +479,154 @@ func (suite *StartosisInterpreterIdempotentTestSuite) TestInterpretAndOptimize_U
 	require.True(suite.T(), scheduledInstruction4.IsExecuted()) // this instruction is not affected, i.e. it won't be re-run
 }
 
+func (suite *StartosisInterpreterIdempotentTestSuite) TestStartosisInterpreterIdempotent_SetService() {
```
h4ck3rk3y, line 482: how much does idempotency matter you think? 

we want - idempotency / edits 

is just edits enough? change image sure

h4ck3rk3y (approved):

core/server/api_container/server/startosis_engine/startosis_interpreter_idempotent_test.go
tedim52: Discussed internally over slack

PR #2373, opened 2024-04-11
feat: add and check port aliasing

h4ck3rk3y:


cloud/api/protobuf/kurtosis_backend_server_api.proto
```diff
@@ -222,14 +223,17 @@ message User {
   string Updated = 4;
 }
 
-message GetUnlockedPortsRequest {
+message GetPortsMetadataRequest {
```
h4ck3rk3y: Enriching existing requests and reusing the ports table felt the easiest; 

https://github.com/kurtosis-tech/kurtosis-cloud-backend/pull/154 implementation here

cloud/api/protobuf/kurtosis_backend_server_api.proto
```diff
@@ -28,7 +28,8 @@ service KurtosisCloudBackendServer {
   };
   rpc UnlockPort(UnlockPortRequest) returns(google.protobuf.Empty){};
   rpc LockPort(LockPortRequest) returns(google.protobuf.Empty){};
-  rpc GetUnlockedPorts(GetUnlockedPortsRequest) returns(GetUnlockedPortsResponse){}
```
h4ck3rk3y, line 31: i realize that simply removing this will break for people who might be on the current version of the engine

h4ck3rk3y: I will keep the previous endpoint;

cloud/api/protobuf/kurtosis_backend_server_api.proto
```diff
@@ -28,7 +28,9 @@ service KurtosisCloudBackendServer {
...
+  rpc GetUnlockedPorts(GetUnlockedPortsRequest) returns(GetUnlockedPortsResponse){};
+  rpc GetCloudPorts(GetCloudPortsRequest) returns(GetCloudPortsResponse){};
```
laurentluce: Can we drop `Cloud`, the methods here are cloud related.

laurentluce (approved):

PR #2406, opened 2024-04-26
feat: set service image in em ui

tedim52:
## Description
Adds UI button for setting service image using `set_service` instruction.

Screen Shot 2024-04-26 at 2 01 06 PM

https://www.loom.com/share/c6e87488214047c782bbffb1d3fa0bb5

## Is this change user facing?
YES

## References
https://github.com/kurtosis-tech/kurtosis/pull/2372

PR #2419, opened 2024-05-01
fix: update image UI arg parsing

tedim52:
## Description
Fixes how args are parsed and passing into script that updates images. Also fixes logic with how `initialSerializedParams` are set.

NO

## References
https://github.com/kurtosis-tech/kurtosis/pull/2406


enclave-manager/web/packages/app/src/emui/enclaves/components/widgets/ImageButton.tsx
```diff
@@ -109,17 +102,36 @@ export const SetImageModel = ({ isOpen, onClose, currentImage, serviceName, encl
...
+    }
 
-    let argsRecord;
-    if (starlarkRun.value.initialSerializedParams) {
-      console.log(`initial serialized params: ${starlarkRun.value.initialSerializedParams}`);
-      argsRecord = deserializeParams(starlarkRun.value.initialSerializedParams);
+    let args;
+    console.log(packageArgs);
+    if (packageArgs.length === 2 && packageArgs[1].name === "args") {
+      args = objectToStarlark(wrapWithArgs(argsRecord), 4);
```
tedim52, line 129: if the package is not one that uses `kwargs` and instead still uses the old method of passing in an `args` object, we have to wrap the `serializedParams` in an `args` object. i do this by retrieving info about a packages `args` from the packageIndexer and checking if a package only has two params and one of them is the `args` object, in which case we wrap in an `args`

h4ck3rk3y (approved):

PR #2422, opened 2024-05-02
fix: set service idempotency logic

tedim52:
## Description
Fixes bugs that occurred when running `set_service` consecutive times against the same enclave.

NO


core/server/api_container/server/startosis_engine/interpretation_time_value_store/interpretation_time_value_store.go
```diff
@@ -83,5 +83,6 @@ func (itvs *InterpretationTimeValueStore) GetNewServiceConfig(name service.Servi
 	if !ok {
 		return nil, stacktrace.NewError("Did not find new service config for '%v' in interpretation time value store.", name)
 	}
+	delete(itvs.setServiceConfigValues, name)
```
tedim52, line 86: after a new service config has been retrieved, don't use it again

h4ck3rk3y (approved):

PR #2426, opened 2024-05-02
feat: retrieve multiple service logs stream at once via cli

tedim52:
## Description
https://www.loom.com/share/aa413be508564af9831a67420d0fb934

YES

### DIFF
--- a/cli/cli/commands/service/logs/logs.go
+++ b/cli/cli/commands/service/logs/logs.go
@@ -8,6 +8,7 @@ package logs
 import (
 	"context"
 	"fmt"
+	"github.com/fatih/color"
 	"github.com/kurtosis-tech/kurtosis/api/golang/core/lib/services"
 	"github.com/kurtosis-tech/kurtosis/api/golang/engine/kurtosis_engine_rpc_api_bindings"
 	"github.com/kurtosis-tech/kurtosis/api/golang/engine/lib/kurtosis_context"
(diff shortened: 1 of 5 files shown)


cli/cli/commands/service/logs/logs.go
```diff
@@ -34,10 +35,12 @@ const (
...
 
 	serviceIdentifierArgKey        = "service"
-	isServiceIdentifierArgOptional = false
-	isServiceIdentifierArgGreedy   = false
+	isServiceIdentifierArgOptional = true // don't need to pass this in if they use the return all services flag
+	isServiceIdentifierArgGreedy   = true
 
 	shouldFollowLogsFlagKey  = "follow"
+	returnAllServiceLogs     = "all-services"
+	allServicesWildcard      = "*"
```
tedim52, line 43: a user requested a wildcard experience like `kurtosis service logs enclave *` where if they passed in the wildcard all service logs would be returned but you hvae to pass it in like `kurtosis service logs enclave '*'` or else my shell tries to interpret the wildcard , lmk if you have better suggestions

h4ck3rk3y: hmm, not great but nothing better immediately comes to mind

h4ck3rk3y: maybe a `--all` flag that autocompletes?

h4ck3rk3y (approved):
beautiful

cli/cli/commands/service/logs/logs.go
tedim52: `-a` , `-all` is already used for "get all logs of the service" right now, so we'll just rock with the wildcard thing for now

PR #2427, opened 2024-05-06
feat: `IsNewKurtosisVersionAvailable` and `UpgradeKurtosisVersion` GRPC endpoints added

leoporoli:
## Description
Adding this new endpoint because it will needed in the cloud-backend to implement the endpoints

NO

## References (if applicable)
This is part of the `upgrade Kurtosis from UI` project


cloud/api/protobuf/kurtosis_backend_server_api.proto
```diff
@@ -281,4 +283,17 @@ message  AddAliasRequest {
   string access_token=1;
   Port port=2;
   string alias=3;
-}
\ No newline at end of file
+}
+
+message IsNewKurtosisVersionAvailableRequest {
+  string current_kurtosis_version = 2;
```
laurentluce, line 290: Those requests go through the cloud gateway afaik so it should contain an access token.

cloud/api/protobuf/kurtosis_backend_server_api.proto
```diff
@@ -281,4 +283,17 @@ message  AddAliasRequest {
...
+}
+
+message IsNewKurtosisVersionAvailableResponse {
+  bool isAvailable = 1;
+  string latestVersion = 2;
+}
+
+message UpgradeKurtosisVersionRequest {
+  string api_key = 1;
```
laurentluce: If this is coming from the EM UI then it should take an access token (jwt).

cloud/api/protobuf/kurtosis_backend_server_api.proto
leoporoli: @laurentluce access token added

cloud/api/protobuf/kurtosis_backend_server_api.proto
leoporoli: @laurentluce access token added

laurentluce (approved):

PR #2441, opened 2024-05-08
fix: restart APICs containers when the engine is restarted

leoporoli:
## Description
Fixing the issue when running the `kurtosis engine restart --restart-apic-containers`. The issue was happening when the Engine's REST API was trying to reconnect to all the enclaves while the APIC containers were in the restart process.

We are also increasing the engine's readiness check time because the restart process could take more than 1 minutes if there are many APIC containers running

YES

## References (if applicable)
This is part of the `upgrade kurtosis from the UI` project

### DIFF
--- a/container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
+++ b/container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
@@ -29,8 +29,8 @@ const (
 	enclaveManagerUIPort                        = 9711
 	enclaveManagerAPIPort                       = 8081
 	engineDebugServerPort                       = 50102 // in ClI this is 50101 and 50103 for the APIC
-	maxWaitForEngineAvailabilityRetries         = 10
-	timeBetweenWaitForEngineAvailabilityRetries = 1 * time.Second
+	maxWaitForEngineAvailabilityRetries         = 40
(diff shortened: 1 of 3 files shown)


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -29,8 +29,8 @@ const (
 	enclaveManagerUIPort                        = 9711
 	enclaveManagerAPIPort                       = 8081
 	engineDebugServerPort                       = 50102 // in ClI this is 50101 and 50103 for the APIC
-	maxWaitForEngineAvailabilityRetries         = 10
-	timeBetweenWaitForEngineAvailabilityRetries = 1 * time.Second
+	maxWaitForEngineAvailabilityRetries         = 40
```
leoporoli, line 32: It's increasing the timeout because now the engine could take more time to restart if it has to restart several APIC containers.
I've tested it with 20 APIC containers running at the same time

engine/server/engine/main.go
```diff
@@ -261,6 +261,12 @@ func runMain() error {
 		}
 	}()
 
+	if serverArgs.RestartAPIContainers {
```
leoporoli, line 264: moving the logic up, before starting the REST API because it will try to reconnect to all the APIC containers and we have to restart them before.

tedim52 (approved):

PR #2445, opened 2024-05-10
feat: update delete instances request

tedim52:
## Description
Extend functionality of DeleteInstance request to delete multiple instances or instances with a certain status. Also, adds auth to it by requiring the system user api key to be provided.

NO


cloud/api/protobuf/kurtosis_backend_server_api.proto
```diff
@@ -156,13 +156,18 @@ message UpdateAddressArgs {
   string country = 8;
 }
 
-message DeleteInstanceRequest {
-  string instance_id = 1;
-  string ec2_id = 2;
```
tedim52, line 161: ec2 id was actually never used in the request

cloud/api/protobuf/kurtosis_backend_server_api.proto
```diff
@@ -156,13 +156,18 @@ message UpdateAddressArgs {
...
+message DeleteInstancesRequest {
+  repeated string instance_ids = 1;
+
+  // must be system user to delete instances
+  string system_user_api_key = 2;
```
laurentluce: I am wondering if we should rename this to `api_key` for consistency with the other methods and also when we allow the user to delete its own instance.  We can do authz as part of the method implementation.

laurentluce (approved):

cloud/api/protobuf/kurtosis_backend_server_api.proto
tedim52: good idea

PR #2474, opened 2024-06-05
feat: port name validation RFC-6335

leovct:
## Description

Kubernetes adheres to [RFC-6335](https://www.rfc-archive.org/getrfc.php?rfc=6335#gsc.tab=0), which specifies that port names must be between 1 and 15 characters. However, this constraint is currently not enforced by the Kurtosis engine, leading to deployment failures in Kubernetes environments when port names exceed 15 characters.

Here is an example that the Kurtosis validation step failed to catch:

image (1)

This PR resolves the issue by adding an additional validation step for port names. It also refactors the regex definitions in `service.go` and improves the test design for validating service and port names in `service_test.go`.

YES

## References (if applicable)

- https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L326-L351
- https://github.com/0xPolygon/kurtosis-cdk/pull/146#issue-2332946061


container-engine-lib/lib/backend_interface/objects/service/service_test.go
```diff
@@ -1,34 +1,65 @@
...
 
-func TestServiceNameWithSpecialChars(t *testing.T) {
-	require.True(t, IsServiceNameValid(ServiceName("a-b")))
-	require.False(t, IsServiceNameValid(ServiceName("-bc")))
-	require.False(t, IsServiceNameValid(ServiceName("a--")))
-	require.False(t, IsServiceNameValid(ServiceName("a_b")))
-	require.False(t, IsServiceNameValid(ServiceName("a%b")))
-	require.False(t, IsServiceNameValid(ServiceName("a:b")))
-	require.False(t, IsServiceNameValid(ServiceName("a/b")))
+	for _, name := range validServiceNames {
```
tedim52, line 29: nice refactor

tedim52 (approved):

PR #2485, opened 2024-06-11
feat: Self cloud hosting

laurentluce:
## Description
The Kurtosis CLI supports the local use case (localhost) and the Kurtosis cloud use case (cloud.kurtosis.com).  We need to be able to build the CLI once and deploy it to multiple environments with different domain names.  Add option to the engine to set a domain name used by the EM UI.  The domain name is set in a static js file read by the EM UI.  More details on this approach discussed here: https://kurtosistech.slack.com/archives/C032ZB39AHH/p1717770540781579

Adding Nginx in front of the engine + install script will be done as a cloud-config PR.

YES


enclave-manager/web/packages/app/src/client/enclaveManager/LocalKurtosisClient.ts
```diff
@@ -7,11 +7,12 @@ import { KurtosisClient } from "./KurtosisClient";
 export class LocalKurtosisClient extends KurtosisClient {
   constructor() {
     const defaultUrl = new URL(`${window.location.protocol}//${window.location.host}`);
+    var baseUrl = KURTOSIS_EM_API_DEFAULT_URL;
```
laurentluce, line 10: If domain name set by the user, use that instead.

enclave-manager/web/packages/app/src/emui/enclaves/components/tables/PortsTable.tsx
```diff
@@ -47,6 +47,9 @@ export const getPortTableRows = (
           `${port.number}-${shortUUID(serviceUUID)}-${shortUUID(enclaveUUID)}-${shortUUID(instanceUUID)}` +
           `.${KURTOSIS_CLOUD_HOST}`;
       }
+    } else if (window.env !== undefined && window.env.domain !== undefined) {
```
laurentluce, line 50: Public ports using the user domain name.

leoporoli (approved):

PR #2499, opened 2024-06-21
fix: logs for tasks

tedim52:
## Description
This PR manipulates `run_sh` and `run_python` logic to stream logs from the command or python script thats being executed on the container. This also allows retrieving task logs after a run has finished, whereas before, task containers would not return logs from `kurtosis service logs`.

An improvement to this would be to enable streaming back logs via run output so users don't have to go through Docker Desktop or docker cli to view logs from a long running task that might have err'd. 

https://www.loom.com/share/89181c37442048b9bdc746bfd37fe2ec

YES


core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_sh.go
```diff
@@ -276,7 +276,7 @@ func (builtin *RunShCapabilities) Execute(ctx context.Context, _ *builtin_argume
 	if err != nil {
 		return "", stacktrace.Propagate(err, "error occurred while preparing the sh command to execute on the image")
 	}
-	fullCommandToRun := []string{shellWrapperCommand, "-c", commandToRun}
+	fullCommandToRun := []string{shellWrapperCommand, "-c", fmt.Sprintf("%v \"$(%v)\" %v %v", "printf \"%s\\n\"", commandToRun, ">>", "/proc/1/fd/1")}
```
tedim52: worst case if this err's can do a retry with `[]string{shellWrapperCommand, "-c", commandToRun}`?

h4ck3rk3y (approved):
can we run the ethereum package against this image?

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_sh.go
h4ck3rk3y: https://github.com/kurtosis-tech/avalanche-package/blob/main/src/node_launcher.star#L47-L122 how about this approach

we wrap up the exec and make it write to some file in a logs dir and tail the logs there; but piping to fd1 makes sense too

tedim52: sounds good will try this

tedim52:
The video I put in the description runs against the ethereum package.

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_sh.go
tedim52: implemented this and it works for streaming logs!

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
```diff
@@ -50,14 +50,24 @@ const (
...
-var runTailCommandToPreventContainerToStopOnCreating = []string{"tail", "-f", "/dev/null"}
+var runCommandToStreamTaskLogs = []string{shellWrapperCommand, "-c", fmt.Sprintf("touch %s && tail -F %s", taskLogFilePath, taskLogFilePath)}
+
+// Wraps [commandToRun] to enable streaming logs from tasks.
+// Uses curly braces to execute the command(s) in the current shell.
+// Adds an extra echo to ensure each log ends with a newline.
+// Uses tee to direct output to the task log file while maintaining output to stdout.
+// Redirects stderr to stdout.
+func getCommandToRunForStreamingLogs(commandToRun string) []string {
+	return []string{shellWrapperCommand, "-c", fmt.Sprintf("{ %v; } %v %v %v %v %v %v %v %v", commandToRun, "2>&1", "|", "tee", taskLogFilePath, "&&", "echo", ">>", taskLogFilePath)}
```
tedim52, line 69: The extra echo adds an empty log line, but it's necessary because some tasks like read-prague-time in ethereum-package output without a newline, causing tail in the entrypoint to miss them. tail needs a newline to detect the log it seems. Let me know if there's a workaround; otherwise, we'll keep it until there's a complaint about the extra empty log line in tasks.

PR #2525, opened 2024-08-09
feat: make kurtosis service logs faster

tedim52:
## Description
Users were experiencing `kurtosis service logs` taking a long time. After running tests, I discovered that a majority of execution time during log processing was spent in the following lines:
```
logLines := []logline.LogLine{*logLine}
userServicesLogLinesMap := map[service.ServiceUUID][]logline.LogLine{
      serviceUuid: logLines,
}
logsByKurtosisUserServiceUuidChan <- userServicesLogLinesMap
```
Prior to this change, we were sending logs one at a time on an unbuffered channel - unbuffered channels block until the receiving goroutine reads the value. This was causing a lot of time being wasted waiting to send log lines across the channel.

This change implements a `LogLineSender` that:
1. uses a buffered go channel (won't block on sending line unless buffer is full)
2. batches log lines (reduces read overhead, receiving goroutine performs fewer reads/sends)

With this change, the time to read 20 minutes of `cl-lighthouse-geth` logs with log level set to debug went from `1min53sec` to `30.055` seconds. The time to read 2 hours 10 minutes worth of `cl-lighthouse` debug logs (around 3.4 gb of logs) went from `15min1sec` to `3min31` sec. (As a benchmark, `cat logs.json` on `3.4 gb` of logs takes around `2min` - on my machine -  so much closer) This can likely be improved further by tuning the buffer size and batch amount.

YES

## References:
https://discord.com/channels/783719264308953108/1267837033032974467/1267842228072611881

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,7 +145,7 @@ version: 2.1
 
 orbs:
   npm-publisher: uraway/npm-publisher@0.2.0
-  kurtosis-docs-checker: kurtosis-tech/docs-checker@0.2.7
+  kurtosis-docs-checker: kurtosis-tech/docs-checker@0.2.8
   slack: circleci/slack@4.10.1
 
 executors:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1474,17 +1474,18 @@ workflows:
           all_architectures: true
           name: "Check if CLI builds for all os and arch pairs"
           <<: *filters_ignore_main
-
-      - test_enclave_manager_web_ui:
-          name: "Test Basic Web UI Functionality in Docker"
-          context:
-            - docker-user
-          requires:
-            - build_cli
-            - build_api_container_server
-            - build_engine_server
-            - build_files_artifacts_expander
-          <<: *filters_ignore_main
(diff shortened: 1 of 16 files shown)


engine/server/engine/centralized_logs/logline/logline_sender.go
```diff
@@ -0,0 +1,62 @@
+package logline
+
+import (
+	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/service"
+	"sync"
+)
+
+const (
+	batchLogsAmount    = 500
+	logsChanBufferSize = 300
```
tedim52, line 10: tested a few values here, but can tune later

.circleci/config.yml
```diff
@@ -1474,17 +1474,18 @@ workflows:
...
-          name: "Test Basic Web UI Functionality in Docker"
-          context:
-            - docker-user
-          requires:
-            - build_cli
-            - build_api_container_server
-            - build_engine_server
-            - build_files_artifacts_expander
-          <<: *filters_ignore_main
+#      08/10/2024, tedi: turning these CI tests off for now as they depend on the package indexer for catalog.kurtosis.com which was shut down with
```
h4ck3rk3y: delete it instead?

h4ck3rk3y (approved):

Issue #2443, opened 2024-05-09
make log rotation more granular to a day


h4ck3rk3y:
### Background & motivation

7 days is too long; need. to lower to a day

### Desired behaviour

kurtosis engine restart --retention 1d/1w

### How important is this to you?

Painful; the lack of this feature makes using Kurtosis frictionful.

### What area of the product does this pertain to?

CLI: the Command Line Interface

h4ck3rk3y:
Closing this as @tedim52 seems to have fixed this

Issue #2190, opened 2024-02-20
The log saved during kurtosis log collection is too large


icesfeathers:
### Background & motivation

I'm running a geth private network using kurtosis, and the kurtosis logs used 150G of the hard drive in one day, but the actual geth network only used 20G.
I checked the container kurtosis-logs-collection and found that it was continuously collecting debug logs, causing my hard disk space to be filled up with logs.
![20240220-133736](https://github.com/kurtosis-tech/kurtosis/assets/42756626/fb2533ff-132a-40c6-84d3-3c8bfc590eb4)

### Desired behaviour

Currently, I stopped kurtosis's log collection container and cleared kurtosis-logs-storage to free up space.
For me, kurtosis logs are not more important than eth network. 
It is recommended to increase the log collection level of kurtosis logs or disable it.

### How important is this to you?

Critical; Kurtosis is unusable for me without it.

### What area of the product does this pertain to?

CLI: the Command Line Interface

icesfeathers:
Here is the command information and configuration information for running kurtosis
```shell
kurtosis run --cli-log-level "panic" --enclave my-testnet github.com/kurtosis-tech/ethereum-package@1.4.0 "$(cat /data/fleet/kurtosis_network_params.yaml)"
```
```yarm
participants:
    - validator_count: 512
      beacon_extra_params:
          - "--reconstruct-historic-states"
      el_extra_params:
          - "--gcmode=archive"
      el_max_mem: 8192
      bn_max_mem: 4096
      v_max_mem: 4096
    - validator_count: 1
      el_max_mem: 8192
      bn_max_mem: 4096
      v_max_mem: 4096
network_params:
    seconds_per_slot: 6
    eth1_follow_distance: 2
    deneb_fork_epoch: 9999999999999
additional_services:
    - tx_spammer
    - blob_spammer
    - el_forkmon
    - blockscout
    - beacon_metrics_gazer
    - dora
    - prometheus_grafana
mev_type: "full"
```

tedim52:
Thanks for filing this @icesfeathers ! As mentioned in Discord, it seems like what would make sense as a solution here is:

the ability the ability to turn off the collection of logs (in which case you won't be able to view logs after services have died)

OR

a byte based retention mechanism where historical logs are removed after a certain threshold is passed so it logs fill up disk space.(For context, currently, we have time based log retention where old logs are removed after 4 weeks but no byte based log retention)

icesfeathers:
[...]

PR #2534, opened 2024-08-14
feat: [log retention improvements pt. 1] introduce file layout interface

tedim52:
## Description
Log Retention Improvements:

Users have had issues with logs from long running enclaves taking up tons of storage. We have a retention mechanism that automatically rotates logs after some time but currently it a) is only able to rotate logs weekly b) can not be configured.

These improvements will allow retention to be as granular as hourly and will allow users to configure the retention period (eg. `1hr`, `2hr`, `1day`, `1week`). To do this, a few changes need to happen. Most notably the way logs are stored and retrieved needs to change to support rotating log files hourly. Implementing this requires changes across a few components(`LogsAggregator`, `LogsDatabaseClient`,  `LogFileManager`, cli, so I'll be making them incrementally.

- [x] PR 1: Introduce `LogFileLayout` and `PerWeekLogFileLayout` and adjust `LogFileManager` and `LogsDatabaseClient` to use this for retrieving log file paths
- [ ] PR 2: Make retention configurable via a CLI flag
- [ ] PR 3: Implement `PerHourFileLayout` and swap storage format from `PerWeekFileLayout` to `PerHourFileLayout`
- [ ] PR 4: Make `LogsAggregator` use `LogFileLayout` for determining storage format and set it to use `PerHourFileLayout`
------------------------------
This first PR sets the stage for this change by introducing a new interface called `LogFileLayout` and migrating the existing storage format to use it. Right now, knowledge about the log storage format is spread across several entities (`LogsAggregator` - to store, `LogFileManager` - to rotate logs, `StreamLogsStrategy` - to read logs). This interface consolidates knowledge of how logs are stored into one module that can be used by any entity doing operations that require retrieving log files from the filesystem. This will make the transition to a different storage format safer (only requires implementing + testing `PerHourFileLayout` module) and makes it easier to swap out the storage format in the future (eg. if even more granular retention is need) without breaking other entities.

NO

## References 
https://github.com/kurtosis-tech/kurtosis/issues/2443
https://github.com/kurtosis-tech/kurtosis/issues/2190


engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/file_layout.go
```diff
@@ -0,0 +1,26 @@
...
+
+type LogFileLayout interface {
+	// GetLogFileLayoutFormat returns a string representation the "format" that files are laid out in
+	// Formats are composed:
+	// - "/" - representing a nested directory
+	// - "<enclaveUuid>" - representing where an enclave uuid is inserted
+	// - "<serviceUuid>" - representing where a service uuid is inserted
+	// - time formats specified by strftime https://cplusplus.com/reference/ctime/strftime/
+	// - any other ascii text
+	GetLogFileLayoutFormat() string
```
tedim52, line 16: this is the format that Vector LogsAggregator expect, probably a better way to represent this, open to idea

h4ck3rk3y (approved):

PR #2543, opened 2024-08-30
feat!: impl dependency list flag 

tedim52:
## Description
`kurtosis run github.com/kurtosis-tech/... --dependencies` returns a `yaml` with a list of images and packages the run will need to depend on.

Adding `--pull` will pull all the dependencies locally. The `kurtosis.yml` will be updated to use local versions of packages and Docker images are pulled onto the machine locally

YES


api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -531,6 +531,41 @@ func (enclaveCtx *EnclaveContext) GetStarlarkRun(ctx context.Context) (*kurtosis
 	return response, nil
 }
 
+func (enclaveCtx *EnclaveContext) GetStarlarkPackagePlanYaml(ctx context.Context, packageId string, serializedParams string, dependenciesOnly bool) (*kurtosis_core_rpc_api_bindings.PlanYaml, error) {
+	serializedParams, err := maybeParseYaml(serializedParams)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred when parsing YAML args for package '%v'", serializedParams)
```
mieubrisse: should this be s/serializedParams/packageId/?

api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -531,6 +531,41 @@ func (enclaveCtx *EnclaveContext) GetStarlarkRun(ctx context.Context) (*kurtosis
...
+	}
+	response, err := enclaveCtx.client.GetStarlarkPackagePlanYaml(ctx, &kurtosis_core_rpc_api_bindings.StarlarkPackagePlanYamlArgs{
+		PackageId:              packageId,
+		SerializedParams:       &serializedParams,
+		RelativePathToMainFile: nil,
+		MainFunctionName:       nil,
+		DependenciesOnly:       &dependenciesOnly,
+	})
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred while getting the starlark package plan yaml run.")
```
mieubrisse: Is the "....run" at the end of the sentence perhaps a copy-paste that wasn't deleted? (my understanding is this returns the actual YAML; feel free to ignore this comment if not useful)

api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -531,6 +531,41 @@ func (enclaveCtx *EnclaveContext) GetStarlarkRun(ctx context.Context) (*kurtosis
...
+		return nil, stacktrace.Propagate(err, "An error occurred when parsing YAML args for package '%v'", serializedParams)
+	}
+	response, err := enclaveCtx.client.GetStarlarkScriptPlanYaml(ctx, &kurtosis_core_rpc_api_bindings.StarlarkScriptPlanYamlArgs{
+		SerializedScript: serializedScript,
+		SerializedParams: &serializedParams,
+		MainFunctionName: nil,
+		DependenciesOnly: &dependenciesOnly,
+	})
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred while getting the last starlark script plan yaml run.")
```
mieubrisse: Same question

cli/cli/commands/run/run.go
```diff
@@ -55,6 +60,11 @@ const (
 	dryRunFlagKey = "dry-run"
 	defaultDryRun = "false"
 
+	dependenciesFlagKey     = "dependencies"
+	defaultDependencies     = "false"
```
mieubrisse: suggestion to make this `dependenciesFlagDefault`, both to pattern match other varnames and not suggest that this variable is in fact containing a list of default dependencies

cli/cli/commands/run/run.go
```diff
@@ -55,6 +60,11 @@ const (
...
+	pullDependenciesFlagKey = "pull"
+	defaultPullDependencies = "false"
```
mieubrisse: same idea here

cli/cli/commands/run/run.go
```diff
@@ -128,6 +139,18 @@ var StarlarkRunCmd = &engine_consuming_kurtosis_command.EngineConsumingKurtosisC
 			Type:    flags.FlagType_Bool,
 			Default: defaultDryRun,
 		},
+		{
+			Key:     dependenciesFlagKey,
+			Usage:   "If true, a yaml will be returned with a list of images and packages that this run depends on.",
```
mieubrisse: Reading this, I had the question "how?" Might be worth clarifying that it's printed to STDOUT

cli/cli/commands/run/run.go
```diff
@@ -128,6 +139,18 @@ var StarlarkRunCmd = &engine_consuming_kurtosis_command.EngineConsumingKurtosisC
...
+			Type:    flags.FlagType_Bool,
+			Default: defaultDependencies,
+		},
+		{
+			Key:     pullDependenciesFlagKey,
+			Usage:   "If true, and the dependencies flag is passed, attempts to pull all images and packages that the run depends on locally. kurtosis.yml is updated with the replace directives pointing to locally pulled packages.",
```
mieubrisse: suggestion: to make this code more refactorable, you can:
1. swap the hardcoded `and the dependencies flag is passed` with `depenciesFlagKey`
2. same thing for hardcoded `kurtosis.yml` text

mieubrisse: Extra: might be worth adding some color to "kurtosis.yml is updated with the replace directives pointing to locally pulled packages." - as someone out of Kurtosis for a while, I was confused for a second what would actually happen.

Extra question: what happens if I run this against a script, without a kurtosis.yml?

cli/cli/commands/run/run.go
```diff
@@ -366,6 +399,44 @@ func run(
 		defer output_printers.PrintEnclaveName(enclaveCtx.GetEnclaveName())
 	}
 
+	isRemotePackage := strings.HasPrefix(starlarkScriptOrPackagePath, githubDomainPrefix)
+
+	if dependencies {
```
mieubrisse: minor suggestion: rename to `isDependenciesOnly` (I did a double-take on `if dependencies`)

cli/cli/commands/run/run.go
```diff
@@ -366,6 +399,44 @@ func run(
...
+		dependencyYaml, err := getPackageDependencyYaml(ctx, enclaveCtx, starlarkScriptOrPackagePath, isRemotePackage, packageArgs)
+		if err != nil {
+			return stacktrace.Propagate(err, "An error occurred getting package dependencies")
+		}
+		fmt.Printf("%v\n", dependencyYaml.PlanYaml)
```
mieubrisse: Not exactly sure what `dependencyYaml.PlanYaml` is; maybe worth passing through a  YAML pretty-printer if it's not already?

cli/cli/commands/run/run.go
```diff
@@ -366,6 +399,44 @@ func run(
...
+
+		if pullDependencies {
+			type PackageDependencies struct {
+				Images   []string `yaml:"images"`
+				Packages []string `yaml:"packageDependencies"`
+			}
+			var pkgDeps PackageDependencies
+			err := yaml.Unmarshal([]byte(dependencyYaml.PlanYaml), &pkgDeps)
+			if err != nil {
+				return stacktrace.Propagate(err, "An error occurred unmarhsaling dependency yaml string")
```
mieubrisse: typo in "unmarshaling"

cli/cli/commands/run/run.go
```diff
@@ -366,6 +399,44 @@ func run(
...
+			}
+
+			err = pullImagesLocally(ctx, pkgDeps.Images)
+			if err != nil {
+				return err
```
mieubrisse, line 435: should this be a `stacktrace.Propagate`?

cli/cli/commands/run/run.go
```diff
@@ -366,6 +399,44 @@ func run(
...
+			}
+
+			packageNamesToLocalFilepaths, err := pullPackagesLocally(pkgDeps.Packages)
+			if err != nil {
+				return err
```
mieubrisse, line 440: should this be stacktrace.Propagate?

cli/cli/commands/run/run.go
```diff
@@ -366,6 +399,44 @@ func run(
...
+			}
+
+			err = updateKurtosisYamlWithReplaceDirectives(packageNamesToLocalFilepaths)
+			if err != nil {
+				return err
```
mieubrisse, line 445: should this be stacktrace.Propagate?

cli/cli/commands/run/run.go
```diff
@@ -588,6 +658,26 @@ func getOrCreateEnclaveContext(
...
+	ctx context.Context,
+	enclaveCtx *enclaves.EnclaveContext,
+	starlarkScriptOrPackageId string,
+	isRemote bool,
+	packageArgs string,
+) (*kurtosis_core_rpc_api_bindings.PlanYaml, error) {
+	var packageYaml *kurtosis_core_rpc_api_bindings.PlanYaml
+	var err error
+	if isRemote {
+		packageYaml, err = enclaveCtx.GetStarlarkPackagePlanYaml(ctx, starlarkScriptOrPackageId, packageArgs, true)
```
mieubrisse: What's the magic `true` constant here?

cli/cli/commands/run/run.go
```diff
@@ -588,6 +658,26 @@ func getOrCreateEnclaveContext(
...
+	} else {
+		packageYaml, err = enclaveCtx.GetStarlarkScriptPlanYaml(ctx, starlarkScriptOrPackageId, packageArgs, true)
```
mieubrisse: Same question

cli/cli/commands/run/run.go
```diff
@@ -734,3 +824,92 @@ func isHttpUrl(maybeHttpUrl string) bool {
 	httpProtocolRegex := regexp.MustCompile(httpProtocolRegexStr)
 	return httpProtocolRegex.MatchString(maybeHttpUrl)
 }
+
+func pullImagesLocally(ctx context.Context, images []string) error {
+	kurtosisBackend, err := backend_creator.GetDockerKurtosisBackend(backend_creator.NoAPIContainerModeArgs, configs.NoRemoteBackendConfig)
```
mieubrisse, line 867: Seems like this only works for Docker; worth updating the flag with that information?

cli/cli/commands/run/run.go
```diff
@@ -734,3 +824,92 @@ func isHttpUrl(maybeHttpUrl string) bool {
...
+			return stacktrace.Propagate(err, "An error occurred pulling '%v' locally.", img)
+		}
+	}
+	return nil
+}
+
+func pullPackagesLocally(packageDependencies []string) (map[string]string, error) {
+	localPackagesToRelativeFilepaths := map[string]string{}
+
+	wd, err := os.Getwd()
```
mieubrisse: just personally I generally prefer more explicit varnames for reading clarity (e.g. `workingDirectory` rather than `wd` but really up to you)

cli/cli/commands/run/run.go
```diff
@@ -734,3 +824,92 @@ func isHttpUrl(maybeHttpUrl string) bool {
...
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred getting rel path between '%v' and '%v'.", wd, relParentCwd)
+	}
+
+	for _, dependency := range packageDependencies {
+		packageIdParts := strings.Split(dependency, "/")
+		packageName := packageIdParts[len(packageIdParts)-1]
+		logrus.Infof("Pulling package: %v", dependency)
+
+		// assume dependency doesn't have http:// prefix or .git suffix
```
mieubrisse: just to double-check, is it worth verifying this in code or does it not matter?

cli/cli/commands/run/run.go
```diff
@@ -734,3 +824,92 @@ func isHttpUrl(maybeHttpUrl string) bool {
...
+		repoUrl := fmt.Sprintf("%s%s%s", "http://", dependency, ".git")
+		localPackagePath := fmt.Sprintf("%s/%s", parentCwd, packageName)
+		logrus.Infof("repo url: %v", repoUrl)
```
mieubrisse: Potentially better as DEBUG?

cli/cli/commands/run/run.go
```diff
@@ -734,3 +824,92 @@ func isHttpUrl(maybeHttpUrl string) bool {
...
+		_, err := git.PlainClone(localPackagePath, false, &git.CloneOptions{
```
mieubrisse: Maybe worth using a named constant for `false` magic var?

cli/cli/commands/run/run.go
```diff
@@ -734,3 +824,92 @@ func isHttpUrl(maybeHttpUrl string) bool {
...
+	}
+
+	for _, dependency := range packageDependencies {
+		packageIdParts := strings.Split(dependency, "/")
+		packageName := packageIdParts[len(packageIdParts)-1]
+		logrus.Infof("Pulling package: %v", dependency)
+
+		// assume dependency doesn't have http:// prefix or .git suffix
+		repoUrl := fmt.Sprintf("%s%s%s", "http://", dependency, ".git")
+		localPackagePath := fmt.Sprintf("%s/%s", parentCwd, packageName)
```
mieubrisse, line 919: Wouldn't we need to find the `kurtosis.yml` in the tree above, both to:
1. ensure we have one to update with `replace` directives
2. ensure we're cloning the dependencies to the right spot?

cli/cli/commands/run/run.go
```diff
@@ -734,3 +824,92 @@ func isHttpUrl(maybeHttpUrl string) bool {
...
+		localPackagesToRelativeFilepaths[dependency] = fmt.Sprintf("%s/%s", relParentCwd, packageName)
+	}
+
+	return localPackagesToRelativeFilepaths, nil
+}
+
+func updateKurtosisYamlWithReplaceDirectives(packageNamesToLocalFilepaths map[string]string) error {
+	file, err := os.OpenFile(kurtosisYMLFilePath, os.O_WRONLY|os.O_APPEND, kurtosisYMLFilePerms)
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred opening kurtosis.yml file.")
```
mieubrisse: hardcoded `kurtosis.yml` can be replaced with the variable to make this more refactor-resistant

cli/cli/commands/run/run.go
```diff
@@ -734,3 +824,92 @@ func isHttpUrl(maybeHttpUrl string) bool {
...
+	}
+	defer file.Close()
+
+	_, err = file.Write([]byte("replace:\n"))
```
mieubrisse: 1. would this cause problems if there's already a `replace` directive, either from a user manually creating it or having run this before? thinking as a user, I think my ideal would be that the `replace` section just gets updated. that seems like it'd be a lot of work for us here, so maybe instead we could error if `replace` directive already exists in the kurtosis.yml, and print the block they need to manually add into the preexisting `replace` block? just solutionizing
2. rather than using a hardcoded value here and below, wdyt if we use the constant from wherever it is in the SL engine that actually processes the word `replace`?

cli/cli/commands/run/run.go
```diff
@@ -734,3 +824,92 @@ func isHttpUrl(maybeHttpUrl string) bool {
...
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred writing 'replace:' to kurtosis.yml")
```
mieubrisse: same question about hardcoded `replace`

cli/cli/commands/run/run.go
```diff
@@ -734,3 +824,92 @@ func isHttpUrl(maybeHttpUrl string) bool {
...
+	}
+
+	for packageName, localFilepath := range packageNamesToLocalFilepaths {
+		_, err := file.Write([]byte(fmt.Sprintf("  %s: %s\n", packageName, localFilepath)))
```
mieubrisse: Potential bug: we're assuming the user's YAML is indented by two spaces per indentation level; seems like it could also be 4 spaces, tabs, etc. If it's hard to fix in a proper way, maybe we could at least add a TODO here showing that we recognize the risk and how to address this properly?

core/server/api_container/server/startosis_engine/plan_yaml/plan_yaml.go
```diff
@@ -8,11 +8,13 @@ const (
 	exec   TaskType = "exec"
 )
 
-type privatePlanYaml struct {
-	PackageId      string           `yaml:"packageId,omitempty"`
-	Services       []*Service       `yaml:"services,omitempty"`
-	FilesArtifacts []*FilesArtifact `yaml:"filesArtifacts,omitempty"`
-	Tasks          []*Task          `yaml:"tasks,omitempty"`
+type PlanYaml struct {
```
mieubrisse, line 11: Potentially worth a quick comment on why this struct is a keyed dict; as a new engineer I would have expected it to be a list of instructions. Obv I Kevin know the context (b/c we hacked this together for Ben), but might be worth jotting a quick note here

core/server/api_container/server/startosis_engine/plan_yaml/plan_yaml_generator.go
```diff
@@ -22,38 +22,56 @@ const (
 	outputFutureRefType    = "output"
 )
 
-// PlanYaml is a yaml representation of the effect of an Instructions Plan or sequence of instructions on the state of the Enclave.
-type PlanYaml struct {
-	privatePlanYaml *privatePlanYaml
+// PlanYamlGenerator is a yaml representation of the effect of an Instructions Plan or sequence of instructions on the state of the Enclave.
```
mieubrisse: this might need a slight update given it's now (more accurately, as I understand) named as a generator

core/server/api_container/server/startosis_engine/plan_yaml/plan_yaml_generator.go
```diff
@@ -22,38 +22,56 @@ const (
...
 		futureReferenceIndex: map[string]string{},
 		filesArtifactIndex:   map[string]*FilesArtifact{},
 		latestUuid:           0,
 	}
 }
 
-func (planYaml *PlanYaml) GenerateYaml() (string, error) {
-	yamlBytes, err := yaml.Marshal(planYaml.privatePlanYaml)
+func (planYaml *PlanYamlGenerator) GenerateYaml(dependenciesOnly bool) (string, error) {
+	planYaml.AddImages()
```
mieubrisse: Puzzled by this.. are we adding empty images here? If so, why do we need to do this?

core/server/api_container/server/startosis_engine/plan_yaml/plan_yaml_generator.go
```diff
@@ -22,38 +22,56 @@ const (
...
+
+	var planYamlCopy PlanYaml
+	planYamlCopy = *planYaml.privatePlanYaml
+	if dependenciesOnly {
+		planYamlCopy = PlanYaml{
+			PackageId:           "",
```
mieubrisse: I was confused for a few seconds by why we're nulling out the package ID, as well as a bunch of other stuff. I'm guessing it's because we expect this YAML to be consumed by the CLI, which will process the images & dependencies for the `--pull` flag?

I think my expectation would have been that the plan YAML we return to the CLI is always complete, and either the user or the CLI can consume whatever sub-parts it needs (rather than us suppressing info on the APIC end).

Rationale: it seems like we're then tying the APIC's workings too much to what sub-parts of the output the user might want (e.g. then we'd be opened up to the APIC needing to change if the user wants a `--services-only` flag). 

Would it be feasible to have the PlanYamlGenerator always produce the full YAML, and then the CLI or the user can just pull out the parts they want?

core/server/api_container/server/startosis_engine/plan_yaml/plan_yaml_generator.go
```diff
@@ -375,6 +397,20 @@ func (planYaml *PlanYaml) RemoveService(serviceName string) {
 	}
 }
 
+func (planYaml *PlanYamlGenerator) AddPackageDependencies(packageDependency map[string]bool) {
+	for dependency := range packageDependency {
+		planYaml.privatePlanYaml.PackageDependencies = append(planYaml.privatePlanYaml.PackageDependencies, dependency)
+	}
+	slices.Sort(planYaml.privatePlanYaml.PackageDependencies)
```
mieubrisse: This opens us up to duplicates, and as an engineer I'd expect `AddPackageDepencies` with a map-set argument to be idempotent. Would it be better to store the package dependencies as a map-set in the generator too, and only write it out to list at generation time?

core/server/api_container/server/startosis_engine/plan_yaml/plan_yaml_generator.go
```diff
@@ -375,6 +397,20 @@ func (planYaml *PlanYaml) RemoveService(serviceName string) {
...
+}
+
+func (planYaml *PlanYamlGenerator) AddImages() {
+	for img := range planYaml.imageSet {
+		planYaml.privatePlanYaml.Images = append(planYaml.privatePlanYaml.Images, img)
+	}
+	slices.Sort(planYaml.privatePlanYaml.Images)
```
mieubrisse, line 401: Same comment about storing this in the generator as a list vs a map-set

core/server/api_container/server/startosis_engine/plan_yaml/plan_yaml_generator.go
```diff
@@ -375,6 +397,20 @@ func (planYaml *PlanYaml) RemoveService(serviceName string) {
...
 }
 
+func (planYaml *PlanYamlGenerator) AddPackageDependencies(packageDependency map[string]bool) {
+	for dependency := range packageDependency {
+		planYaml.privatePlanYaml.PackageDependencies = append(planYaml.privatePlanYaml.PackageDependencies, dependency)
+	}
+	slices.Sort(planYaml.privatePlanYaml.PackageDependencies)
+}
+
+func (planYaml *PlanYamlGenerator) AddImages() {
```
mieubrisse, line 397: I'm puzzled about the usage of this no-args `AddImages`. I would have expected images to be a params here, same as `AddPackageDependencies`. If they both have params, my mental model of the PlanYamlGenerator feels complete: "it's a builder for the plan YAML, where I chonk things into it and they get serialized out when I press 'generate'".

With `AddImages` not having params, I'm confused because it seems like there's this hidden `planYaml` private variable where images somehow get populated (but not sure how that is). That would suggest to me that the PlanYamlGenerator is actually not the space where I chonk in the data I want the plan YAML to have, but that I need to initialize it at creation with the data I want it to have (and the data-chonking happens elsewhere), but I don't see a constructor that instantiates a generator from a preexisting `planYaml` object (and that feels like it'd be more complex to me than the model of the PlanYamlGenerator as a builder).

core/server/api_container/server/startosis_engine/startosis_interpreter.go
```diff
@@ -386,6 +391,16 @@ func (interpreter *StartosisInterpreter) buildBindings(
 	return &predeclared, nil
 }
 
+const (
+	moduleIdSeperator     = "/"
```
mieubrisse: Thought: wouldn't this constant already be declared somewhere in the Startosis engine? Should we use that instead, so we don't have two constants doing the same thing?

core/server/api_container/server/startosis_engine/startosis_interpreter.go
```diff
@@ -386,6 +391,16 @@ func (interpreter *StartosisInterpreter) buildBindings(
...
+	numModuleIdSeparators = 4
+	prefixNum             = 3
+)
+
+func getModulePrefix(moduleId string) string {
+	return strings.Join(strings.SplitN(moduleId, moduleIdSeperator, numModuleIdSeparators)[:prefixNum], moduleIdSeperator)
```
mieubrisse: From what I understand of `SplitN`, isn't `numModuleIdSeparators` always just `prefixNum` + 1 (so we don't need to maintain the extra const)?

core/server/api_container/server/startosis_engine/startosis_interpreter_plan_yaml_test.go
```diff
@@ -115,7 +115,7 @@ func (suite *StartosisIntepreterPlanYamlTestSuite) TestAddServiceWithFilesArtifa
 	require.Nil(suite.T(), interpretationError)
```
mieubrisse: Should this file be renamed to `startosis_interpreter_plan_yaml_generator_test` to match the name of PlanYaml -> PlanYamlGenerator?

core/server/api_container/server/startosis_engine/startosis_interpreter_plan_yaml_test.go
```diff
@@ -813,6 +833,9 @@ tasks:
   command:
   - echo {{ kurtosis.4.code }} {{ kurtosis.4.output }}
   image: badouralix/curl-jq
+images:
+- badouralix/curl-jq
+- postgres:latest
 `
 	require.Equal(suite.T(), expectedYaml, planYaml)
 }
```
mieubrisse, line 841: Should there be a test case here where the `dependenciesOnly` flag is `true`? (assuming that, per the comment above, we maintain the `dependenciesOnly` flag on the PlanYamlGenerator and it's not the responsibility of the CLI or user to pull out the info they want)

core/server/api_container/server/startosis_engine/startosis_interpreter.go
```diff
@@ -386,6 +391,16 @@ func (interpreter *StartosisInterpreter) buildBindings(
 	return &predeclared, nil
 }
 
+const (
+	moduleIdSeperator     = "/"
+	numModuleIdSeparators = 4
+	prefixNum             = 3
+)
+
+func getModulePrefix(moduleId string) string {
```
mieubrisse, line 401: Reviewing this, it'd have been helpful to have a comment giving an example of `getModulePrefix` is supposed to return so that I know how to model its output in the above code

mieubrisse (approved):
Wasn't sure what level of detail you wanted on this so just did the full shebang

cli/cli/commands/run/run.go
tedim52: this was for debugging so I removed

cli/cli/commands/run/run.go
tedim52: worth verifying in case plan yaml output ever changes, and its a one liner so y not

cli/cli/commands/run/run.go
tedim52: agree with both

cli/cli/commands/run/run.go
tedim52: the hardcoded constant was for indicating only grabbing the dependencies but this will get removed once I process the yaml on the CLI said instead of in the APIC

cli/cli/commands/run/run.go
tedim52: the hardcoded constant was for indicating only grabbing the dependencies but this will get removed once I process the yaml on the CLI said instead of in the APIC

core/server/api_container/server/startosis_engine/plan_yaml/plan_yaml_generator.go
tedim52: agreed, i think i did the initially bc it was faster than processing the yaml on the CLI said but yes

core/server/api_container/server/startosis_engine/startosis_interpreter_plan_yaml_test.go
tedim52: removed the boolean so shouldn't need to test that case

core/server/api_container/server/startosis_engine/plan_yaml/plan_yaml_generator.go
tedim52: removed `AddImages`. Its not needed as `addImage` handles accumulating all images into the image map set and which gets converted to a list when the plan yaml is generated

cli/cli/commands/run/run.go
tedim52: switched to using CLI output printer after CLI parses out the info it wants from the returned plan yaml

core/server/api_container/server/startosis_engine/startosis_interpreter.go
tedim52: ah good point, ye its declared in git package content provider

cli/cli/commands/run/run.go
tedim52: using `GetStarlarkScriptPlanYaml` , it will return dependences

PR #2549, opened 2024-09-11
chore: Update Nix flake

makemake-kbo:
## Description
Updated nix flake.

NO


barnabasbusa:
Closing — this flake.lock update is 18 months stale and conflicts with main. A fresh `nix flake update` would be needed if this is still desired.

PR #2551, opened 2024-09-15
docs: use the right link for docker start install

ilikecodewithmyhand:
Yo, It looks like that your docs path is modified, I fixed it with new path, you are welcome.


tedim52 (approved):

tedim52:
thank you!

PR #2113, opened 2024-02-02
feat: kurtosis github login

tedim52:
## Description:
This change enables GitHub users over Docker to authorize with Kurtosis CLI so that Kurtosis can perform git operations such as cloning packages in private repositories. This primarily enables the use of private GitHub locators for package runs, `import_module`, and `upload_files`.

This is accomplished via an OAuth flow where `kurtosis github login` directs the user to authorize Kurtosis CLI to take actions on their behalf. A token is retrieved upon success and is used by Kurtosis for subsequent git operations. Kurtosis attempts to store the token in secure system storage, but if not found, the token is stored in a plain text file at `kurtosis config path`.

Github commands added:
- `kurtosis github login`
- `kurtosis github logout`
- `kurtosis github token`
- `kurtosis github status`

Flags added (these override existing GitHub login for one off authorization use cases):
- `kurtosis engine start --github-auth-token=< token.txt`
- `kurtosis engine restart --github-auth-token=< token.txt`

## Is this change user facing?
YES

## References:
https://github.com/kurtosis-tech/kurtosis/issues/2020

PR #2560, opened 2024-10-02
feat: support auth from docker config

### LINKED ISSUE #2503: Unable to do dockerhub login
### What's your CLI version?

0.90.1

You can get around getting rate limited by dockerhub by logging in.
 However, dockerhub login with `docker login` does not seem to matter to images that are pulled by kurtosis engine.

### Desired behavior

Would be good to either:

* be able to use the dockerhub credentials present in ~/.docker/config
* be able to pass dockerhub credentials as a runtime parameter

skylenet:
## Description

Supports docker credentials from config (`~/.docker/config.json`). Read https://github.com/kurtosis-tech/kurtosis/issues/2503 

yes

## References (if applicable)
Fixes https://github.com/kurtosis-tech/kurtosis/issues/2503


skylenet:
> #### ⚠️ GitGuardian has uncovered 1 secret following the scan of your pull request.

False positive, it's just `user:password` b64 encoded in a test.

skylenet:
As @pk910 mentioned it to me, this will probably not work since the engine is running in its own container, so there's no access to the users config file there. So we might need to:

- Option 1: Change this fetch the credentials at the kurtosis cli and then propagate them to the engine API. 

- Option 2: Mount the docker config dir to the engine container. Which will probably cause more problems due to it not having access to the different credential helpers that each system uses.

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"encoding/base64"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// WriteStaticConfig writes a static Docker config.json file to a temporary directory
+func WriteStaticConfig(t *testing.T, configContent string) string {
```
tedim52: can we make make this a private helper function if it's not used outside of the package?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		t.Fatalf("Failed to write config.json: %v", err)
+	}
+
+	// Set the DOCKER_CONFIG environment variable to the temp directory
+	os.Setenv("DOCKER_CONFIG", tmpDir)
+	return tmpDir
+}
+
+func TestGetAuthConfigForRepoPlain(t *testing.T) {
```
tedim52, line 30: nice tests : )

tedim52:
Good callout by @pk910. I think `Option1` is the way to go here. I did a similar thing to propagate github auth credentials to the engine. Here is that PR: https://github.com/kurtosis-tech/kurtosis/pull/2113. Specifically the `github_auth_store` finds the token and passes it to the engine. The engine then passes the token to each created enclave.

skylenet:
@tedim52 I've updated the PR, but I'm strugeling testing this end-to-end. 

I'm building the engine with `./engine/scripts/build.sh` and then also the cli with `./cli/scripts/build.sh` . The CLI seems to work fine and I can see my changes there, but somehow I can't get an updated engine container running. 

I noticed that when I build the engine it gets some tag locally.. e.g. `04c515-dirty` , I then ran ` cli/cli/dist/cli_darwin_arm64/kurtosis engine start --version 04c515-dirty` which works fine, and the engine starts. But somehow it doesn't reflect my code changes. E.g. if I change a string from a log line, I don't seem to see that change. 

Example: I wanted to change this specific line for debugging purposes:

```go
return stacktrace.Propagate(err, "Tried pulling image '%v' with platform '%v' but failed: %v", imageName, platform), false
```

To this:
```go
return stacktrace.Propagate(err, "Tried pulling image '%v' with platform '%v' but failed. XXXXX: %v", imageName, platform, imagePullOptions), false
```

But I don't see that change, even after running `./engine/scripts/build.sh` and using the image tag that it printed ( via `kurtosis engine start -version XXXX`)

Any idea?

tedim52:
This is awesome! You likely have to run an engine restart `cli/cli/dist/cli_darwin_arm64/kurtosis engine restart --version 04c515-dirty`. Running an engine start on an already running engine won't start it again even if provided a different version. Once you've restarted, run `engine status` to ensure the version is `04c515-dirty`.

skylenet:
Yeah, I did something similar. I did stop it and then start it with the version flag. I also verified that it was running the right image using ‘engine status’ and also via ‘docker ps’, and it did. But somehow the code change was still not there. I’ll try to figure out the problem today. I was just wondering if I’m doing anything wrong on the local dev workflow.

skylenet:
@tedim52 I think I'm starting to understand the problem. I did that change to that stackTrace.Propagate message and was expecting it to show up at some point at the `engine` container. Which wasn't the case. So just to make sure that my code changes were reflecting the engine container, I just changed some startup debug log there, and I was able to see that.

So, this means that the part that is trying to get the docker auth from the `config.json` is happening somewhere else. This somewhere else seems to be the `api` container. On the `api` container I can see errors like `error reading Docker config file: open /root/.docker/config.json: no such file or directory` . So I was assuming that the engine container would run the `pullImage()` function of the `docker_manager`, but that doesn't seem to be the case. It seems to be happening on the `api` container.  Does it make sense to add the docker `config.json` with the auth creds to the API container?

skylenet:
fyi @tedim52 , I just added the volume mount also to the `api` container. So it has the `config.json` available under  `/root/.docker/config.json` . I've tested this by pulling an image from a private repo on dockerhub and it worked.

So the way it works now is that:

1. Get all your credentials for any registries based on your local docker `config.json`. It should support all ways of obtaining creds, e.g. base64 encoded plain text, os using the docker credential helpers. In my case I'm using ` "credsStore": "osxkeychain"` locally since I'm on a mac. 

2. Create a "docker config" volume (under `/root/.docker`) with a separate `config.json` that is generated based on all credentials that it was able to get from the local system. This `config.json` will have the credentials in plain text, b64 encoded, so that we don't need to deal with having to support all credential helpers on the engine/api containers. 

3. On `imagePull()` we check if an image has a registry defined in the `config.json` with any auth information. If that's the case, we use that as the auth config for `imagePullOptions`. 

4. It's still possible to provide credentials (username/password) via the [ImageSpec](https://docs.kurtosis.com/api-reference/starlark-reference/service-config). This will override any previous auth.  


Some caveats: 
- It's required to restart the engine and recreate any enclave API container if registry credentials are updated/added.

tedim52:
Hey @skylenet I think this approach is really good - and yes regarding where the `imagePull` takes place - the `container-engine-lib` library is used in both the `api` and `engine` containers and `cli` - sometimes the cli/engine will invoke `imagePull` command -we use the same imagePull function to pull the kurtosis images needed for starting enclaves so that's smth important to keep in mind but I think for now, mounting to just the `api` container is ok.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/docker_config_storage_creator/docker_config_storage_creator.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_impls/docker/docker_manager"
+	"github.com/kurtosis-tech/stacktrace"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	// We use this image and version because we already are using this in other projects so there is a high probability
+	// that the image is in the local machine's cache
+	creatorContainerImage = "alpine:3.17"
+	creatorContainerName  = "kurtosis-docker-config-storage-creator"
```
tedim52, line 20: Not something that is blocking this PR but thinking out loud for the future - we have a few of these "config creation" containers that get spun up in the kurtosis lifecycle (fluentbit log creator, docker config, github auth, logs aggregator config) - each one of these gets spun up as its own container and then removed. It would reduce overhead if all these operations were performed on one container (all of them are just rlly mounting to a volume and running a command).

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
```diff
@@ -0,0 +1,178 @@
...
+	configFilePath := os.Getenv("DOCKER_CONFIG")
+	if configFilePath == "" {
+		configFilePath = os.Getenv("HOME") + "/.docker/config.json"
+	} else {
+		configFilePath = configFilePath + "/config.json"
+	}
+
+	file, err := os.ReadFile(configFilePath)
+	if err != nil {
+		return RegistryAuthConfig{}, fmt.Errorf("error reading Docker config file: %v", err)
```
tedim52: Do you mind changing the use `fmt.Errorf` to `stacktrace.Propagate`? We primarily use `stacktrace.Propagate` and `stacktrace.NewError` in our codebase so for consistency would like to stick to using them

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/docker_config_storage_creator/docker_config_storage_creator.go
```diff
@@ -0,0 +1,175 @@
...
+	// Add the auths for each registry
+	for _, registry := range registries {
+		creds, err := docker_manager.GetAuthFromDockerConfig(registry)
+		if err != nil {
+			return stacktrace.NewError("An error occurred getting auth for registry '%v' from Docker config: %v", registry, err)
+		}
+		cfg.Auths[registry] = *creds
+	}
+
+	b, err := json.Marshal(cfg)
```
tedim52: Can we make this variable name more explicit? maybe `cfgJsonStr`?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	configPath := tmpDir + "/config.json"
+	err = os.WriteFile(configPath, []byte(configContent), 0600)
+	if err != nil {
+		t.Fatalf("Failed to write config.json: %v", err)
+	}
+
+	// Set the DOCKER_CONFIG environment variable to the temp directory
+	os.Setenv("DOCKER_CONFIG", tmpDir)
```
tedim52: Can we make `DOCKER_CONFIG` into a shared const variable in `docker_auth.go` since we're using it here and in `loadDockerAuth`?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/docker_config_storage_creator/docker_config_storage_creator.go
skylenet: Makes sense

skylenet:
@tedim52 thank you for the review 🙏  I just updated the PR with the changes that you've requested.

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
```diff
@@ -0,0 +1,183 @@
...
+	cmd := exec.Command(credHelperCmd, "get")
+	cmd.Stdin = strings.NewReader(registryURL)
+
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+
+	if err := cmd.Run(); err != nil {
+		return nil, stacktrace.Propagate(err, "error executing credential helper %s: %s", credHelperCmd, stderr.String())
```
tedim52: I missed this but can we change `credHelperCmd` to `cmd.String()` this way an error outputs `docker-credential-<store> get` or `.. list` as opposed to just `docker-credential-<store>`

tedim52: went ahead and made the change since it was small

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
```diff
@@ -0,0 +1,189 @@
...
+	configFilePath := os.Getenv(ENV_DOCKER_CONFIG)
+	if configFilePath == "" {
+		configFilePath = os.Getenv("HOME") + "/.docker/config.json"
+	} else {
+		configFilePath = configFilePath + "/config.json"
+	}
+
+	file, err := os.ReadFile(configFilePath)
+	if errors.Is(err, os.ErrNotExist) {
+		// If the auth config doesn't exist, return an empty auth config
```
tedim52, line 41: @skylenet I added this check so that if config.json doesn't exist we'll just return empty auth config. What do you think? Otherwise engine won't start if it doesn't exist

tedim52 (approved):

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
skylenet: Yup! Nice catch

PR #2567, opened 2024-10-16
fix: add suffix to fluentbit configurator

### LINKED ISSUE #2568: Fluentbit configurator sticks around when creating enclaves concurrently and prevents new enclaves from being created
### What's your CLI version?

1.3.0

When multiple processes try to create enclaves at the same time, sometimes the fluentbit configurator container sticks around. This then causes subsequent creations to fail. You can see an example of this here: https://app.circleci.com/pipelines/github/ethereum-optimism/optimism/68789/workflows/125efdd1-ffc2-4034-8517-80a841d856d2/jobs/2846767/parallel-runs/0/steps/0-104. 

I tried my hand at fixing this here by adding a suffix to the container name: https://github.com/kurtosis-tech/kurtosis/pull/2567.

### Desired behavior

Concurrent enclaves should be created successfully.

mslipper:
## Description

We run Kurtosis in our CI pipeline. Multiple processes create Kurtosis enclaves concurrenctly. When the timing is just right, it can lead to a race condition where the fluentbit configurator container sticks around and no further enclaves can be created until we manually delete the fluentbit container.

This PR adds a suffix to the configurator container so that it is isolated from other concurrent runs.

No


mslipper:
cc @kurtosis-tech/engineers for review.

tedim52:
Once this is merged, the CI should pass so I'll rerun them. https://github.com/kurtosis-tech/kurtosis/pull/2572.

tedim52 (approved):

PR #2536, opened 2024-08-19
feat: [log retention improvements pt. 2] add flag for configuring log retention period

tedim52:
## Description
Log Retention Improvements:

This PR allows users to configure log retention period with `--log-retention-period` flag on engine starts/restarts. Default is set to 1 week. 

- [x] PR 1: Introduce `LogFileLayout` and `PerWeekLogFileLayout` and adjust `LogFileManager` and `LogsDatabaseClient` to use this for retrieving log file paths
- [x] PR 2: Make retention configurable via a CLI flag
- [ ] PR 3: Implement `PerHourFileLayout` and swap storage format from `PerWeekFileLayout` to `PerHourFileLayout`
- [ ] PR 4: Make `LogsAggregator` use `LogFileLayout` for determining storage format and set it to use `PerHourFileLayout`
------------------------------

## Is this change user facing?
NO

## References 
https://github.com/kurtosis-tech/kurtosis/issues/2443
https://github.com/kurtosis-tech/kurtosis/issues/2190
https://github.com/kurtosis-tech/kurtosis/pull/2534

PR #2570, opened 2024-10-17
feat: [log retention improvements pt. 3] impl and use per hour file layout

tedim52:
## Description
Log Retention Improvements:

- [x] PR 1: Introduce `LogFileLayout` and `PerWeekLogFileLayout` and adjust `LogFileManager` and `LogsDatabaseClient` to use this for retrieving log file paths
- [x] PR 2: Make retention configurable via a CLI flag
- [x] PR 3: Implement `PerHourFileLayout` and swap storage format from `PerWeekFileLayout` to `PerHourFileLayout`
- [ ] PR 4: Make log retention configurable via kurtosis config

YES

## References
https://github.com/kurtosis-tech/kurtosis/pull/2536


engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,141 @@
...
+	// ... enclave-uuid/service-uuid<filetype>
+	perHourFilePathFmtSt = perHourDirPathFmtStr + "%s/%s%s"
+)
+
+type PerHourFileLayout struct {
+	time             logs_clock.LogsClock
+	baseLogsFilePath string
+}
+
+func NewPerHourFileLayout(time logs_clock.LogsClock, baseLogsFilePath string) *PerHourFileLayout {
```
tedim52, line 28: This implementation of `FileLayout` stores log files per hour so that logs can be removed per hour.

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/stream_logs_strategy_impl_test.go
```diff
@@ -0,0 +1,276 @@
...
+	"strings"
+	"testing"
+	"time"
+)
+
+const (
+	retentionPeriodInWeeksForTesting = 5
+)
+
+func TestIsWithinRetentionPeriod(t *testing.T) {
```
tedim52, line 21: these tests were migrated over from the prior `PerWeekStreamLogStrategy`

engine/server/engine/main.go
```diff
@@ -405,27 +410,25 @@ func getKurtosisBackend(ctx context.Context, kurtosisBackendType args.KurtosisBa
...
+		logrus.Infof("Setting log retention period to '%v' hours(s).", logRetentionPeriod.Hours())
 		osFs := volume_filesystem.NewOsVolumeFilesystem()
-		perWeekFileLayout := file_layout.NewPerWeekFileLayout(realTime)
-		logFileManager := log_file_manager.NewLogFileManager(kurtosisBackend, osFs, perWeekFileLayout, realTime, logRetentionPeriodInWeeks)
-		perWeekStreamLogsStrategy := stream_logs_strategy.NewPerWeekStreamLogsStrategy(realTime, logRetentionPeriodInWeeks)
-
-		logsDatabaseClient = persistent_volume.NewPersistentVolumeLogsDatabaseClient(kurtosisBackend, osFs, logFileManager, perWeekStreamLogsStrategy)
+		perHourFileLayout := file_layout.NewPerHourFileLayout(realTime, volume_consts.LogsStorageDirpath)
+		logFileManager := log_file_manager.NewLogFileManager(kurtosisBackend, osFs, perHourFileLayout, realTime, logRetentionPeriod, volume_consts.LogsStorageDirpath)
+		if logFileManager.GetLogFileLayoutFormat() != vector_consts.VectorLogsFilepathFormat {
```
tedim52, line 423: This check ensures that the log file format thats being used to read log files matches the format the logs aggregator is writing log files - if anyone changes the write format or the read format without updating the other, this should fail.

internal_testsuites/golang/testsuite/startlark_user_passing_test/starlark_user_passing_test.go
```diff
@@ -14,7 +14,7 @@ const (
 	userOverrideServiceName = "user-override"
 
 	starlarkScriptWithUserIdPassed = `
-IMAGE = "hyperledger/besu:latest"
+IMAGE = "hyperledger/besu:24.3"
```
tedim52, line 17: This was causing a root user test to fail - it seems in the latest images of besu the default user is `root` so I pinned it to a prior version we were depending on for that test.

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+func (phf *PerHourFileLayout) GetLogFilePaths(
+	filesystem volume_filesystem.VolumeFilesystem,
+	retentionPeriod time.Duration,
+	retentionPeriodIntervals int,
+	enclaveUuid, serviceUuid string,
+) ([]string, error) {
+	var paths []string
+	retentionPeriodInHours := DurationToHours(retentionPeriod)
+
+	if retentionPeriodIntervals < 0 {
```
tedim52, line 55: maybe could find a way to merge this `getLogFilePathsFromNowTillRetentionPeriod` and `getLogFilePathsBeyondRetentionPeriod`?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/consts.go
```diff
@@ -21,12 +21,11 @@ const (
 	fileSinkIdSuffix = "file"
 	fileTypeId       = "\"file\""
 
-	// We instruct vector to store log files per-year, per-week (00-53), per-enclave, per-service
 	// To construct the filepath, we utilize vectors template syntax that allows us to reference fields in log events
 	// https://vector.dev/docs/reference/configuration/template-syntax/
-	baseLogsFilepath = "\"" + logsStorageDirpath + "%%Y/%%V/"
+	baseLogsFilepath = "\"" + logsStorageDirpath + "%%Y/%%V/%%u/%%H/"
```
mieubrisse, line 26: Might be worth a comment explaining what this does, equivalent to the one removed

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+
+const (
+	// basepath year/week/day/hour/
+	perHourDirPathFmtStr = "%s%s/%s/%s/%s/"
+
+	// ... enclave-uuid/service-uuid<filetype>
+	perHourFilePathFmtSt = perHourDirPathFmtStr + "%s/%s%s"
+)
+
+type PerHourFileLayout struct {
```
mieubrisse, line 23: What does this struct actually represent?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func NewPerHourFileLayout(time logs_clock.LogsClock, baseLogsFilePath string) *PerHourFileLayout {
+	return &PerHourFileLayout{
+		time:             time,
+		baseLogsFilePath: baseLogsFilePath,
+	}
+}
+
+func (phf *PerHourFileLayout) GetLogFileLayoutFormat() string {
```
mieubrisse, line 35: Given that this seems to be very vector-specific, should this be called `GetVectorLayoutFormat` or something like that?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+	// Right now this format is specifically made for Vector Logs Aggregators format
+	// This wil be used my Vector LogsAggregator to determine the path to output to
+	return fmt.Sprintf("\"%s%%%%Y/%%%%V/%%%%u/%%%%H/{{ enclave_uuid }}/{{ service_uuid }}.json\"", volume_consts.LogsStorageDirpath)
```
mieubrisse, line 38: in this case, rather than wrestling with all the crazy escaping, might be cleaner to read with a simple `"` + volume_consts.LogsStorageDirpath + `...the rest...`

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (phf *PerHourFileLayout) GetLogFilePath(time time.Time, enclaveUuid, serviceUuid string) string {
+	year, week, day, hour := TimeToWeekDayHour(time)
```
mieubrisse, line 42: Tiny little sixth sense wondering why this is a public function, and if it should be private first (but will wait to see what happens in the rest of the PR)

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+	// basepath year/week/day/hour/
+	perHourDirPathFmtStr = "%s%s/%s/%s/%s/"
+
+	// ... enclave-uuid/service-uuid<filetype>
+	perHourFilePathFmtSt = perHourDirPathFmtStr + "%s/%s%s"
+)
+
+type PerHourFileLayout struct {
+	time             logs_clock.LogsClock
+	baseLogsFilePath string
```
mieubrisse, line 25: Is this a filepath, or a dirpath? I'd expect it to be a directory, and all logs go under a single dir.. right?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+const (
+	// basepath year/week/day/hour/
+	perHourDirPathFmtStr = "%s%s/%s/%s/%s/"
+
+	// ... enclave-uuid/service-uuid<filetype>
+	perHourFilePathFmtSt = perHourDirPathFmtStr + "%s/%s%s"
+)
+
+type PerHourFileLayout struct {
+	time             logs_clock.LogsClock
```
mieubrisse, line 24: Likewise, what is this guy? I initially thought that a "PerHourFileLayout" had one instance per each hour, or something similar. But now I'm seeing that's not the case, so curious about this prop

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
mieubrisse: Can we put this in a code comment so it doesn't get lost when this PR is closed?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+	// This wil be used my Vector LogsAggregator to determine the path to output to
+	return fmt.Sprintf("\"%s%%%%Y/%%%%V/%%%%u/%%%%H/{{ enclave_uuid }}/{{ service_uuid }}.json\"", volume_consts.LogsStorageDirpath)
+}
+
+func (phf *PerHourFileLayout) GetLogFilePath(time time.Time, enclaveUuid, serviceUuid string) string {
+	year, week, day, hour := TimeToWeekDayHour(time)
+	return phf.getHourlyLogFilePath(year, week, day, hour, enclaveUuid, serviceUuid)
+}
+
+func (phf *PerHourFileLayout) GetLogFilePaths(
```
mieubrisse, line 46: Curious about the purpose of this function. At first blush, I'd expect that `PerHourFileLayout` is an abstraction that hides all the internals away.

Also curious why I need to pass in a `filesystem`, `retentionPeriod`, etc. for this call, but not for `GetLogFilePath`?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+	filesystem volume_filesystem.VolumeFilesystem,
+	retentionPeriod time.Duration,
```
mieubrisse, line 48: My gut is saying that it's odd I need to tell the `PerHourFileLayout` what retention it has in a getter method. Wouldn't that stuff be declared in the constructor of the PHFL?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
mieubrisse: That seems sensible to me!

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+	retentionPeriodIntervals int,
+	enclaveUuid, serviceUuid string,
+) ([]string, error) {
+	var paths []string
+	retentionPeriodInHours := DurationToHours(retentionPeriod)
+
+	if retentionPeriodIntervals < 0 {
+		return phf.getLogFilePathsFromNowTillRetentionPeriod(filesystem, retentionPeriodInHours, enclaveUuid, serviceUuid)
```
mieubrisse, line 56: Is there a missing errcheck & `stacktrace.Propagate` here? seems like we wouldn't be capturing the params that we pass to `getLogFilePathsFromNowTillRetentionPeriod`

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+			firstHourWithLogs = i
+			break
+		} else {
+			// return if error is not due to nonexistent file path
+			if !os.IsNotExist(err) {
+				return paths, err
+			}
+		}
+	}
+
```
mieubrisse, line 84: Bug question: if we don't find any log files in the loop above, won't it mean that `firstHourWithLogs` will stay at 0 and then in the loop below will add a bunch of nonexistent paths to the array?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+	// scan for remaining files as far back as they exist before the retention period
+	for i := firstHourWithLogs + 1; i < retentionPeriodInHours; i++ {
+		year, week, day, hour := TimeToWeekDayHour(currentTime.Add(time.Duration(-i) * time.Hour))
+		filePathStr := phf.getHourlyLogFilePath(year, week, day, hour, enclaveUuid, serviceUuid)
+		if _, err := fs.Stat(filePathStr); err != nil {
+			break
```
mieubrisse, line 90: Won't this mean that a single missing file blocks the return of all log files after that (even if they're successful)? Maybe an error would be better here

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+		paths = append(paths, filePathStr)
+	}
+
+	// reverse for oldest to most recent
+	slices.Reverse(paths)
+
+	return paths, nil
+}
+
+func (phf *PerHourFileLayout) getLogFilePathsBeyondRetentionPeriod(fs volume_filesystem.VolumeFilesystem, retentionPeriodInHours int, retentionPeriodIntervals int, enclaveUuid, serviceUuid string) []string {
```
mieubrisse, line 101: Reading this signature, I'm confused about the relationship between `retentionPeriodInHours` and `retentionPeriodIntervals`

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+	var paths []string
+	currentTime := phf.time.Now()
+
+	// scan for log files just beyond the retention period
+	for i := 0; i < retentionPeriodIntervals; i++ {
```
mieubrisse, line 106: I'm sort of confused about this loop - why are we adding `retentionPeriodIntervals` (as `i`) to `retentionPeriodInHours`?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	return paths
+}
+
+func (phf *PerHourFileLayout) getHourlyLogFilePath(year, week, day, hour int, enclaveUuid, serviceUuid string) string {
+	// match the format in which Vector outputs week, hours, days
+	formattedWeekNum := fmt.Sprintf("%02d", week)
+	formattedHourNum := fmt.Sprintf("%02d", hour)
+	return fmt.Sprintf(perHourFilePathFmtSt, phf.baseLogsFilePath, strconv.Itoa(year), formattedWeekNum, strconv.Itoa(day), formattedHourNum, enclaveUuid, serviceUuid, volume_consts.Filetype)
```
mieubrisse, line 123: Small suggestion: might make it easier to read if you newlined each of these

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
```diff
@@ -0,0 +1,139 @@
...
+	// convert sunday in golang's time(0) to sunday (0) in strftime/Vector log aggregator time(7)
+	if day == 0 {
+		day = 7
+	}
+	return year, week, day, hour
+}
+
+func DurationToHours(duration time.Duration) int {
+	return int(math.Ceil(duration.Hours()))
+}
```
mieubrisse, line 139: Okay finishing this class, I think I'm still a bit confused what the purpose of this struct is. I think the confusion is coming from the idea of retention being sort-of baked into this struct, but not fully handled by it.

I think my gut would expect either:
1. this class fully handles retention (maybe some background goroutine that cleans up stuff or something) and it's transparent to the caller of the struct or
2. this struct doesn't know anything about retention - literally just the per-hour file layouting - and whoever calls this struct is responsible for the retention-tracking

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout_test.go
```diff
@@ -0,0 +1,326 @@
...
+	logFilePaths, err := fileLayout.GetLogFilePaths(filesystem, retentionPeriod, -1, testEnclaveUuid, testUserService1Uuid)
+
+	require.NoError(t, err)
+	require.Equal(t, len(expectedLogFilePaths), len(logFilePaths))
+	for i, filePath := range expectedLogFilePaths {
+		require.Equal(t, filePath, logFilePaths[i])
+	}
+}
+
+func TestGetLogFilePathsWithHourlyRetentionAcrossDays(t *testing.T) {
```
mieubrisse, line 63: Nice job on writing all these tests btw!

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout_test.go
```diff
@@ -0,0 +1,326 @@
...
+		hourThreeFp,
+		hourFiveFp,
+	})
+
+	retentionPeriod := 6 * time.Hour // this would return all filepaths, but hour three is missing
+	logFilePaths, err := fileLayout.GetLogFilePaths(filesystem, retentionPeriod, -1, testEnclaveUuid, testUserService1Uuid)
+	require.NoError(t, err)
+	require.Len(t, logFilePaths, 1)
+	require.Equal(t, hourFiveFp, logFilePaths[0]) // should only return hour 5 3 because hour 4 is missing
+}
```
mieubrisse, line 222: Should there be tests for the `GetLogFilePathsBeyondRetention`?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout_test.go
```diff
@@ -0,0 +1,326 @@
...
+			name:          "One hour duration",
+			inputDuration: time.Hour,
+			expectedHours: 1,
+		},
+		{
+			name:          "Fractional hour duration",
+			inputDuration: 90 * time.Minute, // 1.5 hours
+			expectedHours: 2,                // should round up
+		},
+		{
```
mieubrisse, line 299: Nice test cases

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_week_file_layout.go
```diff
@@ -16,30 +16,32 @@ const (
...
-	return &PerWeekFileLayout{time: time}
+func NewPerWeekFileLayout(time logs_clock.LogsClock, baseLogsFilePath string) *PerWeekFileLayout {
+	return &PerWeekFileLayout{time: time, baseLogsFilePath: baseLogsFilePath}
 }
 
 func (pwf *PerWeekFileLayout) GetLogFileLayoutFormat() string {
 	// Right now this format is specifically made for Vector Logs Aggregators format
 	// This wil be used my Vector LogsAggregator to determine the path to output to
-	return "/var/log/kurtosis/%%Y/%%V/{{ enclave_uuid }}/{{ service_uuid }}.json"
+	// is there a way to get rid of the /var/log/kurtosis?
```
mieubrisse, line 38: is this comment still relevant? seems like you actually did get rid of it

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_week_file_layout.go
```diff
@@ -114,11 +116,11 @@ func (pwf *PerWeekFileLayout) getLogFilePathsBeyondRetentionPeriod(fs volume_fil
...
+func (pwf *PerWeekFileLayout) getWeeklyFilePath(year, week int, enclaveUuid, serviceUuid string) string {
+	formattedWeekNum := fmt.Sprintf("%02d", week)
+	return fmt.Sprintf(PerWeekFilePathFmtStr, pwf.baseLogsFilePath, strconv.Itoa(year), formattedWeekNum, enclaveUuid, serviceUuid, volume_consts.Filetype)
 }
 
-func getLogFilePath(year, week int, enclaveUuid, serviceUuid string) string {
-	formattedWeekNum := fmt.Sprintf("%02d", week)
-	return fmt.Sprintf(PerWeekFilePathFmtStr, volume_consts.LogsStorageDirpath, strconv.Itoa(year), formattedWeekNum, enclaveUuid, serviceUuid, volume_consts.Filetype)
+func DurationToWeeks(d time.Duration) int {
+	return int(math.Round(d.Hours() / float64(oneWeekInHours)))
```
mieubrisse, line 125: Is `math.Round` expected? I would have expected that if someone specifies 3.5 weeks, the retention period wouldn't go down

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
```diff
@@ -2,25 +2,30 @@ package log_file_manager
...
+	removeLogsWaitHours = 5 * time.Minute
+
+	createLogsWaitMinutes = 1 * time.Minute
+
+	emptyEnclaveUuid = ""
+
+	retentionPeriodIntervals = 1
 )
 
 // LogFileManager is responsible for creating and removing log files from filesystem.
```
mieubrisse, line 31: From this comment, I'd expect this struct (and only this struct) to be responsible for all retention-related stuff

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
```diff
@@ -143,22 +146,21 @@ func (manager *LogFileManager) RemoveLogsBeyondRetentionPeriod(ctx context.Conte
 			serviceNameStr := string(serviceRegistration.GetName())
 			serviceShortUuidStr := uuid_generator.ShortenedUUIDString(serviceUuidStr)
 
-			retentionPeriod := time.Duration(manager.logRetentionPeriodInWeeks) * oneWeek
-			oldServiceLogFilesByUuid, err := manager.fileLayout.GetLogFilePaths(manager.filesystem, retentionPeriod, 1, string(enclaveUuid), serviceUuidStr)
+			oldServiceLogFilesByUuid, err := manager.fileLayout.GetLogFilePaths(manager.filesystem, manager.logRetentionPeriod, retentionPeriodIntervals, string(enclaveUuid), serviceUuidStr)
```
mieubrisse, line 149: Per comment above, I'd have expected that all the calculations to answer which log files should be pulled respective to retention would be done here, rather than being pushed down to the fileLayout

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
```diff
@@ -143,22 +146,21 @@ func (manager *LogFileManager) RemoveLogsBeyondRetentionPeriod(ctx context.Conte
...
 			if err != nil {
 				logrus.Errorf("An error occurred getting log file paths for service '%v' in enclave '%v' logs beyond retention: %v", serviceUuidStr, enclaveUuid, err)
 			} else {
 				pathsToRemove = append(pathsToRemove, oldServiceLogFilesByUuid...)
 			}
 
-			oldServiceLogFilesByName, err := manager.fileLayout.GetLogFilePaths(manager.filesystem, retentionPeriod, 1, string(enclaveUuid), serviceNameStr)
+			oldServiceLogFilesByName, err := manager.fileLayout.GetLogFilePaths(manager.filesystem, manager.logRetentionPeriod, retentionPeriodIntervals, string(enclaveUuid), serviceNameStr)
```
mieubrisse, line 156: Ditto, etc etc

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
```diff
@@ -253,20 +264,23 @@ func (manager *LogFileManager) createSymlinkLogFile(targetLogFilePath, symlinkLo
...
-	logsDirPathForYear := getLogsDirPathForYear(year)
-	formattedWeekNum := fmt.Sprintf("%02d", week)
-	return fmt.Sprintf("%s%s/", logsDirPathForYear, formattedWeekNum)
-}
-
-// creates a directory path of format /<filepath_base>/year/
-func getLogsDirPathForYear(year int) string {
-	return fmt.Sprintf("%s%s/", volume_consts.LogsStorageDirpath, strconv.Itoa(year))
+// if [enclaveUuid] is empty, gets log file paths from all enclaves
+func (manager *LogFileManager) getAllLogFilePaths(enclaveUuid string) ([]string, error) {
```
mieubrisse, line 268: I think this is totally fine as-is, but I generally avoid encoding an optional type with a "magic" value (even though I know that's Golang's style). I would have indicated that `enclaveUuid` is actually optional rather than required, likely formally  (`Optional[string]`), and would probably call it `maybeEnclaveUuid`. All that said, it's totally a personal style thing (again, I think this is fine as-is, especially for a private function. If it were public, I'd prob be more insistent on the API)

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager.go
```diff
@@ -253,20 +264,23 @@ func (manager *LogFileManager) createSymlinkLogFile(targetLogFilePath, symlinkLo
...
+		if info.IsDir() {
+			return nil
+		}
+		if enclaveUuid == emptyEnclaveUuid || strings.Contains(path, enclaveUuid) {
+			paths = append(paths, path)
+		}
+		return nil
+	}
+	if err := manager.filesystem.Walk(manager.baseFilePath, walkFunc); err != nil {
+		return []string{}, stacktrace.Propagate(err, "An error occurred walking file path.")
```
mieubrisse, line 283: Which file path?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/log_file_manager/log_file_manager_test.go
```diff
@@ -48,7 +50,7 @@ func TestRemoveLogsBeyondRetentionPeriod(t *testing.T) {
 	_, _ = mockFs.Create(week1filepath)
 	_, _ = mockFs.Create(week2filepath)
 
-	logFileManager := NewLogFileManager(mockKurtosisBackend, mockFs, fileLayout, mockTime, 5)
+	logFileManager := NewLogFileManager(mockKurtosisBackend, mockFs, fileLayout, mockTime, persistent_volume_helpers.ConvertWeeksToDuration(5), volume_consts.LogsStorageDirpath)
```
mieubrisse, line 53: Hehe you know my thoughts on public static helper functions, and I thiiiiink you might have one of those related issues here: you have `ConvertWeeksToDuration` in `persistent_volume_helpers`, but `DurationToWeeks` (which is also public) is on the `PerWeekFileLayout`.

This seems likely to either:
- Lead to someone recreating `DurationToWeeks` inside persistent_volume_helpers
- Recreate `WeeksToDuration` inside `PerWeekFileLayout`

All that said I don't think it's the end of the world, but just wondering if you can avoid the public static functions file

engine/server/engine/centralized_logs/client_implementations/persistent_volume/persistent_volume_helpers/persistent_volume_helpers_test.go
```diff
@@ -0,0 +1,44 @@
+package persistent_volume_helpers
+
+import (
+	"testing"
+	"time"
+)
+
+func TestConvertWeeksToDuration(t *testing.T) {
```
mieubrisse, line 8: I love how methodical you are with testing 😍

engine/server/engine/main.go
mieubrisse: Nice context! Can this be a comment to capture that this is an extra sanity check?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/stream_logs_strategy_impl.go
```diff
@@ -22,27 +22,29 @@ import (
 )
 
 const (
-	oneWeek = 7 * 24 * time.Hour
+	nowToRetentionPeriod = -1
 )
 
-// PerWeekStreamLogsStrategy pulls logs from filesystem where there is a log file per year, per week, per enclave, per service
+// StreamLogsStrategyImpl pulls logs from filesystem where there is a log file per year, per week, per enclave, per service
```
mieubrisse, line 28: Is this comment still true? From what I understand, it's now transparent to whether it's per-week or per-hour right?

engine/server/engine/centralized_logs/client_implementations/persistent_volume/stream_logs_strategy/stream_logs_strategy_impl.go
```diff
@@ -22,27 +22,29 @@ import (
...
 // Weeks are denoted 01-52
 // e.g.
 // [.../28/d3e8832d671f/61830789f03a.json] is the file containing logs from service with uuid 61830789f03a, in enclave with uuid d3e8832d671f,
 // in the 28th week of the current year
-type PerWeekStreamLogsStrategy struct {
-	time                      logs_clock.LogsClock
-	logRetentionPeriodInWeeks int
+type StreamLogsStrategyImpl struct {
+	time               logs_clock.LogsClock
+	logRetentionPeriod time.Duration
```
mieubrisse, line 35: I'm trying to figure out why this struct cares about the retention period and fileLayout. From what I understood above, it's the `LogManager` that's responsible for retention, and interfacing with the underlying fileLayout, so wouldn't that  mean that `StreamLogsStrategy` should interface with `LogsManager` instead?

Otherwise, isn't StreamLogsStrategy then going to have to reimplement logic for retention-y stuff? (Or, maybe this is why FileLayout has the retention-y stuff in it, that confused me earlier..? 🤔)

engine/server/engine/centralized_logs/client_implementations/persistent_volume/persistent_volume_logs_database_client_test.go
```diff
@@ -154,6 +157,56 @@ func TestStreamUserServiceLogsPerWeek_WithFilters(t *testing.T) {
 	require.NoError(t, testEvaluationErr)
 }
 
+func TestStreamUserServiceLogsPerHour_WithFilters(t *testing.T) {
```
mieubrisse, line 160: Again, nice job with these tests! I know that testing something as complex as this can feel like a drag, but it makes the code sooo much safer 👍

mieubrisse (approved):
Made it through!

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_hour_file_layout.go
tedim52: mmm yes, will add a test for this and fix : )

engine/server/engine/centralized_logs/client_implementations/persistent_volume/file_layout/per_week_file_layout.go
tedim52: ahh this should be `math.ceil`

barnabasbusa:

Issue #2593, opened 2024-11-07
`kurtosis service logs --follow` creates a busy loop in `closeChannelWhenEmpty`


praetoriansentry:
### What's your CLI version?

1.4.1

### Description & steps to reproduce

Spin up something:
```bash
kurtosis run --enclave my-testnet github.com/ethpandaops/ethereum-package@4.3.0
```

Follow the logs a few times:
```bash
kurtosis service logs --follow my-testnet el-1-geth-lighthouse &
kurtosis service logs --follow my-testnet el-1-geth-lighthouse &
kurtosis service logs --follow my-testnet el-1-geth-lighthouse &
kurtosis service logs --follow my-testnet el-1-geth-lighthouse &
kurtosis service logs --follow my-testnet el-1-geth-lighthouse &
kurtosis service logs --follow my-testnet el-1-geth-lighthouse &
kurtosis service logs --follow my-testnet el-1-geth-lighthouse &
kurtosis service logs --follow my-testnet el-1-geth-lighthouse &
kurtosis service logs --follow my-testnet el-1-geth-lighthouse &
```

At this point everything is fine. But if you stop those processes, each call will use 100% CPU:
```bash
ps aux | grep 'kurtosis service logs' | grep -v grep | awk '{print $2}' | xargs kill
```

At this point the kurtosis-engine process will be using roughly 1000% cpu.
![image](https://github.com/user-attachments/assets/3abd2b3b-a0cd-4c41-9129-5517ee89ee4b)

### Desired behavior

I guess this loop should either have a delay for some kind of upper bound on the amount of time that it will wait.
https://github.com/kurtosis-tech/kurtosis/blob/v1.4.1/engine/server/engine/centralized_logs/client_implementations/persistent_volume/persistent_volume_logs_database_client.go#L178-L185

Since it's looping forever, I would guess the check on `len(logsChan)` isn't doing what we expect.

### What is the severity of this bug?

Papercut; this bug is frustrating, but I have a workaround.

### What area of the product does this pertain to?

CLI: the Command Line Interface

tedim52:
Looking into this now

PR #2594, opened 2024-11-07
fix: remove closeChannelWhenEmpty busy loop

tedim52:
## Description
Addresses https://github.com/kurtosis-tech/kurtosis/issues/2593 by getting rid of the busy loop entirely. The loop was checking to make sure all values from the channel were read before closing the channel.  When following logs, if a stream is cancelled before reading all logs in the channel, it loops infinitely. This changes the consumer to loop until it's read everything from the channel, even after its been closed.

 Also, adds a port for pprof to help with profiling in the future. 

YES 

## References 
https://github.com/kurtosis-tech/kurtosis/issues/2593


praetoriansentry (approved):
Looks good to me.

The only minor comment I would have is that adding an additional HTTP server and port might be unnecessary. The pprof endpoints can be handled over port `9711` as well if that's preferable. That being said, a dedicated pprof port is a bit clearer for someone just running `docker ps` and seeing which ports are mapped.

tedim52:
good point, would prefer kurtosis engine to not take up another port if it doesn't need to so I'll remove 6060

Issue #2634, opened 2025-01-17
Newest Go API version cannot be downloaded by tag


MegaRedHand:
### What's your CLI version?

v2.1.0

### Description & steps to reproduce

Trying to update the dependency to the latest version fails. I tried in multiple ways, but the only way I got it to work is by specifying the commit:

```bash
$ go get github.com/kurtosis-tech/kurtosis/api/golang@84e2fe2
go: downloading github.com/kurtosis-tech/kurtosis/api/golang v1.4.4-0.20250116034716-84e2fe2b35f5
go: downloading github.com/kurtosis-tech/kurtosis v1.4.2-0.20250116034716-84e2fe2b35f5
go: upgraded github.com/kurtosis-tech/kurtosis/api/golang v1.4.3 => v1.4.4-0.20250116034716-84e2fe2b35f5
```

I tried with `@v2.1.0`:

```bash
$ go get github.com/kurtosis-tech/kurtosis/api/golang@v2.1.0
go: github.com/kurtosis-tech/kurtosis/api/golang@v2.1.0: invalid version: module contains a go.mod file, so module path must match major version ("github.com/kurtosis-tech/kurtosis/api/golang/v2")
```

`@v2.1.0`, and appending a `/v2` to the url:
```bash
$ go get github.com/kurtosis-tech/kurtosis/api/golang/v2@v2.1.0
go: github.com/kurtosis-tech/kurtosis/api/golang@v2.1.0: invalid version: module contains a go.mod file, so module path must match major version ("github.com/kurtosis-tech/kurtosis/api/golang/v2")
```

with the release tag (`2.1.0`):
```bash
$ go get github.com/kurtosis-tech/kurtosis/api/golang@2.1.0
go: module github.com/kurtosis-tech/kurtosis@2.1.0 found (v1.4.2-0.20250116034716-84e2fe2b35f5), but does not contain package github.com/kurtosis-tech/kurtosis/api/golang
```

Deleting the dependency and adding it again also doesn't work: it just adds the old `v1.4.3` version to `go.mod`.

### Desired behavior

I expected that running:

```bash
go get -u github.com/kurtosis-tech/kurtosis/api/golang
```

would update the Go API to the latest version. Using it after running that returns:

```
failed to initialize kurtosis context: An error occurred validating the Kurtosis engine API version
[...]

PR #2635, opened 2025-01-18
chore!: migrate golang api to v2

tedim52:
## Description
v1 api available on branch: https://github.com/kurtosis-tech/kurtosis/tree/v1

YES

## References
https://github.com/kurtosis-tech/kurtosis/issues/2634


tedim52:
favoring downgrading to 1.4.4 instead of migrating golang api

PR #2636, opened 2025-01-18
fix: don't get historical enclaves for service logs completion

tedim52:
## Description
Makes it so that autocomplete no longer shows old enclaves.

YES


barnabasbusa (approved):

praetoriansentry (approved):
<3

PR #2640, opened 2025-01-30
fix: flags exclusivity for service logs

christophercampbell:
## Description
This PR adds an exclusivity check for flags for `service logs`, which should not allow `--follow` and `--num N` simultaneously. It would be nice if the command-framework generically supported exclusivity, but the underlying cli library does not seem to support that. 

@tedim52 

YES


christophercampbell:
closing, has issue with detecting present flags

PR #2653, opened 2025-02-15
feat: fluent bit logs collector in k8s backend

tedim52:
## Description
This PR is part 1 in adding centralized logging to the K8s backend. Similar to the Docker backend, this change implements the deployment/management of a [fluentbit](https://docs.fluentbit.io/manual) logs collector's via a k8s `DaemonSet`. The `DaemonSet` deploys fluent bit log collection pods on every node in a k8s cluster. The logs collector is configured to pick up logs from all user services in enclaves running on that node. 

Currently, it's configured to output the log stream to stdout - in pt 2. the logs will be sent to a Vector aggregator.

NO


container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_collector_functions/create_logs_collector.go
```diff
@@ -116,16 +116,16 @@ func CreateLogsCollectorForEnclave(
 		return nil, stacktrace.Propagate(err, "An error occurred getting logs collector object using container ID '%v', labels '%+v', status '%v' and host machine port bindings '%+v'", containerId, containerLabels, defaultContainerStatusForNewLogsCollectorContainer, hostMachinePortBindings)
 	}
 
-	logsCollectorAvailabilityChecker := fluentbit.NewFluentbitAvailabilityChecker(logsCollectorObj.GetBridgeNetworkIpAddress(), logsCollectorObj.GetPrivateHttpPort().GetNumber())
```
tedim52, line 119: `fluentbit.NewFluentbitAvailabilityChecker(logsCollectorObj.GetBridgeNetworkIpAddress(), logsCollectorObj.GetPrivateHttpPort().GetNumber())` was leaking implementation details and also didn't think there needed to be a specific fluent bit availability checker since its a simple http request so i made availability checker a general purpose function

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
```diff
@@ -0,0 +1,461 @@
+package fluentbit
```
tedim52, line 1: The bulk of the important stuff for fluent bit configuration happens here -namespace is created , then config map to hold the configuration, then daemon set.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/consts.go
```diff
@@ -0,0 +1,61 @@
...
+[SERVICE]
+    HTTP_Server       On
+    HTTP_Listen       0.0.0.0
+    HTTP_PORT         9713
+    Parsers_File      /fluent-bit/etc/parsers.conf
+
+[INPUT]
+    Name              tail
+    Tag               kurtosis.*
+    Path              /var/log/containers/*_kt-*_user-service-container-*.log
```
tedim52, line 44: have log collector pick up logs from any log files associated with kt user service containers - this will be parameterized /templated in the next PR

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/logs_collector_object_attributes_provider.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (provider *kubernetesLogsCollectorObjectAttributesProviderImpl) ForLogsCollectorDaemonSet() (KubernetesObjectAttributes, error) {
+	name, err := getCompositeKubernetesObjectName([]string{logsCollectorNamePrefix, string(provider.logsCollectorGuid)})
+	if err != nil {
+		return nil, err
+	}
+
+	labels := map[*kubernetes_label_key.KubernetesLabelKey]*kubernetes_label_value.KubernetesLabelValue{
+		kubernetes_label_key.KurtosisResourceTypeKubernetesLabelKey: label_value_consts.LogsCollectorKurtosisResourceTypeKubernetesLabelValue,
```
tedim52, line 47: I considered giving the logs collector a guid but made it through with just this label being sufficient for managing the logs collector retrieval/creation/removal - LMK if you think there's any other good reason to add a unique guid for the logs collector

engine/server/engine/enclave_manager/enclave_creator.go
```diff
@@ -85,20 +85,18 @@ func (creator *EnclaveCreator) CreateEnclave(
...
-		defer func() {
-			if shouldDeleteLogsCollector {
-				err = creator.kurtosisBackend.DestroyLogsCollectorForEnclave(teardownCtx, enclaveUuid)
-				if err != nil {
-					logrus.Errorf("Couldn't cleanup logs collector for enclave '%v' as the following error was thrown:\n%v", enclaveUuid, err)
-				}
-			}
-		}()
+	// TODO the logs collector has a random private ip address in the enclave network that must be tracked
+	if _, err := creator.kurtosisBackend.CreateLogsCollectorForEnclave(setupCtx, enclaveUuid, defaultHttpLogsCollectorPortNum, defaultTcpLogsCollectorPortNum); err != nil {
```
tedim52, line 89: now that kubernetes backend implements it, we can call this function at the engine level - at this point the engine should've already started the logs collector daemonset, so this will just serve as a check that its up and running for the k8s backend

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -250,6 +253,22 @@ func CreateEngine(
 			return nil, stacktrace.Propagate(err, "An error occurred waiting for the engine grpc proxy port '%v/%v' to become available", privateGrpcProxyPortSpec.GetTransportProtocol(), privateGrpcProxyPortSpec.GetNumber())
 		}*/
 
+	logrus.Infof("Starting the centralized logs components...")
+	logsCollectorDaemonSet := fluentbit.NewFluentbitLogsCollector()
+
+	_, removeLogsCollectorFunc, err := logs_collector_functions.CreateLogsCollector(ctx, logsCollectorTcpPortNum, logsCollectorHttpPortNum, logsCollectorDaemonSet, nil, kubernetesManager, objAttrsProvider)
```
tedim52, line 261: Unlike the `DockerBackend`, where the log collectors are deployed by the engine during enclave creation, in the `k8s` backend, the logs collector lifecycle gets managed with the engine's and is created during engine creation.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/consts.go
```diff
@@ -0,0 +1,62 @@
+package fluentbit
+
+const (
+	healthCheckEndpointPath = "api/v1/health"
+
+	fluentBitContainerName = "fluent-bit"
+	fluentBitImage         = "fluent/fluent-bit:latest-debug"
```
tedim52: using debug image for now with testing toolkit (`curl`, `netstat`) eventually will move to latest regular img

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_container.go
```diff
@@ -8,12 +8,25 @@ import (
 	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_interface/objects/enclave"
 	"github.com/kurtosis-tech/stacktrace"
 	"github.com/sirupsen/logrus"
+	"net"
 )
 
-type fluentbitLogsCollectorContainer struct{}
+type fluentbitLogsCollectorContainer struct {
+	fluentbitHealthCheckPath string
```
mieubrisse: Does this change? Being a variable, I'd expect it to vary by container, but that seems odd to me

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_container.go
```diff
@@ -8,12 +8,25 @@ import (
...
+	portNumber uint16
+}
 
 func NewFluentbitLogsCollectorContainer() *fluentbitLogsCollectorContainer {
-	return &fluentbitLogsCollectorContainer{}
+	return &fluentbitLogsCollectorContainer{
+		fluentbitHealthCheckPath: healthCheckEndpointPath,
+
+		// these will get set after container is created
+		bridgeNetworkIpAddress: nil,
```
mieubrisse: This makes the object mutable, which generally introduces complexity. Is it necessary to create the object and mutate it later, or can we just create the object with the bridgeNetwork and portNumber as params to the constructor?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_container.go
```diff
@@ -81,7 +93,7 @@ func (fluentbitContainer *fluentbitLogsCollectorContainer) CreateAndStart(
 		volumeLabelStrs[labelKey.GetString()] = labelValue.GetString()
 	}
 
-	//This method will create the volume if it doesn't exist, or it will get it if it exists
+	//This method doesn't will create the volume if it dexist, or it will get it if it exists
```
mieubrisse: Oops looks like some find-replace issues

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/engine_functions/create_engine.go
mieubrisse: Good info; seems worthy of putting this as a comment 👍

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/engine_functions/stop_engines.go
```diff
@@ -68,5 +69,10 @@ func StopEngines(
 		successfulEngineGuids[engineGuid] = true
 	}
 
+	// Stop centralized logging components
+	if err := logs_collector_functions.DestroyLogsCollector(ctx, kubernetesManager); err != nil {
+		return nil, nil, stacktrace.Propagate(err, "An error occurred removing the logs colector .")
```
mieubrisse: s/logs colector ./logs collector./g

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/kubernetes_kurtosis_backend.go
```diff
@@ -458,18 +460,44 @@ func (backend *KubernetesKurtosisBackend) DestroyLogsAggregator(ctx context.Cont
 }
 
 func (backend *KubernetesKurtosisBackend) CreateLogsCollectorForEnclave(ctx context.Context, enclaveUuid enclave.EnclaveUUID, logsCollectorHttpPortNumber uint16, logsCollectorTcpPortNumber uint16) (*logs_collector.LogsCollector, error) {
-	// TODO IMPLEMENT
-	return nil, stacktrace.NewError("Creating the logs collector isn't yet implemented on Kubernetes")
+	// TODO: check that logs aggregator exists first when logs aggregator is implemented
```
mieubrisse: Not sure I understand this comment

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/kubernetes_kurtosis_backend.go
```diff
@@ -458,18 +460,44 @@ func (backend *KubernetesKurtosisBackend) DestroyLogsAggregator(ctx context.Cont
...
+		ctx,
+		logsCollectorTcpPortNumber,
+		logsCollectorHttpPortNumber,
+		logsCollectorDaemonSet,
+		nil, // TODO: provide when logs aggregator is implemented
+		backend.kubernetesManager,
+		backend.objAttrsProvider,
+	)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred creating the logs collector using the '%v' TCP port number, the '%v' HTTP port number and the los collector container '%+v'", logsCollectorTcpPortNumber, logsCollectorHttpPortNumber, nil)
```
mieubrisse: Prob worth putting a TODO to remember to swap the `nil` at the end with actual info

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/create_logs_collector.go
```diff
@@ -0,0 +1,101 @@
...
+	error,
+) {
+	var logsCollectorObj *logs_collector.LogsCollector
+	var kubernetesResources *logsCollectorKubernetesResources
+	var removeLogsCollectorFunc func()
+	var err error
+
+	logsCollectorObj, kubernetesResources, err = getLogsCollectorObjAndResourcesForCluster(ctx, kubernetesManager)
+	if err != nil {
+		return nil, nil, stacktrace.Propagate(err, "An error occurred getting logs aggregator container.")
```
mieubrisse: Isn't "logs collector" distinct from "logs aggregator"? My understanding is that this is the logs collector, and logs aggregator is still yet to be implemented

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/create_logs_collector.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	shouldRemoveLogsCollector := true
+	if logsCollectorObj != nil {
```
mieubrisse, line 47: It seems like this function is trying to be idempotent. If so, shouldn't `shouldRemoveLogsCollector` be inside the "we're creating a new logs collector" branch? Otherwise, it seems like we could get into a situation where this accidentally _destroys_ an already-existing logs collector

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/create_logs_collector.go
```diff
@@ -0,0 +1,101 @@
...
+			logsCollectorHttpPortNumber,
+			logsCollectorTcpPortId,
+			logsCollectorHttpPortId,
+			objAttrsProvider,
+			kubernetesManager,
+		)
+		if err != nil {
+			return nil, nil, stacktrace.Propagate(
+				err,
+				"An error occurred starting the logs collector daemon set with logs aggregator host '%v', logs aggregator port '%v', HTTP port number '%v', TCP port id '%v', and HTTP port id '%v'",
```
mieubrisse: Nice details

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/destroy_logs_collector.go
```diff
@@ -0,0 +1,46 @@
...
+		logsCollectorNamespace = logsCollectorResources.namespace
+	} else {
+		// assume if there is no namespace, there's no logs collector but log a debug in case
+		logrus.Debug("No logs collector namespace found. Returning without attempting to destroy remaining logs collector resources.")
+		return nil
+	}
+
+	if logsCollectorResources.daemonSet != nil {
+		if err := kubernetesManager.RemoveDaemonSet(ctx, logsCollectorNamespace.Name, logsCollectorResources.daemonSet); err != nil {
+			return stacktrace.Propagate(err, "An error occurred removing logs collector daemon set.")
```
mieubrisse: Just checking, an error that causes an eject could leave partially-created stuff behind (e.g. if we eject here, we leave the ConfigMap and Namespace behind). That's not going to cause any problems for anything else, right?

The way I used to think about this is, the "identifying resource" for a thing would be the last thing I removed. E.g. an enclave is identified by the Docker subnet, so as long as the subnet exists there are some resources for the enclave remaining.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/consts.go
mieubrisse: Seems like there should be a commnet explaining this, as there's a high risk of forgetting

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
```diff
@@ -0,0 +1,465 @@
...
+	}
+
+	containerPorts, err := shared_helpers.GetKubernetesContainerPortsFromPrivatePortSpecs(privatePorts)
+	if err != nil {
+		return nil, nil, nil, nil, stacktrace.Propagate(err, "An error occurred getting the logs collector fluent bit container ports from the port specs")
+	}
+
+	daemonSet, err := createLogsCollectorDaemonSet(ctx, namespace.Name, configMap.Name, containerPorts, logsCollectorAttrProvider, kubernetesManager)
+	if err != nil {
+		return nil, nil, nil, nil, stacktrace.Propagate(err, "An error occurred while trying to daemonset for fluent-bit log collector.")
```
mieubrisse: "while trying to daemonset" seems like it's missing a word

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
```diff
@@ -0,0 +1,465 @@
...
+		}
+	}()
+
+	// wait until the first pod associated with this daemon set is online before returning
+	if err = waitForAtLeastOneActivePodManagedByDaemonSet(ctx, daemonSet, kubernetesManager); err != nil {
+		return nil, nil, nil, nil, stacktrace.Propagate(err, "An error occurred waiting for at least one active pod managed by logs collector daemon set '%v'", daemonSet.Name)
+	}
+
+	removeLogsCollectorFunc := func() {
+		removeDaemonSetFunc()
```
mieubrisse, line 159: Nice, I like how you assigned these to variables so you can use them above in the `defer`s and also here

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
```diff
@@ -0,0 +1,465 @@
...
+	name := daemonSetAttrProvider.GetName().GetString()
+	labels := shared_helpers.GetStringMapFromLabelMap(daemonSetAttrProvider.GetLabels())
+	annotations := shared_helpers.GetStringMapFromAnnotationMap(daemonSetAttrProvider.GetAnnotations())
+
+	containers := []apiv1.Container{
+		{
+			Name:  fluentBitContainerName,
+			Image: fluentBitImage,
+			Args: []string{
+				"/fluent-bit/bin/fluent-bit",
```
mieubrisse, line 197: Oh huh, sort of weird this is necessary. I would have thought that we wouldn't need to specify container args for this to work?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
```diff
@@ -0,0 +1,465 @@
...
+
+	return logsCollectorDaemonSet, nil
+}
+
+func createLogsCollectorConfigMap(
+	ctx context.Context,
+	namespace string,
+	objAttrProvider object_attributes_provider.KubernetesLogsCollectorObjectAttributesProvider,
+	kubernetesManager *kubernetes_manager.KubernetesManager) (*apiv1.ConfigMap, error) {
+	configMapAttrProvider, err := objAttrProvider.ForLogsCollectorConfigMap()
```
mieubrisse, line 344: Man all the work we put into naming things correctly and explicitly really paid off. I'm having quite the easy time reading this PR as a result

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
```diff
@@ -0,0 +1,465 @@
...
+
+	return configMap, nil
+}
+
+func createLogsCollectorNamespace(
+	ctx context.Context,
+	objAttrProvider object_attributes_provider.KubernetesLogsCollectorObjectAttributesProvider,
+	kubernetesManager *kubernetes_manager.KubernetesManager,
+) (*apiv1.Namespace, error) {
+	namespaceProvider, err := objAttrProvider.ForLogsCollectorNamespace()
```
mieubrisse: this hsould be `namespaceAttrProvider` right, since it doesn't provide namespaces themselves?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -0,0 +1,238 @@
...
+			)
+		}
+		if len(logsCollectorNamespaceForLabel) == 0 {
+			// if no namespace for logs collector, assume it doesn't exist at all
+			return &logsCollectorKubernetesResources{
+				daemonSet: nil,
+				configMap: nil,
+				namespace: nil,
+			}, nil
+		} else {
```
mieubrisse: Slight simplification: you eject-early, eject-often in the `if` clause (good!) so you don't need the `else`

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+	} else {
+		return &logsCollectorKubernetesResources{
+			daemonSet: nil,
+			configMap: nil,
+			namespace: nil,
+		}, nil
+	}
+
+	logsCollectorCfgConfigMaps, err := kubernetes_resource_collectors.CollectMatchingConfigMaps(
```
mieubrisse: is `CfgConfigMaps` a typo?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -0,0 +1,238 @@
...
+		configMap: configMap,
+		namespace: namespace,
+	}
+
+	return logsCollectorKubernetesResources, nil
+}
+
+func getLogsCollectorsObjectFromKubernetesResources(ctx context.Context, kubernetesManager *kubernetes_manager.KubernetesManager, logsCollectorKubernetesResources *logsCollectorKubernetesResources) (*logs_collector.LogsCollector, error) {
+	if logsCollectorKubernetesResources.namespace == nil || logsCollectorKubernetesResources.daemonSet == nil || logsCollectorKubernetesResources.configMap == nil {
+		// if any resources not found for logs collector, don't return an object
```
mieubrisse, line 144: On one hand I like this because it declares "a logs collector object in Go only exists if all the k8s resources exist". On the other hand it seems like it could potentially fool code that calls this function into thinking there are no logs collector k8s resources when there could be some. No suggestion, just bringing for your consideration.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -0,0 +1,238 @@
...
+		bridgeNetworkIpAddr net.IP
+		err                 error
+	)
+
+	logsCollectorStatus, err = getLogsCollectorStatus(ctx, kubernetesManager, logsCollectorKubernetesResources.daemonSet)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred getting the status of the logs collector.")
+	}
+
+	// daemon sets don't have an entrypoint so leave these blank for now
```
mieubrisse: Reading this I thought "entrypoint" = Dockerfile's ENTRYPOINT, and was confused what it has to do with IPs. Here I'm less sure; is entrypoint a Kubernetes networking term?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -0,0 +1,238 @@
...
+// in the meantime logs collector status is container.ContainerStatus_Running if all pods managed by the logs collector DaemonSet are running
+// if one is failing/or stopped, the logs collector is to considered to be stopped
+func getLogsCollectorStatus(ctx context.Context, kubernetesManager *kubernetes_manager.KubernetesManager, logsCollectorDaemonSet *v1.DaemonSet) (container.ContainerStatus, error) {
+	logsCollectorPods, err := kubernetesManager.GetPodsManagedByDaemonSet(ctx, logsCollectorDaemonSet)
+	if err != nil {
+		return container.ContainerStatus_Stopped, stacktrace.Propagate(err, "An error occurred getting pods managed by logs collector daemon set '%v'.", logsCollectorDaemonSet.Name)
+	}
+	if len(logsCollectorPods) < 1 {
+		// if there are no pods associated with logs collector daemon set, assume something is wrong and err
+		return container.ContainerStatus_Stopped, stacktrace.NewError("No pods managed by logs collector daemon set were found. There should be at least one. This is likely a bug in Kurtosis.")
```
mieubrisse, line 187: If a daemonset's pod is in the process of being bounced, will it show up here? I.e. if the daemonset pod is in a crash-restart loop, will the pod still show up (with a "restarting" condition or something) or will it be invisible? Just food for thought.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -0,0 +1,238 @@
...
+		podStatus, err := shared_helpers.GetContainerStatusFromPod(pod)
+		if err != nil {
+			return container.ContainerStatus_Stopped, stacktrace.Propagate(err, "An error occurred retrieving container status for a pod managed by logs collectors collector daemon set '%v' with name: %v\n", logsCollectorDaemonSet.Name, pod.Name)
+		}
+
+		switch podStatus {
+		case container.ContainerStatus_Running:
+			continue
+		case container.ContainerStatus_Stopped:
+			logsCollectorStatus = container.ContainerStatus_Stopped
```
mieubrisse: do you even need the `logsCollectorStatus` variable? seems you could just `return container.ContainerStatus_Stopped` and `return container.ContainerStatus_Running` at the end of the function

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -0,0 +1,238 @@
...
+		)
+	}
+	for _, pod := range pods {
+		if len(pod.Spec.Containers) < 1 {
+			return stacktrace.NewError("Pod '%v' managed by logs collector daemon set '%v' doesn't have any containers associated with it. There should be at least one container.", pod.Name, logsCollectorDaemonSet.Name)
+		}
+
+		// NOTE: ideally we'd actually curl the health endpoint of the fluent bit container (like we do for Docker)
+		// this part of code runs on users machine logs collector isn't exposed so we exec onto the pod containers - which may not have curl on them, hence netstat check
+		if err = shared_helpers.WaitForPortAvailabilityUsingNetstat(kubernetesManager, k8sResources.namespace.Name, pod.Name, pod.Spec.Containers[0].Name, httpPortSpec, maxAvailabilityCheckRetries, timeToWaitBetweenChecksDuration); err != nil {
```
mieubrisse: Nit: this line feels really big; prob worth breaking up with local variables

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/consts/kubernetes_consts.go
```diff
@@ -24,6 +24,8 @@ const (
 	PersistentVolumesKubernetesResource      = "persistentvolumes"
 	PersistentVolumeClaimsKubernetesResource = "persistentvolumeclaims"
 	IngressesKubernetesResource              = "ingresses"
+	ConfigMapKubernetesResource              = "configmaps"
```
mieubrisse: Miniscule nit (mi-nit-scule?): think this should be plural, ConfigMaps

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/logs_collector_object_attributes_provider.go
mieubrisse: Only reason I can think of is a creation-deletion cycle that happens quickly, which could end up with grabbing a still-deleting collector and then the new one. But not sure if that's a real risk

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/logs_collector_object_attributes_provider.go
```diff
@@ -0,0 +1,98 @@
...
+func newKubernetesLogsCollectorObjectAttributesProvider(logsCollectorGuid logs_collector.LogsCollectorGuid) *kubernetesLogsCollectorObjectAttributesProviderImpl {
+	return &kubernetesLogsCollectorObjectAttributesProviderImpl{
+		logsCollectorGuid: logsCollectorGuid,
+	}
+}
+
+func (provider *kubernetesLogsCollectorObjectAttributesProviderImpl) ForLogsCollectorDaemonSet() (KubernetesObjectAttributes, error) {
+	name, err := getCompositeKubernetesObjectName([]string{logsCollectorNamePrefix, string(provider.logsCollectorGuid)})
+	if err != nil {
+		return nil, err
```
mieubrisse: looks like we're not stacktrace.Propagate'ing here and below. Intentional?

mieubrisse (approved):

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_container.go
tedim52: agreed, got rid of those

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_container.go
tedim52: doesn't change, its returned as a const so removed this field

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_container.go
tedim52: fixed good catch

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/kubernetes_kurtosis_backend.go
tedim52: reworded for clarity

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/kubernetes_kurtosis_backend.go
tedim52: just replaced it with the `logsCollectorDaemonSet`

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/create_logs_collector.go
tedim52: yep that was a copy paste left unchanged - just updated to correct error statement

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/logs_collector_object_attributes_provider.go
tedim52: good point - if we find this to be an issue can add then, will leave as is for now

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/create_logs_collector.go
tedim52: ah originally - my thinking was:
- we find a logs collector
- we don't find a logs collector so we create one

- either way, check the logs collector is available
- if its not available, remove it (even if it already existed)

but the issue is I'm now seeing - is in case 1 - where there already exists a logs collector, no `removeLogsCollectorFunc` gets set, so it wouldn't be removed anyways. I did as u suggested and moved it inside.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/destroy_logs_collector.go
tedim52: So in this case we actually are cleaning up the identifying resource  (the namespace) last. 

Say, the daemon set was removed successfully, but the config map failed to be removed. The namespace (and potentially config map) are left behind. 

My thinking is this function is to make this function idempotent so a user would either manually clean up or this function just gets called again in which case `daemonSet` is `nil` , but (maybe?) config map and namespace aren't and an attempt is made to remove them again.

What do you think?

tedim52: We could follow what `[DestroyEngines](https://github.com/kurtosis-tech/kurtosis/blob/bf315c6993d5e70a47265f1f41f0f7273fdd904b/container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/engine_functions/destroy_engines.go#L59)` is doing and return the last error that happens. 

In which case an attempt would be made to destory each object. But in that case - say the config map fails to get removed, but removing the namespace is successful. The config map err is returned. There would be no way to remove the config map on a subsequent call to the function bc the identifying namespace doesn't exist.

tedim52: Well actually, if the namespace is removed then all the resources in it should get removed too so that may work actually. But in that case why not just remove only the log collector namespace? Cleaner clean up?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
tedim52: I think I'm okay with this if the removal logic is safe enough. Semantically, I want callers of this function to take the view that "there is only a logs collector if all k8s resources exist for there to be one". 

I'll add function spec explaining this behavior to make it explicit to consumers of the function

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
tedim52: I meant to convey generally a "way to access the daemonset from outside the cluster" - probably closer to an ingress. but I get the confusion here so I'll change it.

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/logs_collector_object_attributes_provider.go
tedim52: intentional laziness

tedim52: jk that err already comes wrapped w a detailed propagate from `getCompositeKubernetesObjectName` so i didnt see value in wrapping again at this layer, added a comment explaining that

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
tedim52: Yep - `GetPodsManagedByDaemonSet` returns all pods that have the logs collector label on it regardless of status

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
tedim52: explicit >> implicit

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
tedim52: great point

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
tedim52: using the same config the official fluent bit helm chart uses: https://github.com/fluent/helm-charts/blob/f87fec5d36cf871ad5174c021b8d6ceb6a1b2001/charts/fluent-bit/values.yaml#L489

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/destroy_logs_collector.go
mieubrisse: Yeah destroying the entire namespace (including the configmap and such in it) actually seems like the easiest thing

Issue #2649, opened 2025-02-07
Add `acceptable_codes` to `plan.run_sh` and `plan.exec`


tedim52:
### Background & motivation

`plan.exec` has an `acceptable_codes` field that fails the starlark script if one of the acceptable_codes isn't set.

### Desired behaviour

 `plan.run_sh` and `plan.run_python` should have the field on it. 

### How important is this to you?

Painful; the lack of this feature makes using Kurtosis frictionful.

### What area of the product does this pertain to?

Other: anything not covered by the above

tedim52:
cc. @samlaf

samlaf:
Not familiar at all with the kurtosis codebase, but just looking at https://github.com/kurtosis-tech/kurtosis/blob/afe0d87dad2e17e64ced1d1e00bb5b20d6745416/core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_sh.go#L298 doesn't this check for a non-zero error code and return an error? Maybe the error is not handled properly?

PR #2654, opened 2025-02-19
fix: exit codes on tasks

tedim52:
## Description
... and adds `acceptable_codes` and `skip_code_check` fields to `plan.run-sh` and `plan.run_python`, similar to `plan.exec`

Exit codes on tasks were broken by this [change](https://github.com/kurtosis-tech/kurtosis/pull/2499) - the exit code was from the last `echo` which was always 0 / success, meaning even if `commandToRun` failed, `run_sh` would return success. This fixes that issue by saving the exit code of the original `commandToRun` and exiting with that as the final step.

TODO:

- [X] update docs
- [X] add some acceptable code tests
- [X] remove reliance on bash if possible

YES

## References
https://github.com/kurtosis-tech/kurtosis/pull/2499
Addresses https://github.com/kurtosis-tech/kurtosis/issues/2649 and https://github.com/kurtosis-tech/kurtosis/issues/2649


core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
```diff
@@ -58,15 +61,22 @@ const (
...
-// Adds an extra echo to ensure each log ends with a newline.
-// Uses tee to direct output to the task log file while maintaining output to stdout.
-// Redirects stderr to stdout.
+// This command is crafted carefully to allow outputting logs to task log file, outputting to stdout, and retaining the exit code from [commandToRun]
+// Solution is adapted from 3rd answer in this stack exchange post: https://unix.stackexchange.com/questions/14270/get-exit-status-of-process-thats-piped-to-another. Read detailed explanation of command.
+// compared to solution in post, an extra echo >> is added to add a newline after all logs are processed, this is a hack to make sure the last log line gets processed runCommandToStreamTaskLogs
 func getCommandToRunForStreamingLogs(commandToRun string) []string {
-	return []string{shellWrapperCommand, "-c", fmt.Sprintf("{ %v; } %v %v %v %v %v %v %v %v", commandToRun, "2>&1", "|", "tee", taskLogFilePath, "&&", "echo", ">>", taskLogFilePath)}
+	fullCmd := []string{shellWrapperCommand, "-c", fmt.Sprintf("{ { { { %v; echo $? >&3; } | tee %v >&4; echo >> %v; } 3>&1; } | { read xs; exit $xs; } } 4>&1", commandToRun, taskLogFilePath, taskLogFilePath)}
```
tedim52, line 79: @mieubrisse Ended up going with [this](https://unix.stackexchange.com/questions/14270/get-exit-status-of-process-thats-piped-to-another) solution instead of of `/tmp/exec-succeeded` solution bc `/tmp/exec-succeeded`, `rm, `test -e` bc 1) couldn't capture non 0,1 codes returned (eg. `return 42`) although maybe it could be modified though?) also see `run_task_sh_task_test.go` 2) small plus that this solution doesn't open/rely on another file so minimizes potential kurtosis side effects.

Thus far, this solution has only failed for me when doing an explicit `run=exit 0`, bc the `exit 0` will exit the process before echoing the status code, thus `read xs` gets no input and errors (causing an exit code of 1) 

I feel comfortable assuming users won't do an explicit `run=.....;...;exit 0`. Could make that explicit in the docs? What do you think?

tedim52:
deploy docs CI job is failing which I need to fix but shouldn't block this PR

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_python.go
```diff
@@ -127,6 +141,8 @@ func NewRunPythonService(
 				wait:                   DefaultWaitTimeoutDurationStr,
 				description:            "",  // populated at interpretation time
 				returnValue:            nil, // populated at interpretation time
+				skipCodeCheck:          false,
+				acceptableCodes:        []int64{},
```
mieubrisse: Should this include `0` by default?

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_sh.go
```diff
@@ -109,6 +121,8 @@ func NewRunShService(
 				wait:                   DefaultWaitTimeoutDurationStr,
 				description:            "",  // populated at interpretation time
 				returnValue:            nil, // populated at interpretation time
+				skipCodeCheck:          false,
+				acceptableCodes:        []int64{},
```
mieubrisse: Same question as above

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
```diff
@@ -58,15 +61,22 @@ const (
 	tiniEnabled = true
 )
 
+var defaultAcceptableCodes = []int64{
```
mieubrisse, line 64: oh looks like you already have this; should this be used up above, when you're creating the default RunPython/RunSh objects?

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
```diff
@@ -58,15 +61,22 @@ const (
...
+	0, // EXIT_SUCCESS
+}
+
+// runCommandToStreamTaskLogs sets the entrypoint of a task container with a command that creates and tails a log file for tasks run on the container
+// all tasks redirect output to the task log file (see getCommandToRunForStreamingLogs for details) where they will be picked up by the main process
+// and streamed to stdout via tail -F
```
mieubrisse, line 70: Perhaps worth explaining "why" too

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
mieubrisse: Ah, yeah right now you're directly executing the user's code inside the same process. You don't want that (what does the exit code even mean at that point? the exit code of the last command). Instead, execute their stuff inside its own subprocess, e.g. `sh '%v'`, and probably even better would be to write the user's commands to a script and execute that (e.g. `sh the-exec-script.sh`).  I think it'll make it easier for users to think about the 'sh' command too: "You write whatever you'd write in a script, and we use that".

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
```diff
@@ -337,3 +347,14 @@ func getTaskNameFromArgs(arguments *builtin_argument.ArgumentValuesSet) (string,
 		return fmt.Sprintf("task-%v", randomUuid.String()), nil
 	}
 }
+
+func isAcceptableCode(acceptableCodes []int64, recipeResult map[string]starlark.Comparable) bool {
+	isAcceptable := false
+	for _, acceptableCode := range acceptableCodes {
+		if recipeResult[runResultCodeKey] == starlark.MakeInt64(acceptableCode) {
+			isAcceptable = true
```
mieubrisse: Can simplify this function with `return true` here and `return false` at the end

docs/docs/api-reference/starlark-reference/plan.md
```diff
@@ -570,6 +570,15 @@ The `run_python` instruction executes a one-time execution task. It runs the Pyt
             "/coinbase/address.json"
         ],
 
+        # If the recipe returns a code that does not belong on this list, this instruction will fail.
```
mieubrisse: maybe "that does not belong on" -> "that isn't a part of", and "this instruction will fail" -> "this instruction will fail at execution time"?

docs/docs/api-reference/starlark-reference/plan.md
```diff
@@ -570,6 +570,15 @@ The `run_python` instruction executes a one-time execution task. It runs the Pyt
...
+        # OPTIONAL (Defaults to [0])
+        acceptable_codes = [0, 0], # Here both 0 and 1 are valid codes that we want to accept and not fail the instruction
```
mieubrisse: Looks like a typo; should be `[0, 1]` right?

docs/docs/api-reference/starlark-reference/plan.md
```diff
@@ -655,6 +664,15 @@ The `run_sh` instruction executes a one-time execution task. It runs the bash co
...
```
mieubrisse: Same comments as above

docs/docs/api-reference/starlark-reference/plan.md
```diff
@@ -570,6 +570,15 @@ The `run_python` instruction executes a one-time execution task. It runs the Pyt
...
+
+        # If False, instruction will never fail based on code (acceptable_codes will be ignored).
+        # You can chain this call with assert to check codes after request is done.
+        # OPTIONAL (Defaults to False)
+        skip_code_check = False,
```
mieubrisse, line 580: IMO a simpler way to implement this would have been to have `acceptable_codes` take in `None` to mean "accept every code" rather than adding a new parameter, but it's in now

mieubrisse (approved):
Some typos and a broader question about how commands are executed, but otherwise good

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
mieubrisse: This will also allow you to take `exit X` (and I suspect there are a class of other, similar commands that you might not be able to handle with the current approach)

tedim52: - so note - this solution is focused around ensuring the exit code is that of the users command. overall, i agree with you on this approach. if we experience any issues with this solution, i'll implement "write to a shell script, execute shell script" approach.

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/run_python.go
tedim52: gets populated at interpretation time, added a comment clarifying this

core/server/api_container/server/startosis_engine/kurtosis_instruction/tasks/tasks_shared.go
tedim52: yep that is what is populated during interpretation time if the user doesn't supply them

docs/docs/api-reference/starlark-reference/plan.md
tedim52: agree w u, i was following the pattern that `plan.exec` has and think its better if all instructions follow the same pattern

PR #2658, opened 2025-02-25
fix: stdout for file inspect

### LINKED ISSUE #2657: `kurtosis files inspect` doesn't work since `1.5.0`
### What's your CLI version?

1.5.0

In 1.4.4, `kurtosis files inspect` used to work properly even though we had to use a quick hack, `tail -n +2` to be able to parse JSON files with `jq`:

Image

However, since 1.5.0, it looks like `kurtosis files inspect` does not return anything...

Image

### Desired behavior

`kurtosis files inspect` should return the content of the file to stdout and, if possible, without appending anything at the top like `File contents: ...`.

christophercampbell:
## Description
This PR fixes [issue-26657](https://github.com/kurtosis-tech/kurtosis/issues/2657) 

YES

@praetoriansentry @leovct


praetoriansentry (approved):
Looks good to me.

leovct (approved):
lgtm

PR #2661, opened 2025-02-26
chore: remove old docs links

tedim52:
https://docs.kurtosis.com/sdk no longer exists


api/golang/core/lib/enclaves/enclave_context.go
```diff
@@ -373,7 +363,6 @@ func (enclaveCtx *EnclaveContext) GetServiceContext(serviceIdentifier string) (*
 	return serviceContext, nil
 }
 
-// Docs available at https://docs.kurtosis.com/sdk#getservicecontexts--mapstring--bool---servicecontext-servicecontext
 func (enclaveCtx *EnclaveContext) GetServiceContexts(serviceIdentifiers map[string]bool) (map[services.ServiceName]*services.ServiceContext, error) {
 	getServiceInfoArgs := binding_constructors.NewGetServicesArgs(serviceIdentifiers)
 	response, err := enclaveCtx.client.GetServices(context.Background(), getServiceInfoArgs)
```
samlaf, line 368: Would be good to change these functions to accept a context instead of using `context.Background()` silently.
Can also create other methods like `GetServiceContextsWithContext` that take a context. geth does this in its libraries afair.

samlaf (approved):
Thanks for this cleanup. So is there no sdk docs anymore? Why not?

api/golang/core/lib/enclaves/enclave_context.go
tedim52: good point, ill do this in a separate change

tedim52:
@samlaf They are here: https://docs.kurtosis.com/engine-apic-reference. Do you find it useful to have them in code comments? I think it may be easier to link to them from the golang api readme.

samlaf:
> @samlaf They are here: https://docs.kurtosis.com/engine-apic-reference. Do you find it useful to have them in code comments? I think it may be easier to link to them from the golang api readme.

Ah nice thanks. Totally agree with you, linking to documentation which is auto-generated from the code, from the code, makes no sense lol. The other way around (linking to source code from the docs) would make more sense if that's easy to do.

samlaf:
@tedim52 we need a reviewer in here lol

praetoriansentry (approved):

PR #2663, opened 2025-02-28
feat: add resources to logs collector daemonset for k8s api server access 

tedim52:
## Description
The vector logs aggregator needs labels such as `enclave_uuid` , `service_uuid`, etc. to write logs out to the proper filepaths. To enrich the log stream with these labels, the fluentbit logs collector uses the `kubernetes` filter: https://docs.fluentbit.io/manual/pipeline/filters/kubernetes. The filter works by having the logs collector query the k8s api server for necessary information (pods labels in this case) about each pod. To query the api server, the logs collector needs perms. 

This change adds those perms in the form of a `ServiceAccount` given a `ClusterRole` with acess via a `ClusterRoleBinding`. 

NO

## References
https://github.com/kurtosis-tech/kurtosis/pull/2653


container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/consts.go
```diff
@@ -47,6 +47,15 @@ const (
     DB.sync           normal
     Read_from_Head    true
 
+[FILTER]
```
tedim52, line 54: This filter is what requires k8s api server access - https://docs.fluentbit.io/manual/pipeline/filters/kubernetes

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_collector_functions/logs_collector_labels.go
```diff
@@ -4,20 +4,17 @@ import (
 	"github.com/kurtosis-tech/kurtosis/container-engine-lib/lib/backend_impls/docker/object_attributes_provider/docker_labels_for_logs"
 )
 
-// These are the list of container's labels that the Docker's logging driver (currently, the Fluentd logging driver)
-// will add into the logs stream when it sends them to the destination (currently, Fluentbit logs collector)
-type LogsCollectorLabels []string
+// GetServiceLabelsForLogsTracking returns list of labels to add to kurtosis user service containers
```
mieubrisse, line 7: Name doesn't match function name

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/destroy_logs_collector.go
```diff
@@ -37,9 +37,31 @@ func DestroyLogsCollector(ctx context.Context, kubernetesManager *kubernetes_man
 		}
 	}
 
+	if logsCollectorResources.serviceAccount != nil {
+		if err := kubernetesManager.RemoveServiceAccount(ctx, logsCollectorResources.serviceAccount); err != nil {
+			destroyErr = stacktrace.Propagate(err, "An error occurred removing logs collector service account.")
```
mieubrisse: If I'm understanding correctly, if multiple errors occur then doesn't it mean only the last one will be captured? Since `destroyErr` will get overwritten? Seems like it should be capturing all the errors, and then joining them at the end, no?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/consts.go
```diff
@@ -47,6 +47,15 @@ const (
...
+    Name              kubernetes
+    Match             *
+    Kube_URL          https://kubernetes.default.svc:443
```
mieubrisse: Just verifying, this will Always Work right? No matter if I'm running Minikube locally, or a Kubernetes server remotely?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
```diff
@@ -78,9 +82,75 @@ func (fluentbit *fluentbitLogsCollector) CreateAndStart(
...
+		}
+	}
+	shouldRemoveLogsCollectorServiceAccount := true
+	defer func() {
+		if shouldRemoveLogsCollectorServiceAccount {
+			removeServiceAccountFunc()
+		}
+	}()
+
+	clusterRole, err := createLogsCollectorClusterRole(ctx, logsCollectorAttrProvider, kubernetesManager)
```
mieubrisse, line 107: Could be wrong, but I have a hazy memory of ClusterRoles not being easy to simply create. I.e., they have to be created by the k8s admin, since they're cluster wide. Kurtosis creating and removing ClusterRoles could (??) potentially be something that won't work in most clusters.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
```diff
@@ -233,6 +310,7 @@ func createLogsCollectorDaemonSet(
 			Stdin:                    false,
 			StdinOnce:                false,
 			TTY:                      false,
+			TerminationMessagePath:   "",
```
mieubrisse, line 313: What dis?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/logs_collector_daemonset.go
```diff
@@ -19,7 +20,15 @@ type LogsCollectorDaemonSet interface {
...
-	) (*appsv1.DaemonSet, *apiv1.ConfigMap, *apiv1.Namespace, func(), error)
+	) (
+		*appsv1.DaemonSet,
+		*apiv1.ConfigMap,
+		*apiv1.Namespace,
+		*apiv1.ServiceAccount,
+		*rbacv1.ClusterRole,
+		*rbacv1.ClusterRoleBinding,
+		func(),
+		error)
```
mieubrisse: Suggestion: if you put the `)` on a newline below (so:
```
    func(),
    error,
)
```
It becomes visually easier to see that the open-paren above is closed

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -13,12 +13,13 @@ import (
...
 	v1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"net"
 	"time"
 )
 
 const (
-	maxAvailabilityCheckRetries     = 30
+	maxRetries                      = 30
```
mieubrisse: My spidey sense is going off, saying that this _might_ be [over-DRY'ing](https://github.com/kurtosis-tech/style-guide?tab=readme-ov-file#dont-over-dry-your-code-and-create-connections-between-components-when-none-are-needed) code. I.e. that you have at least two separate semantic values (number of retries on waiting for a namespace to get removed, number of retries on waiting for availability), that are now incorrectly getting squashed into one value. 

However, I'm not positive on this - just suggesting it for consideration

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -240,11 +332,32 @@ func waitForLogsCollectorAvailability(
 			pod.Name,
 			fluentBitContainerName,
 			httpPortSpec,
-			maxAvailabilityCheckRetries,
+			maxRetries,
```
mieubrisse: Same comment as above. Here it's a wait for availaibility check...

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -240,11 +332,32 @@ func waitForLogsCollectorAvailability(
...
 
 	return nil
 }
+
+func waitForNamespaceRemoval(
+	ctx context.Context,
+	namespace string,
+	kubernetesManager *kubernetes_manager.KubernetesManager) error {
+
+	for i := uint(0); i < maxRetries; i++ {
```
mieubrisse: ...while here it's a "number of times to wait for namespace removal"

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
```diff
@@ -240,11 +332,32 @@ func waitForLogsCollectorAvailability(
...
+		_, err := kubernetesManager.GetNamespace(ctx, namespace)
```
mieubrisse: Could condense to
```
		if _, err := kubernetesManager.GetNamespace(ctx, namespace); err != nil {
			// if err was returned, namespace doesn't exist, or it's been marked for deleted
			return nil
		}
```

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_manager/kubernetes_manager.go
```diff
@@ -546,6 +546,10 @@ func (manager *KubernetesManager) RemoveNamespace(ctx context.Context, namespace
 	return nil
 }
 
+// GetNamespace returns the namespace object associated with [name] or returns err
```
mieubrisse, line 549: Love me some comments!

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/enclave_object_attributes_provider.go
```diff
@@ -354,7 +356,14 @@ func (provider *kubernetesEnclaveObjectAttributesProviderImpl) getLabelsForEncla
 	if err != nil {
 		return nil, stacktrace.Propagate(err, "An error occurred creating a Kubernetes label value from UUID string '%v'", uuid)
 	}
+	shortUuidStr := uuid_generator.ShortenedUUIDString(uuid)
+	shortUuidLabelValue, err := kubernetes_label_value.CreateNewKubernetesLabelValue(shortUuidStr)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred creating a short GUID Kubernetes label value from GUID string '%v'", uuid)
+	}
 	labels[kubernetes_label_key.GUIDKubernetesLabelKey] = uuidLabelValue
+	labels[kubernetes_label_key.LogsServiceUUIDKubernetesLabelKey] = uuidLabelValue
```
mieubrisse: Cool, I like that you're storing both using the same logic here that goes long -> short (rather than trying to shorten on the consumer side, which would mean there's shortening logic drift between generation & consumption)

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/enclave_object_attributes_provider.go
```diff
@@ -368,6 +377,7 @@ func (provider *kubernetesEnclaveObjectAttributesProviderImpl) getLabelsForEncla
 		return nil, stacktrace.Propagate(err, "An error occurred creating a Kubernetes label value from ID string '%v'", id)
 	}
 	labels[kubernetes_label_key.IDKubernetesLabelKey] = idLabelValue
+	labels[kubernetes_label_key.LogsServiceNameKubernetesLabelKey] = idLabelValue
```
mieubrisse: Something seems wrong here... the function is `EnclaveObject` (so we don't know what type of object it is), but we're attaching a `ServiceName` label (assuming it's a service without knowing)

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/kubernetes_label_key/label_key_consts.go
```diff
@@ -28,6 +28,14 @@ const (
 
 	// As of 2022-05-17, these get attached to files artifact expansion volumes
 	userServiceGuidKeyStr = labelKeyPrefixStr + "user-service-guid"
+
+	// We create a duplicate of the enclave uuid and service uuid label key because:
+	// the logs aggregator (vector) needs the enclave uuid and service uuid label keys to create the filepath where logs are stored in persistent volume
+	// but vectors template syntax can't interpret the "kurtosistech.com/" prefix, so we can't use the existing label keys
+	logsEnclaveUuidLabelKeyStr             = "enclave_uuid"
```
mieubrisse: 1. The other labels are using kebab-case but these are using underline_case. Is this intentional?
2. So are these label key strings are quite different, in that they'll literally *just* be `enclave_uuid` without any namespacing (no `kurtosistech.com/`)? If so, it seems worth making them even more distinct - like `logsOnlyEnclaveUuidLabelKeyStr`, making the comment much bigger, and probably adding something like `kurtosis_` in the value name so it doesn't collide with other stuff in the user's cluster

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/kubernetes_label_key/label_key_consts.go
```diff
@@ -28,6 +28,14 @@ const (
 
 	// As of 2022-05-17, these get attached to files artifact expansion volumes
 	userServiceGuidKeyStr = labelKeyPrefixStr + "user-service-guid"
+
+	// We create a duplicate of the enclave uuid and service uuid label key because:
+	// the logs aggregator (vector) needs the enclave uuid and service uuid label keys to create the filepath where logs are stored in persistent volume
+	// but vectors template syntax can't interpret the "kurtosistech.com/" prefix, so we can't use the existing label keys
```
mieubrisse: Might be worth double-checking - the `kurtosistech.com/` domain prefixing is, to my knowledge, standard Kubernetes practice to ensure you don't have label collisions in your cluster

mieubrisse (approved):
Approved, but have some doubts about some things

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/destroy_logs_collector.go
tedim52: i updated to add all errs to a list and then concatenate and return the final err msg

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
tedim52: ClusterRole is being created fine - likely bc this code is called by the CLI and not the engine container. In the same way, the CLI creates ClusterRole, ClusterRoleBinding, and ServiceAccounts for the engine container.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
tedim52: field i had to add to resolve exhaust struct lint err

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/shared_helpers.go
tedim52: separated these values

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/kubernetes_label_key/label_key_consts.go
tedim52: unfortunately vector template syntax will interpret the `.` as nested key. I do think this is a valid concern though but think its okay to wait  to see if users have friction with this or not. as of right now, users usually use kurtosis standalone in a k8s cluster.

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/kubernetes_label_key/label_key_consts.go
tedim52: 1. underline case is intentional, asthese labels are for reading by vector and vectors template syntax example use underline case: https://vector.dev/docs/reference/configuration/template-syntax/, unsure if its a hard requirement has opposed to kebab case though, but docker backend is already using so wanted to be consistent
2. done, added kurtosis_ prefix, changed variable name of these labels, and emphasized comment

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/enclave_object_attributes_provider.go
tedim52: veryy good point, these labels are being attached at the wrong level. I've changed the logic to add these labels only for the user service enclave objects in `ForUserServicePod`

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/consts.go
tedim52: I think its a fine assumption for right now as its the default route to api server in k8s DNS system for single k8s cluster setups but i'll make it a variable and add a comment making this assumption explicit in case it does need to be modified or pulled from the users k8s config.

PR #2668, opened 2025-03-03
feat: export service logs (Docker)

brianle1301:
## Description
The following PR allows users to define custom Vector sinks for the Docker backend.
It also updates vector to v0.45.0 and adds health check to the logs aggregator.

Sink configurations will reside under the `logs-aggregator` key under each cluster config. They are passed almost as-is to Vector, with the exception of the `inputs` field, which we set as `fluent_bit`. Validation of such configurations are also carried out by Vector.

An example configuration to allow logs export to AWS Opensearch is as follows:

```
config-version: 3
should-send-metrics: true
kurtosis-clusters:
  docker:
    type: "docker"
    logs-aggregator: 
      sinks: 
        elasticsearch:
          type: "elasticsearch"
          opensearch_service_type: "serverless"
          bulk: 
            index: "kt-{{ enclave_uuid }}-{{ service_uuid }}"
          aws:
            region: "eu-west-1"
          auth:
            strategy: "aws"
            access_key_id: "<access_key_id>"
            secret_access_key: "<secret_access_key>"
          endpoints:
            - "<opensearch_endpoint>"
```

YES


container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -655,10 +655,19 @@ func (manager *DockerManager) CreateAndStartContainer(
 		}
 	}()
 
-	//Check if the container dies because sometimes users starts containers with a wrong configuration and these quickly dies
-	didContainerStartSuccessfully, err := manager.didContainerStartSuccessfully(ctx, containerId, dockerImage)
-	if err != nil {
-		return "", nil, stacktrace.Propagate(err, "An error occurred checking if container '%v' is running", containerId)
+	// The immediate start check here can be cumbersome when you need access to the containerId even if the container has failed to start.
+	// For example, we might want to perform further inspection or deletion of the container for short-lived containers.
```
tedim52: If the only use case for skipping successful start is debugging, I'd prefer we not merge this code into main, ideally the way containers are configured should result in correct configuration and start. What use cases were you thinking for this besides debuggin?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/vector_container_config_provider.go
```diff
@@ -2,22 +2,88 @@ package vector
...
-func newVectorContainerConfigProvider(config *VectorConfig) *vectorContainerConfigProvider {
-	return &vectorContainerConfigProvider{config: config}
+func (vector *vectorContainerConfigProvider) GetInitContainerArgs(
+	containerName string,
+	containerLabels map[string]string,
+	networkId string,
+) (*docker_manager.CreateAndStartContainerArgs, error) {
+	logsAggregatorConfigContentStr, err := vector.config.getConfigFileContent()
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred getting the Loki server's configuration content")
```
tedim52: I think this error statement needs to be updated/ Maybe something along the lines of "An error occurred getting logs aggregator config file content."

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/vector_container_config_provider.go
```diff
@@ -2,22 +2,88 @@ package vector
...
+		httpPortNumber: httpPortNumber,
+	}
+}
+
+func (vector *vectorContainerConfigProvider) GetPrivateHttpPortSpec() (*port_spec.PortSpec, error) {
+	privateHttpPortSpec, err := port_spec.NewPortSpec(vector.httpPortNumber, httpTransportProtocol, httpProtocolStr, nil, "")
+	if err != nil {
+		return nil, stacktrace.Propagate(
+			err,
+			"An error occurred creating the Fluentbit server's private HTTP port spec object using number '%v' and protocol '%v'",
```
tedim52: `s/Fluentbit/Vector/g`

container-engine-lib/lib/backend_interface/objects/logs_aggregator/components.go
```diff
@@ -0,0 +1,7 @@
+package logs_aggregator
```
tedim52: any reason to create a new golang file for these as opposed to just putting them in `logs_aggregator.go`?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/vector_logs_aggregator_container.go
```diff
@@ -17,13 +23,68 @@ func NewVectorLogsAggregatorContainer() *vectorLogsAggregatorContainer {
...
+	httpPortNumber uint16,
+	logsAggregatorHttpPortId string,
 	targetNetworkId string,
 	objAttrsProvider object_attributes_provider.DockerObjectAttributesProvider,
 	dockerManager *docker_manager.DockerManager,
 ) (string, map[string]string, func(), error) {
-	vectorContainerConfigProviderObj := createVectorContainerConfigProvider(logsListeningPortNumber)
+	vectorContainerConfigProviderObj := createVectorContainerConfigProvider(logsListeningPortNumber, httpPortNumber, sinks)
+
+	logsAggregatorInitAttrs, err := objAttrsProvider.ForLogsAggregatorInit()
```
tedim52: nice job on adding this code to use a container to dynamically create the configuration file on the logs aggregator container! 2 points:

1. can we encapsulate it into a function `CreateConfiguration` using the same pattern thats used for how the logs collector does this: https://github.com/kurtosis-tech/kurtosis/blob/main/container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_configuration_creator.go

2. we usually reserve object attributes providers for containers/services that kurtosis manages the full lifecycle of. having a system for providing uniform attributes for these containers/services makes it easier to add necessary metadata to retrieve/remove those containers. however bc this init containers is a very ephemeral container that is always created and destroyed immediately, its actually less important they have attributes (eg. see how the fluentbit collector init container works) so I'd suggest actually not creating `objAttrsProvider` here and just assigning name, labels, etc. directly. again see fluentbit config creator [here](https://github.com/kurtosis-tech/kurtosis/blob/main/container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_configuration_creator.go)

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
brianle1301: @tedim52 It's actually not for debugging. 

To validate vector sink configurations, I spun up a separate short lived container to run `vector validate` (before running the main vector container). Now, the amount of time it takes for this container to fail (in the case of invalid configuration) varies. If it fails instantly, the `didContainerStartSuccessfully` check would catch it, and therefore I would not be able to get the containerID to show Vector validation logs and to clean up. On the other hand, it if doesn't fail instantly (but still fast enough, which happens when Vector health checks sinks), the check would get bypassed which allowed me to inspect its logs and remove it properly.  I introduced this argument to disable the check entirely, and rely on WaitForExit() to know if the validation failed or not.

Ideally, we would want to return the containerID even if the container failed to start, or perhaps even introduce a new function that runs short lived containers until completion.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/vector_container_config_provider.go
brianle1301: Yes, will do, good catch

container-engine-lib/lib/backend_interface/objects/logs_aggregator/components.go
brianle1301: Wanted to future proof it because we might want to later add other vector components such as Transforms and Sinks, but can move it to the same file

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/vector_container_config_provider.go
brianle1301: Yes, bad copy pasting, good eye

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/vector_logs_aggregator_container.go
brianle1301: Was thinking the same as point 2, but wanted to be consistent. Will fix these

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: ah yes I did notice that - in this case it makes sense although as I commented below, following the same approach as the fluentbit configuration creator removes the need for this.

docs/docs/guides/exporting-logs.md
```diff
@@ -0,0 +1,206 @@
...
+sidebar_position: 15
+---
+
+This guide assumes that you have [Kurtosis installed](../get-started/installing-the-cli.md).
+
+Kurtosis comes with a logs aggregator component that aggregates logs from services across enclaves. This component uses
+Vector under the hood. Logs aggregator can be configured independently for each cluster, in the Kurtosis [config file](../cli-reference/index.md#configuration-file-path).
+
+Kurtosis uses the notion of "sinks" to describe the location where you want your logs exported. Sink configurations are
+forwarded as-is (with some exceptions, see below) to Vector, therefore Kurtosis can export to all sinks that Vector supports.
```
tedim52: maybe link to Vector here and mention that Kurtosis allows configuring export logs via vector sinks?

docs/docs/guides/exporting-logs.md
```diff
@@ -0,0 +1,206 @@
...
+## Verify logs delivery
+
+To verify that Kurtosis are actually exporting logs to Elasticsearch, start by running a package. In this guide, we use
+`ethpandaops/ethereum-package`:
+
+```bash
+kurtosis run github.com/ethpandaops/ethereum-package
+```
+
+Once the package has finished execution, go to  `http://localhost:5601/app/enterprise_search/content/search_indices`.
```
tedim52, line 127: optional but would be great to have a picture or visual of what users might expect both as a "heres what it should look like" and a "hooray, it works" to wrap up the guide.

brianle1301: Yes can do

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/github_auth_storage_creator/github_auth_storage_creator.go
```diff
@@ -147,7 +147,7 @@ func storeTokenInVolume(
 		}
 
 		// Tiny optimization to not sleep if we're not going to run the loop again
-		if i < maxRetries {
+		if i < maxRetries-1 {
```
tedim52, line 150: nice : )

container-engine-lib/lib/backend_impls/docker/docker_manager/create_and_start_container_args.go
```diff
@@ -1,9 +1,10 @@
 package docker_manager
 
 import (
+	"net"
```
tedim52, line 4: I don't think this `net` is needed for file anymore no?

tedim52 (approved):

container-engine-lib/lib/backend_impls/docker/docker_manager/create_and_start_container_args.go
brianle1301: Still needed for `net.IP` I believe

Issue #2623, opened 2025-01-03
`kurtosis` ignores k8s configs from `KUBECONFIG` env var


0x416e746f6e:
### What's your CLI version?

1.4.3

### Description & steps to reproduce

If a cluster config is stored in a file referenced by [`KUBECONFIG`](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#the-kubeconfig-environment-variable) env var, then `kurtosis` does not "see" it (while `kubectl` does, for example).

### Desired behavior

Use effective configuration by merging all files refernced by `KUBECONFIG`.

### What is the severity of this bug?

Painful; this is causing significant friction in my workflow.

### What area of the product does this pertain to?

CLI: the Command Line Interface

PR #2672, opened 2025-03-10
fix(backend/kubernetes): use default KUBECONFIG resolution

edobry:
## Description

This PR modifies the Kubernetes backend and the CLI to use the K8s SDK's default resolution strategy, allowing them to read the user's K8s configuration from the `KUBECONFIG` environment variable, rather than being hardcoded to only read from `$HOME/.kube/config`. This enables compatibility with a wider variety of local K8s configuration strategies by leveraging standard K8s practices.

This change was primarily motivated by difficulties using Kurtosis with my own setup, wherein I have some scripting which dynamically constructs a `KUBECONFIG` envvar in order to work with multiple K8s clusters. It should not cause any issues, as the SDK's default resolution strategy ultimately falls back to reading from `$HOME/.kube/config`.

There is some duplication in this PR as I couldn't figure out how to re-use the `kubernetesConfig`/`kubernetesManager` created in the K8s backend in the `gateway` module. Happy to take guidance on how to improve this if necessary.

I've tested this PR locally and was successfully able to run successfully.

YES

@tedim52 (pls LMK if there's a better person to tag)


tedim52:
this is a great change, will take a look

tedim52:
Addresses: https://github.com/kurtosis-tech/kurtosis/issues/2623

tedim52 (approved):

cli/cli/commands/gateway/gateway.go
```diff
@@ -41,12 +35,13 @@ func run(cmd *cobra.Command, args []string) error {
...
 
-	// TODO Store kube config path in configuration and read from there
-	kubeConfigPath := path.Join(os.Getenv("HOME"), ".kube", "config")
-
-	kubernetesConfig, err := clientcmd.BuildConfigFromFlags(emptyConfigMasterUrl, kubeConfigPath)
+	// TODO: get this from the backend's k8s manager somehow?
+	kubernetesConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		clientcmd.NewDefaultClientConfigLoadingRules(), 
+		&clientcmd.ConfigOverrides{},
+	).ClientConfig()
```
tedim52, line 41: we have a lint rule called exhaust struct that requires filling out every field in a struct - if you add these with nils/empties CI should build!

tedim52:
tested on my end, works nicely! i think the duplication at the gateway level is okay

Issue #2276, opened 2024-03-11
Configurable Logging


corduroybera:
### Background & motivation

I'm attempting to run a grafana, loki, promtail stack inside of my kurtosis enclave so I can have a more tailored logging environment setup for debugging my services. However, there is no clean way to enable promtail to pull logs while running inside a docker container in the enclave.

### Desired behaviour

The most straightforward way to pipe docker logs into promtail involves mounting the docker socket to the promtail container (or some variation of this to allow for persistent logging).

However, any options to make logging more configurable could help enable this sort of feature more easily

### How important is this to you?

Painful; the lack of this feature makes using Kurtosis frictionful.

### What area of the product does this pertain to?

Other: anything not covered by the above

edobry:
I'm also working on this right now, have you managed to get this working? Is there any canonical way of ingesting logs in both Docker/K8s backends?

tedim52:
@corduroybera @edobry 

This PR adds a grafloki command to start Grafana Loki instances where logs from enclaves in services will be sent. 

On the next release, Kurtosis will also support exporting logs from your enclave to custom logging infra and setups(already supported for Docker on 1.6.0) - see this guide: https://docs.kurtosis.com/exporting-logs/

Note: the logging setup is external to the enclave/kurtosis - we believe this should be the preferred way to aggregate logs from all your enclaves (as opposed to within an enclave).

PR #2696, opened 2025-04-03
feat: grafloki start/stop

tedim52:
## Description
Implements `kurtosis grafloki start` and `kurtosis grafloki stop`. These start and stop grafana and loki instances over Docker and K&s depending on the cluster type. The engine will then be restarted and configured to send enclave logs to the instances.

YES

## References 
https://github.com/kurtosis-tech/kurtosis/issues/2276


cli/cli/command_framework/highlevel/engine_consuming_kurtosis_command/engine_consuming_kurtosis_command.go
```diff
@@ -187,7 +187,7 @@ func (cmd *EngineConsumingKurtosisCommand) getSetupFunc() func(context.Context)
 		kurtosisBackend := engineManager.GetKurtosisBackend()
 
 		dontRestartAPIContainers := false
-		engineClient, closeClientFunc, err := engineManager.StartEngineIdempotentlyWithDefaultVersion(ctx, defaults.DefaultEngineLogLevel, defaults.DefaultEngineEnclavePoolSize, defaults.DefaultGitHubAuthTokenOverride, dontRestartAPIContainers, defaults.DefaultDomain, defaults.DefaultLogRetentionPeriod)
+		engineClient, closeClientFunc, err := engineManager.StartEngineIdempotentlyWithDefaultVersion(ctx, defaults.DefaultEngineLogLevel, defaults.DefaultEngineEnclavePoolSize, defaults.DefaultGitHubAuthTokenOverride, dontRestartAPIContainers, defaults.DefaultDomain, defaults.DefaultLogRetentionPeriod, defaults.DefaultSinks)
```
mieubrisse: wonder if it's worth calling these DefaultLogSinks?

cli/cli/commands/cluster/set/set.go
```diff
@@ -124,7 +124,7 @@ func run(ctx context.Context, flags *flags.ParsedFlags, args *args.ParsedArgs) e
 	// TODO - fix the idempotent starter longer term
 	if engineStatus == engine_manager.EngineStatus_Stopped {
 		dontRestartAPIContainers := false
-		_, engineClientCloseFunc, err := engineManagerNewCluster.StartEngineIdempotentlyWithDefaultVersion(ctx, defaults.DefaultEngineLogLevel, defaults.DefaultEngineEnclavePoolSize, defaults.DefaultGitHubAuthTokenOverride, dontRestartAPIContainers, defaults.DefaultDomain, defaults.DefaultLogRetentionPeriod)
+		_, engineClientCloseFunc, err := engineManagerNewCluster.StartEngineIdempotentlyWithDefaultVersion(ctx, defaults.DefaultEngineLogLevel, defaults.DefaultEngineEnclavePoolSize, defaults.DefaultGitHubAuthTokenOverride, dontRestartAPIContainers, defaults.DefaultDomain, defaults.DefaultLogRetentionPeriod, defaults.DefaultSinks)
```
mieubrisse: Same thing about DefaultLogSinks

cli/cli/commands/grafloki/grafloki.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/kurtosis-tech/kurtosis/cli/cli/commands/grafloki/start"
+	"github.com/kurtosis-tech/kurtosis/cli/cli/commands/grafloki/stop"
+	"github.com/spf13/cobra"
+)
+
+// GraflokiCmd suppressing exhaustruct requirement because this struct has ~40 properties
+// nolint: exhaustruct
+var GraflokiCmd = &cobra.Command{
+	Use:   command_str_consts.GraflokiCmdStr,
+	Short: "Start Grafana/Loki command for log collection",
```
mieubrisse: is "command" a typo? Should that be "container"?

cli/cli/commands/grafloki/start/start.go
```diff
@@ -0,0 +1,120 @@
...
+
+const (
+	defaultEngineVersion                   = ""
+	restartEngineOnSameVersionIfAnyRunning = true
+)
+
+var GraflokiStartCmd = &lowlevel.LowlevelKurtosisCommand{
+	CommandStr:               command_str_consts.GraflokiStartCmdStr,
+	ShortDescription:         "Starts a Grafana/Loki instance.",
+	LongDescription:          "Starts a Grafana/Loki instance and configures Kurtosis engine to send logs to it.",
```
mieubrisse, line 29: I was under the impression that users were only going to bring their own Grafana/Loki and Kurtosis would send to the third-party; sounds like Kurtosis can start an in-house logger as well?

cli/cli/commands/grafloki/start/start.go
```diff
@@ -0,0 +1,120 @@
...
+	_ *flags.ParsedFlags,
+	_ *args.ParsedArgs,
+
+) error {
+	clusterConfig, err := kurtosis_config_getter.GetKurtosisClusterConfig()
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred getting Kurtosis cluster config.")
+	}
+
+	// NOTE(tedi  04/03/25): If you're wondering why the grafana / loki instance is being started by the CLI (and not in container-engine-lib via KurtosisBackend as with LogsCollector and LogsAggregator), here's why:
```
mieubrisse, line 48: Love the explanatory power!

cli/cli/commands/grafloki/start/start.go
```diff
@@ -0,0 +1,120 @@
...
+	// 1. now that Kurtosis is purely OSS, it's important to reduce maintenance surface / complexity inside Kurtosis core. wherever possible infra not essential to Kurtosis core should be built outside of it or at the edges (e.g. client)
```
mieubrisse: isn't the CLI just another part of Kurtosis Core, given that we can't start an engine without it? like, is there really any difference between it being in the CLI vs in core?

cli/cli/commands/grafloki/start/start.go
```diff
@@ -0,0 +1,120 @@
...
+	case resolved_config.KurtosisClusterType_Kubernetes:
+		lokiHost, grafanaUrl, err = grafloki.StartGrafLokiInKubernetes(ctx)
+		if err != nil {
+			return stacktrace.Propagate(err, "An error occurred starting Grafana and Loki in Kubernetes.")
+		}
+	default:
+		return stacktrace.NewError("Unsupported cluster type: %v", clusterConfig.GetClusterType().String())
+	}
+
+	lokiSink := map[string]map[string]interface{}{
```
mieubrisse, line 75: Maybe worth a comment saying where this comes from, in case anyone needs to update

cli/cli/commands/grafloki/start/start.go
```diff
@@ -0,0 +1,120 @@
...
+				"codec": "json",
+			},
+			"labels": map[string]string{
+				"job": "kurtosis",
+			},
+		},
+	}
+
+	logrus.Infof("Configuring engine to send logs to Loki...")
+	err = RestartEngineWithLogsSink(ctx, lokiSink)
```
mieubrisse: Okay maybe I'm undertanding a bit more, the fact that it's _Loki_ and _Grafana_ is actually what's living in the CLI (?). The engine just cares about "generic log sink", and the impl detail of it being Graf/Loki is kept in the CLI?

cli/cli/commands/grafloki/start/start.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred restarting engine to be configured to send logs to Loki.")
+	}
+
+	out.PrintOutLn(fmt.Sprintf("Grafana running at %v", grafanaUrl))
+	return nil
+}
+
+func RestartEngineWithLogsSink(ctx context.Context, sink logs_aggregator.Sinks) error {
```
mieubrisse: does this need to be public? seems like it could be private

cli/cli/commands/grafloki/start/start.go
```diff
@@ -0,0 +1,120 @@
...
+	engineManager, err := engine_manager.NewEngineManager(ctx)
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred creating an engine manager.")
+	}
+	var engineClientCloseFunc func() error
+	var restartEngineErr error
+	dontRestartAPIContainers := false
+	_, engineClientCloseFunc, restartEngineErr = engineManager.RestartEngineIdempotently(ctx,
```
mieubrisse: nit: I like to reduce the number of states that variables can be in (I think this is classed under cyclomatic complexity) so I generalliy don't pre-initialize with `var` unless I absolutely have to. in this case, I would have used `:=` rather than `=.

This is like 1/10 importance though lol; just dropping it here as a thing to think about

cli/cli/defaults/defaults.go
```diff
@@ -30,5 +31,6 @@ const (
 	DefaultLogRetentionPeriod = "168h"
 )
 
+var DefaultSinks = logs_aggregator.Sinks{}
```
mieubrisse: comment from above about the names

cli/cli/helpers/engine_manager/engine_manager.go
```diff
@@ -184,13 +185,18 @@ func (manager *EngineManager) StartEngineIdempotentlyWithDefaultVersion(
 	restartAPIContainers bool,
 	domain string,
 	logRetentionPeriodStr string,
-) (kurtosis_engine_rpc_api_bindings.EngineServiceClient, func() error, error) {
+	additionalSinks logs_aggregator.Sinks,
+) (
```
mieubrisse, line 189: thank you for making this easier to read 😄

cli/cli/helpers/engine_manager/engine_manager.go
```diff
@@ -515,3 +527,14 @@ func (manager *EngineManager) waitUntilEngineStoppedOrError(ctx context.Context)
 	}
 	return stacktrace.NewError("Engine did not report stopped status, last status reported was '%v'", status)
 }
+
+func combineAdditionalSinksAndConfigSinks(additionalSinks logs_aggregator.Sinks, configSinks logs_aggregator.Sinks) logs_aggregator.Sinks {
+	combinedSinks := logs_aggregator.Sinks{}
+	for sinkId, sink := range configSinks {
+		combinedSinks[sinkId] = sink
+	}
+	for sinkId, sink := range additionalSinks {
```
mieubrisse, line 538: Minor: might be worth thinking about which you want to take precedence in case of a sinkId clash. I'd guess (??) that the config should take precedence over the additional, meaning it should be the second one??? You know this better tho

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+var EmptyDockerClientOpts = []client.Opt{}
+var lokiContainerLabels = map[string]string{
+	LokiContainerName: "true",
+}
+var grafanaContainerLabels = map[string]string{
+	GrafanaContainerName: "true",
+}
+
+func StartGrafLokiInDocker(ctx context.Context) (string, string, error) {
+	dockerManager, err := docker_manager.CreateDockerManager(EmptyDockerClientOpts)
```
mieubrisse, line 38: Okay I get more what you mean that this doesn't exist in the "core". The CLI is basically hacking around by directly calling Docker/k8s, to start this thing.

My gut is saying that the intention is good, but that it might lead to complications because both CLI and Engine are now starting things in Docker, but they don't really know about each other.

And it really is Kurtosis itself starting these containers, so my gut says it sort of doesn't matter whether it's in CLI or Engine, this is still part of Kurtosis.

wdyt?

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return "", "", stacktrace.Propagate(err, "An error occurred creating the docker manager to start grafana and loki.")
+	}
+
+	var lokiHost string
+	doesGrafanaAndLokiExist, lokiHost := existsGrafanaAndLokiContainers(ctx, dockerManager, lokiContainerLabels, grafanaContainerLabels)
```
mieubrisse: perhaps `checkGrafanaAndLokiContainerExistence`, to keep with the "start functions with verbs"? 1/10 importance tho

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+
+	if !doesGrafanaAndLokiExist {
+		logrus.Infof("No running Grafana and Loki containers found. Creating them...")
+		lokiHost, err = createGrafanaAndLokiContainers(ctx, dockerManager)
+		if err != nil {
+			return "", "", stacktrace.Propagate(err, "An error occurred creating Grafana and Loki containers.")
+		}
+	}
+
+	grafanaUrl := fmt.Sprintf("http://localhost:%v", grafanaPort)
```
mieubrisse: So this bit seems perhaps fragile... IIRC Windows don't have the `localhost` -> `127.0.0.1` DNS mapping, which is why we use `127.0.0.1` in all the Kurtosis local IP addresses.

Might be another downside to "reinventing" starting a container in Docker rather than going through the normal Kurtosis machinery (but the tradeoff is that this way is prob easier than modifying core Kurtosis machinery)

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+	return lokiHost, grafanaUrl, nil
+}
+
+func createGrafanaAndLokiContainers(ctx context.Context, dockerManager *docker_manager.DockerManager) (string, error) {
+	lokiNatPort := nat.Port(strconv.Itoa(lokiPort) + "/tcp")
+	grafanaNatPort := nat.Port(strconv.Itoa(grafanaPort) + "/tcp")
+
+	lokiArgs := docker_manager.NewCreateAndStartContainerArgsBuilder(lokiImage, LokiContainerName, bridgeNetworkId).
+		WithUsedPorts(map[nat.Port]docker_manager.PortPublishSpec{
+			lokiNatPort: docker_manager.NewManualPublishingSpec(lokiPort),
```
mieubrisse, line 72: My gut is giving me this yellow flag of reimplementing all this stuff outside of core Kurtosis machinery, rather than this being done in the Kurtosis backend

mieubrisse: Maybe the speed increase is worth not doing it this way though? Guess it'd be tech debt

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+	defer func() {
+		if shouldDestroyLokiContainer {
+			err := dockerManager.RemoveContainer(ctx, lokiContainerId)
+			if err != nil {
+				logrus.Warnf("Attempted to remove Loki container after an error occurred creating it but an error occurred removing it.")
+				logrus.Warnf("Manually remove Loki container with id: %v", lokiContainerId)
+			}
+		}
+	}()
+	if err != nil {
```
mieubrisse: Bug here with the ordering. Err should be checked before deferring the conditional destroy, b/c if an `err` comes back it means no container's been successfully created

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+		return "", stacktrace.Propagate(err, "An error occurred creating '%v' container.", LokiContainerName)
+	}
+	logrus.Infof("Loki container started.")
+
+	lokiContainer, err := getContainerByLabel(ctx, dockerManager, lokiContainerLabels)
```
mieubrisse: getContainerByLabels sounds like something that would already exist in KurtosisBackend and is being reinvented maybe?

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return "", stacktrace.Propagate(err, "An error occurred getting Loki container by labels: %v", lokiContainerLabels)
+	}
+
+	lokiBridgeNetworkIpAddress := fmt.Sprintf("http://%v:%v", lokiContainer.GetDefaultIpAddress(), lokiPort)
+	lokiHostNetworkIpAddress := fmt.Sprintf("http://localhost:%v", lokiPort)
```
mieubrisse: Same thing with `localhost`.

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+	if err := waitForLokiReadiness(lokiHostNetworkIpAddress, lokiReadinessPath); err != nil {
+		return "", stacktrace.Propagate(err, "An error occurred waiting for Loki container to become ready.")
+	}
+
+	datasourceYamlContent := fmt.Sprintf(`apiVersion: 1
```
mieubrisse: Normally the way I'd do this is create an actual struct and YAML serialize it, so that you're not relying on anything like counting indentation for the next person who edits this

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+		WithUsedPorts(map[nat.Port]docker_manager.PortPublishSpec{
+			grafanaNatPort: docker_manager.NewManualPublishingSpec(grafanaPort),
+		}).
+		WithEnvironmentVariables(map[string]string{
+			"GF_SECURITY_ALLOW_EMBEDDING": "true",
+			"GF_AUTH_ANONYMOUS_ENABLED":   "true",
+			"GF_AUTH_ANONYMOUS_ORG_ROLE":  "Admin",
+		}).
+		WithBindMounts(map[string]string{
+			tmpFile.Name(): "/etc/grafana/provisioning/datasources/loki.yaml",
```
mieubrisse: Because this is a temp file, is there a chance that it gets deleted if the user closes their latop and reopens a day later? Thereby making the Loki bind-mounting a nonexistent file?

I'm not sure if the temp file would get deleted if it's still bound-mount or not

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+		WithRestartPolicy(docker_manager.RestartOnFailure).
+		WithNetworkMode(bridgeNetworkId).
+		WithLabels(grafanaContainerLabels).
+		Build()
+	grafanaContainerId, _, err := dockerManager.CreateAndStartContainer(ctx, grafanaArgs)
+	if err != nil {
+		return "", stacktrace.Propagate(err, "An error creating creating '%v' container.", GrafanaContainerName)
+	}
+	shouldDestroyGrafanaContainer := true
+	defer func() {
```
mieubrisse, line 155: This is the correct ordering of err-checking then defer

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+		resp, err := http.Get(url)
+		if err == nil {
+			resp.Body.Close()
+			if resp.StatusCode == http.StatusOK {
+				return nil
+			}
+		}
+		time.Sleep(retryDelay)
+	}
+	return stacktrace.NewError("Loki did not become ready after 30 attempts")
```
mieubrisse: hardcoded value that should come from the actual variable backing it

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+		existsGrafana = true
+	}
+
+	return existsLoki && existsGrafana, fmt.Sprintf("http://%v:%v", lokiBridgeNetworkIpAddress, lokiPort)
+}
+
+func getContainerByLabel(ctx context.Context, dockerManager *docker_manager.DockerManager, containerLabels map[string]string) (*types.Container, error) {
+	containers, err := dockerManager.GetContainersByLabels(ctx, containerLabels, false)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred getting container by labels '%v'.", containerLabels)
```
mieubrisse: I normally use `%+v` to get it to print map key/values

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	if len(containers) == 0 {
+		return nil, stacktrace.NewError("No containers with labels '%v' found.", containerLabels)
```
mieubrisse: My gut says that not finding any containers isn't an error case, but instead returning a `nil` container (i.e. it's expected)

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+		time.Sleep(retryDelay)
+	}
+	return stacktrace.NewError("Loki did not become ready after 30 attempts")
+}
+
+func existsGrafanaAndLokiContainers(ctx context.Context, dockerManager *docker_manager.DockerManager, lokiContainerLabels map[string]string, grafanaContainerLabels map[string]string) (bool, string) {
+	existsLoki := false
+	existsGrafana := false
+	var lokiBridgeNetworkIpAddress string
+	lokiContainer, err := getContainerByLabel(ctx, dockerManager, lokiContainerLabels)
```
mieubrisse, line 195: Should `existsGrafanaAndLokiContainers` have an `err` return value as well? Related, should the `err` value of `getContainerByLabel` be handled separately?

E.g. if 0 or >1 loki container exists, then `getContainerByLabel` will return an err. In both cases, `existsLoki` will be `false`. Which I think will mean that >1 loki containers will now become 1 + >1 loki containers

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+	if err == nil && lokiContainer != nil {
+		existsLoki = true
+		lokiBridgeNetworkIpAddress = lokiContainer.GetDefaultIpAddress()
+	}
+
+	grafanaContainer, err := getContainerByLabel(ctx, dockerManager, grafanaContainerLabels)
+	if err == nil && grafanaContainer != nil {
```
mieubrisse: Same question

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -0,0 +1,217 @@
...
+	if len(containers) == 0 {
+		return nil, stacktrace.NewError("No containers with labels '%v' found.", containerLabels)
+	}
+	if len(containers) > 1 {
+		return nil, stacktrace.NewError("More than one container with labels '%v' found.", containerLabels)
+	}
+	return containers[0], nil
+}
+
+func StopGrafLokiInDocker(ctx context.Context) error {
```
mieubrisse, line 232: Seems like there's a mismatch in how we identify the Grafana/Loki containers: we get them by labels, but we remove them by name. Seems like this could cause subtle bugs in calculating the existence of containers (which is necessary for starting/removal)

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
```diff
@@ -543,17 +543,19 @@ func (manager *DockerManager) CreateAndStartContainer(
...
-		return "", nil, stacktrace.Propagate(err, "An error occurred checking for the existence of network with ID %v", args.networkId)
-	}
-	if len(networks) == 0 {
-		return "", nil, stacktrace.NewError(
-			"Kurtosis Docker network with ID %v was never created before trying to launch containers. Please call DockerManager.CreateNetwork first.",
-			args.networkId,
-		)
-	} else if len(networks) > 1 {
-		return "", nil, stacktrace.NewError("Kurtosis Docker network with ID %v matches several networks!", args.networkId)
+	if args.networkId != dockerNetworkDriver {
```
mieubrisse: Question: is this comparing apples to oranges? The network ID is a semantically different value than the network driver enum, even if they happen to share the same string, no?

mieubrisse: Also, I think I'm confused about the purpose of this check. Presumably it's to only do the below logic if `args.networkId (bridge) != dockerNetworkDriver (bridge)`. But it seems like an odd special casing, and I'd have imagined that we handle situations where we want containers on the bridge network already (e.g. the engine container)

container-engine-lib/lib/backend_impls/docker/docker_manager/types/container.go
```diff
@@ -14,6 +14,7 @@ type Container struct {
 	entrypointArgs   []string
 	cmdArgs          []string
 	envVars          map[string]string
+	defaultIpAddress string
```
mieubrisse, line 17: I'm confused by this. My understanding is that containers exist independent of what networks they're attached to, and IP addresses are keyed on the network they're attached to. So that there's no such thing as a "default IP address".

And actually maybe this is helping clarify my thinking from the above comment: I think the `bridge` network being treated as a special case feels odd, because (from the container's perspective) it's just another network like any other.

My understanding is that we have containers which get attached to various networks (which may or may not be the bridge network), and from the perspective of both the container and our DockerManager we don't care. If we add special casing to the bridge network, then we make one network special above the others.

Maybe that's necessary since I know the Docker API itself is not well-organized; just flagging that it feels like a break in the "all networks are equal" paradigm

mieubrisse: How I would have imagined this would look (and maybe it's impossible because the Docker API _does_ treat the bridge network fields as different from normal network fields):
1. When creating containers, we attach them to whatever network we want (which may or may not include the `bridge` network, depending on our needs. in this case, with the CLI creating Graf/Loki, the `bridge` network is where we want to create them)
2. When the container is created, we query the container for its IP address on the network-of-interest (which happens to `bridge`, but doesn't need to be in the general case)

In this way, `bridge` isn't special in any way from the DockerManager's perspective

core/files_artifacts_expander/main.go
```diff
@@ -111,7 +111,7 @@ func runMain() error {
 
 		// NOTE: We don't use stacktrace here because the actual stacktraces we care about are the ones from the threads!
 		return fmt.Errorf(
-			"The following errors occurred when trying to expand files artifacts:\n%v",
+			"the following errors occurred when trying to expand files artifacts:\n%v",
```
mieubrisse, line 114: Typo, or too-aggressive find/replace perhaps?

docs/docs/cli-reference/grafloki-start.md
```diff
@@ -0,0 +1,18 @@
+---
+title: grafloki start
+sidebar_label: grafloki start
+slug: /grafloki-start
+---
+
+To start a Grafana/Loki instance in Docker or K8s cluster, run:
```
mieubrisse, line 7: Love docs!

docs/docs/cli-reference/grafloki-stop.md
```diff
@@ -0,0 +1,13 @@
+---
+title: grafloki stop
+sidebar_label: grafloki stop
+slug: /grafloki-stop
+---
+
+To stop a Grafana/Loki instance in Docker or K8s cluster, run:
```
mieubrisse: Perhaps useful to link to `grafloki start` just so people can get there easily if they come across this page

docs/docs/cli-reference/grafloki-start.md
```diff
@@ -0,0 +1,18 @@
...
+```bash
+kurtosis grafloki start
+```
+
+This command starts a local Grafana/Loki instance and restarts the Kurtosis engine with an updated configuration. The new configuration includes a log sink that routes logs to the local Grafana/Loki instance.
+
+Read more about sinks and how to [export logs][export-logs] from Kurtosis.
+
+<!-------------------- ONLY LINKS BELOW THIS POINT ----------------------->
+[export-logs]: ../guides/exporting-logs.md
```
mieubrisse, line 18: Do you want to update `exporting-logs.md` with information about the Grafloki stuff?

cli/cli/commands/grafloki/start/start.go
tedim52: as discussed offline this isn't an in house logger as much as it is, us setting up the grafana loki for the user and kurtosis sending to it.

cli/cli/commands/grafloki/start/start.go
tedim52: discussed offline and will clarify comment: 

"Kurtosis Core" - 1. Engine 2. APIContainer 3. KurtosisBackend in container engine lib 4. Starlark Engine

"kurtosis CLI + API" client of kurtosis core (almost - yes its needed to start the engine right now)

cli/cli/commands/grafloki/start/start.go
tedim52: love it

cli/cli/helpers/engine_manager/engine_manager.go
tedim52: I actually consider `additionalSinks` as a type of override from default sinks but maybe I can rename additionalSinks to override sinks here to make that explicit

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: discussed offline - going to proceed with keeping this in the CLI but update the comment to further explain

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: discussed offline - taking the hit of reimplementing some things in favor of pushing this out of Kurtosis core

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: valid

tedim52: done

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: the tmp file gets deleted in a `defer tmpFile.Close()` right before hand it's only needed within the context of this function then gets removed from the users machine immediately

core/files_artifacts_expander/main.go
tedim52: go lint rule

docs/docs/cli-reference/grafloki-start.md
tedim52: good shout

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: each KurtosisBackend entity implements "getContainers..." as needed `dockerManager.GetContainersByLabels` (eg. https://github.com/kurtosis-tech/kurtosis/blob/c00f357b5ebb9cb9e1b52ab412e489d6260895c6/container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/shared_helpers/shared_helpers.go#L523) so doing the same here

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: agreed

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: down to take these hits to keep this out of core as discussed

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+const (
+	grafanaServiceName                   = "kurtosis-grafana-service"
+	lokiServiceName                      = "kurtosis-loki-service"
+	grafanaDeploymentName                = "kurtosis-grafana-deployment"
+	lokiDeploymentName                   = "kurtosis-loki-deployment"
+	grafanaDatasourceConfigMapName       = "kurtosis-grafana-datasources"
+	graflokiNamespace                    = "kurtosis-grafloki"
+	grafanaNodePort                int32 = 30030
+	lokiNodePort                   int32 = 30031
+	lokiProbInitialDelaySeconds          = 5
```
mieubrisse: s/Prob/Probe/

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+	lokiProbePeriodSeconds               = 10
+	lokiProbeTimeoutSeconds              = 10
+
+	// takes around 30 seconds for loki pod to become ready
+	lokiDeploymentMaxRetires    = 40
```
mieubrisse: s/Retires/Retries/

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+	lokiDeploymentRetryInterval = 1 * time.Second
+)
+
+var lokiLabels = map[string]string{
+	lokiDeploymentName: "true",
```
mieubrisse: Huh, this is odd to me. It seems like this is intended to be a pseudo-set, but rather than a `bool` values we have a string representing a bool val?

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+}
+var grafanaLabels = map[string]string{
+	grafanaDeploymentName: "true",
```
mieubrisse: same Q

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+
+func StartGrafLokiInKubernetes(ctx context.Context) (string, string, error) {
+	k8sManager, err := getKubernetesManager()
+	if err != nil {
+		return "", "", stacktrace.Propagate(err, "An error occurred getting Kubernetes Manager.")
+	}
+
+	var lokiHost string
+	doesGrafanaAndLokiExist, lokiHost := checkGrafanaAndLokiDeploymentExistence(ctx, k8sManager)
+	if err != nil {
```
mieubrisse: Sneaky little bug here - you're actually checking _the previous function's_ `err` value, because `checkGrafanaAndLokiDeploy...` doesn't return an `err`

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+		return "", "", stacktrace.Propagate(err, "An error occurred checking if Grafana and Loki exist.")
+	}
+
+	if !doesGrafanaAndLokiExist {
+		lokiHost, err = createGrafanaAndLokiDeployments(ctx, k8sManager)
+		if err != nil {
+			return "", "", stacktrace.Propagate(err, "An error occurred creating Grafana and Loki deployments.")
+		}
```
mieubrisse, line 60: Do you need a conditional-defer-undo after the err check passes?

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+	}
+
+	logrus.Infof("Run `kubectl port-forward -n %v svc/%v %v:%v` to access Grafana service.", graflokiNamespace, grafanaServiceName, grafanaPort, grafanaNodePort)
+	return lokiHost, getGrafanaUrl(grafanaPort), nil
+}
+
+func createGrafanaAndLokiDeployments(ctx context.Context, k8sManager *kubernetes_manager.KubernetesManager) (string, error) {
+	glNamespace, err := k8sManager.CreateNamespace(ctx, graflokiNamespace, map[string]string{}, map[string]string{})
```
mieubrisse: Potential suggestion: I deffff know it's tempting to sometimes abbreviate (e.g. `glNamespace`). I try to avoid this temptation, because I think it's a bad Taleb bet: the upside is you save yourself a couple keystrokes typing the name the first time (because the rest of the time it's autocompleted), but the downside is that that name gets propagated a bunch of places in the codebase (e.g. if the Future Midnight Kurtosian created a function called inside of this one to do something, then they named the args to that function `glNamespace` as well). The variable name isn't able to stand on its own as well as `grafLokiNamespace`, so now future coders and PR reviewers need to do a bit of research to figure out what the `gl` actually means.

Now, all that's the theory. In practice, do I think there's much negative leverage to naming this `glNamespace`? Not really! Just wanted to give you the theory so you can consider it as a model when you're naming things.

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+	shouldRemoveNamespace := false
+	defer func() {
+		if shouldRemoveNamespace {
+			if err := k8sManager.RemoveNamespace(ctx, glNamespace); err != nil {
+				logrus.Warnf("Attempted to remove namespace '%v' after an error occurred creating it but an error occurred removing it.", graflokiNamespace)
+				logrus.Warnf("Manually remove namepsace with name: %v", glNamespace.Name)
```
mieubrisse: In the past we've been pretty loud about the user needing to take manual action (e.g. `!! ACTION REQUIRED !!`); your choice on whether to do that, but just holding it up for consideration

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+			}
+		}
+	}()
+	if err != nil {
```
mieubrisse: Same bug as comments from yesterday - error-checking is happening after the defer, which means the defer might be trying to delete something that never got created

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+			PodAntiAffinity: nil,
+		})
+	if err != nil {
+		return "", stacktrace.Propagate(err, "An error occurred creating Loki deployment.")
+	}
+	shouldRemoveLokiDeployment := false
+	defer func() {
+		if shouldRemoveLokiDeployment {
+			if err := k8sManager.RemoveDeployment(ctx, graflokiNamespace, lokiDeployment); err != nil {
+				logrus.Warnf("Attempted to remove Loki deployment after an error occurred creating it but an error occurred removing it.")
```
mieubrisse: Technically there's not an error creating the deployment, right? It's actually that an unrelated error happened afterwards, and to clean up things we tried to remove the deployment (but then an error happened with _that_)

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+				logrus.Warnf("Manually remove Loki deployment with name: %v", lokiDeployment.Name)
```
mieubrisse: Same comment about how loud you want to be

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+			}
+		}
+	}()
+	logrus.Infof("Waiting for Loki deployment to come online (can take around 30s)... ")
```
mieubrisse, line 178: Sounds like Loki takes around 30s, and we're giving 40s. That seems like a pretty tight timeline, esp since I'm guessing that you're using your (super highpowered) Mac as benchmarking. What happens on a user's computer that's older or slower or more overloaded?

In Clear Thinking Shane Parrish talks about how a good rule-of-thumb for margin of safety is 2x what you think; might be useful waiting at least a minute for Loki to come online

mieubrisse: If there's a fear around too long of a wait in an infini-failed state, I thiiiiink the Kubernetes probes might have ways to detect this and eject early (maybe worth looking into)

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+		[]apiv1.ServicePort{{
+			Name:        "logs-listening",
+			Port:        lokiNodePort,
+			TargetPort:  intstr.FromInt(lokiPort),
+			Protocol:    apiv1.ProtocolTCP,
+			NodePort:    lokiNodePort,
+			AppProtocol: &httpApplicationProtocol,
+		}})
+	shouldRemoveLokiService := false
+	defer func() {
```
mieubrisse, line 208: Same bug again - doing the defer-undo before checking if `err != nil`

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+		if shouldRemoveLokiService {
+			if err := k8sManager.RemoveService(ctx, lokiService); err != nil {
+				logrus.Warnf("Attempted to remove Loki service after an error occurred creating it but an error occurred removing it.")
```
mieubrisse: Same comment about the language in this (if we got here, we actually _succeeded_ in creating the Loki service)

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+				logrus.Warnf("Manually remove Loki service with name: %v", lokiService.Name)
```
mieubrisse: Same comment about loudness (if you choose)

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+			}
+		}
+	}()
+	if err != nil {
+		return "", stacktrace.Propagate(err, "An error occurred creating Loki service")
+	}
+	lokiHost := getLokiHost(lokiServiceName, graflokiNamespace, lokiNodePort)
```
mieubrisse: So generally Kubernetes doesn't want you getting actual host (node) names. You _can_, but the more canonical way is to get the domain name of the service (which lets k8s do any migrations and such that it deems necessary transparently to the consumer)

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+
+	configMapData := map[string]string{
+		"loki-datasource.yaml": fmt.Sprintf(`apiVersion: 1
```
mieubrisse: Same comment about potentially making this a struct and serializing it

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+    type: loki
+    access: proxy
+    url: %v
+    isDefault: true
+    editable: true`, lokiServiceName, lokiHost),
+	}
+	grafanaConfigMap, err := k8sManager.CreateConfigMap(ctx,
+		graflokiNamespace,
+		grafanaDatasourceConfigMapName,
+		map[string]string{}, // empty labels
```
mieubrisse, line 238: useful comments!

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+		map[string]string{}, // empty annotations
+		configMapData)
+	shouldRemoveGrafanaConfigMap := true
```
mieubrisse, line 244: Same bug about defer-undo before err checking

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+	defer func() {
+		if shouldRemoveGrafanaConfigMap {
+			if err := k8sManager.RemoveConfigMap(ctx, graflokiNamespace, grafanaConfigMap); err != nil {
+				logrus.Warnf("Attempted to remove Grafana datasource config map after an error occurred creating it but an error occurred removing it.")
```
mieubrisse: Same comments as above about the language here, the loudness, etc.

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+						Name:          "",
+						ContainerPort: grafanaPort,
+						HostPort:      0,
+						Protocol:      "",
+						HostIP:        "",
+					},
+				},
+				Env: []apiv1.EnvVar{
+					{
+						Name:      "GF_AUTH_ANONYMOUS_ENABLED",
```
mieubrisse: aren't these envvars used in the Docker version as well? should they be named consts rather than hardcoded, so you don't have to change them in two places if you did in the future?

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+				TerminationMessagePolicy: "",
+				ImagePullPolicy:          "",
+				SecurityContext:          nil,
+				Stdin:                    false,
+				StdinOnce:                false,
+				TTY:                      false,
+			},
+		},
+		[]apiv1.Volume{{
+			Name: "datasources",
```
mieubrisse: isn't hardcoding the string `datasources` here and above pretty fragile vs having named variable to link them? DRY and all that

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+				Ephemeral:             nil,
+			},
+		}},
+		&apiv1.Affinity{
+			NodeAffinity:    nil,
+			PodAffinity:     nil,
+			PodAntiAffinity: nil,
+		})
+	shouldRemoveGrafanaDeployment := true
+	defer func() {
```
mieubrisse, line 384: Defer-undo before err-checking, logtext string description, loudness, etc.

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+		[]apiv1.ServicePort{{
+			Name:        "grafana-dashboard",
+			Port:        grafanaNodePort,
+			TargetPort:  intstr.FromInt(grafanaPort),
+			Protocol:    apiv1.ProtocolTCP,
+			NodePort:    grafanaNodePort,
+			AppProtocol: &httpApplicationProtocol,
+		}})
+	shouldRemoveGrafanaService := true
+	defer func() {
```
mieubrisse, line 415: Defer-undo before err-checking, logtext description, loudness of the user action, etc.

Will skip commenting on this further since I think you get the idea lol

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+	shouldRemoveLokiService = false
+	return lokiHost, nil
+}
+
+func checkGrafanaAndLokiDeploymentExistence(ctx context.Context, k8sManager *kubernetes_manager.KubernetesManager) (bool, string) {
+	existsLoki := false
+	existsGrafana := false
+	var lokiHost string
+	lokiDeployment, err := k8sManager.GetDeployment(ctx, graflokiNamespace, lokiDeploymentName)
+	if err == nil && lokiDeployment != nil {
```
mieubrisse, line 445: Same comment as the Docker version - it miiiight be useful distinguishing "an error occurred checking existence" (i.e. "I don't know if it exists") from "it doesn't exist" by having `checkGrafanaAndLoki...` also return an `err`.

If you don't distinguish, you may try creating something that already exists

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+		existsLoki = true
+		lokiHost = getLokiHost(lokiServiceName, graflokiNamespace, lokiNodePort)
+	}
```
mieubrisse, line 450: Since the bool result is an `AND`, you can eject early by simply returning false if `lokiDeployment` doesn't exist. saves you from making the Grafana check if you already know Loki doesn't exist

mieubrisse: Writing this, might be worth thinking what happens if Grafana exists but Loki doesn't. Can the outer code handle this? Or does it get into a perma-borked state?

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+
+	return existsLoki && existsGrafana, lokiHost
+}
+
+func StopGrafLokiInKubernetes(ctx context.Context) error {
+	k8sManager, err := getKubernetesManager()
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred getting Kubernetes Manager.")
+	}
+	ns, err := k8sManager.GetNamespace(ctx, graflokiNamespace)
```
mieubrisse: Same comment about non-common abbreviations (i.e. not API, etc.). Feel free to totally ignore though; I just personally find it makes the code harder to parse

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+	err = waitForNamespaceRemoval(ctx, graflokiNamespace, k8sManager)
+	if err != nil {
+		return stacktrace.Propagate(err, "An error occurred while waiting for graflokiNamespace '%v' removal.", graflokiNamespace)
+	}
+	return nil
+}
+
+func getKubernetesManager() (*kubernetes_manager.KubernetesManager, error) {
+	kubernetesConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		clientcmd.NewDefaultClientConfigLoadingRules(), nil,
```
mieubrisse: what's the `nil`?

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+	).ClientConfig()
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred creating Kubernetes configuration")
+	}
+	clientSet, err := kubernetes.NewForConfig(kubernetesConfig)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "Expected to be able to create Kubernetes client set using Kubernetes config '%+v', instead a non nil error was returned", kubernetesConfig)
+	}
+	k8sManager := kubernetes_manager.NewKubernetesManager(clientSet, kubernetesConfig, "")
```
mieubrisse: What's the emptystring?

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+	return k8sManager, nil
+}
+
+func waitForNamespaceRemoval(
+	ctx context.Context,
+	namespace string,
+	kubernetesManager *kubernetes_manager.KubernetesManager) error {
+	var (
```
mieubrisse, line 502: can this be `const`? remove cyclomatic complexity?

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+		maxTriesToWaitForNamespaceRemoval       uint = 30
+		timeToWaitBetweenNamespaceRemovalChecks      = 1 * time.Second
+	)
+
+	for i := uint(0); i < maxTriesToWaitForNamespaceRemoval; i++ {
+		if _, err := kubernetesManager.GetNamespace(ctx, namespace); err != nil {
+			// if err was returned, graflokiNamespace doesn't exist, or it's been marked for deleted
```
mieubrisse, line 509: What about errs like connection timeout?

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+			return nil
+		}
+
+		// Tiny optimization to not sleep if we're not going to run the loop again
+		if i < maxTriesToWaitForNamespaceRemoval-1 {
+			time.Sleep(timeToWaitBetweenNamespaceRemovalChecks)
+		}
+	}
+
+	return stacktrace.NewError("Attempted to wait for namespace '%v' removal or to be marked for deletion '%v' times but '%v' was not removed.", namespace, namespace, maxTriesToWaitForNamespaceRemoval)
```
mieubrisse: There's a bug in the ordering of the vars printed. This would print:
```
Attempted to wait for namespace 'graflokiNamespace' removal or to be marked for deletion 'graflokiNamespace' times but '30' was not removed.
```

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+}
+
+func getLokiHost(lokiServiceName, namespace string, lokiPort int32) string {
```
mieubrisse: Ah okay, my comment above is irrelevant. From the `host` I thought it was going to be an actual host, e.g. an IP. Maybe worth calling this `getLokiUrl`?

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -0,0 +1,477 @@
...
+	return fmt.Sprintf("http://%v.%v.svc.cluster.local:%v", lokiServiceName, namespace, lokiPort)
+}
+
+func getGrafanaUrl(grafanaPort int) string {
```
mieubrisse: Yellow flag going off in my brain that Loki is referenced by the service domain name scheme, but Grafana is referenced by 127.0.0.1. Should we be referencing these by the same way? Or is there some sort of implicit information (e.g. this is actually `getGrafanaUrlOnUsersMachine` while `getLokiUrl` is actually `getLokiUrlInsideK8sCluster`?)

docs/docs/guides/exporting-logs.md
```diff
@@ -404,4 +404,12 @@ kurtosis-clusters:
             codec: "json"
 ```
 
+### Grafloki
+
+For ease of setup, Kurtosis CLI comes with a feature to start a local Grafana and Loki instance that the engine gets configured to sends logs to. See [grafloki start][grafloki-start] for more info.
```
mieubrisse: s/to sends/to send/

mieubrisse (approved):

container-engine-lib/lib/backend_impls/docker/docker_manager/types/container.go
tedim52: agree with you here, i've removed defaultIPAddress in favor of a function to get a container ip address on a network. also removed the check so bridge network doesn't get treated as special

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_manager.go
tedim52: addressed with the other comment, there's no longer a special check

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: The engine uses the same pattern of getting by label , removing by id/name.

cli/cli/helpers/grafloki/kubernetes_grafloki.go
tedim52: we encourage loki to ship faster, hence the tight timeline

cli/cli/helpers/grafloki/kubernetes_grafloki.go
tedim52: changed these to be more idiomatic

cli/cli/helpers/grafloki/kubernetes_grafloki.go
tedim52: i like putting them here bc its closer to the context they're used in

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: agree with you - i made three changes here:
- rename exists to check
- have getContainerLabel return nil instead of err if num containers is 0
- have check return an err and handle it

cli/cli/helpers/grafloki/kubernetes_grafloki.go
tedim52: def an annoying corner case, but if this does happen, a simple `kurtosis grafloki stop` followed by start should resolve this since the grafana container is removed first

cli/cli/helpers/grafloki/kubernetes_grafloki.go
tedim52: i debug log the err out in case.

odyslam:
Hey everyone, tried building from source on latest `main` and running grafloki, but the Grafana will just restart all the time with:

It built successfully using the top level build script and the CLI output that it's using the correct tag (dirty) of the kutosis engine.

Also a note, since grafana is usually also used with prometheus (e.g ethpandas/optimism) to gather metrics from various services, it would be nice to be able to configure the grafloki grafana to connect to any running prometheus service as part of an enclave (or perhaps move prometheus outside of the enclave all-together and basically observability is handled by kurtosis) such that we don't have 2 grafana instances running: one to gather metrics and one with `grafloki`.

An additional idea and simple idea would be that while fluentd is the default drive, you could also enable the user to set custom driver -- at least for docker, such that one can use `journald` for example.

```
logger=provisioning t=2025-04-09T11:22:46.661951039Z level=error msg="Failed to provision data sources" error="Datasource provisioning error: open /etc/grafana/provisioning/datasources/loki.yaml: permission denied"
logger=provisioning t=2025-04-09T11:22:46.661988409Z level=error msg="Failed to provision data sources" error="Datasource provisioning error: open /etc/grafana/provisioning/datasources/loki.yaml: permission denied"
logger=server t=2025-04-09T11:22:46.662023279Z level=error msg="Stopped background service" service=*provisioning.ProvisioningServiceImpl reason="Datasource provisioning error: open /etc/grafana/provisioning/datasources/loki.yaml: permission denied"
logger=grafana.update.checker t=2025-04-09T11:22:46.662105969Z level=error msg="Update check failed" error="failed to get stable version from grafana.com: Get \"https://grafana.com/api/grafana/versions/stable\": *provisioning.ProvisioningServiceImpl run error: Datasource provisioning error: open /etc/grafana/provisioning/datasources/loki.yaml: permission denied" duration=129.22µs
logger=infra.usagestats t=2025-04-09T11:22:46.662149039Z level=error msg="Failed to get last sent time" error="context canceled"
logger=serviceaccounts t=2025-04-09T11:22:46.662192079Z level=warn msg="Failed to get usage metrics" error="context canceled"
logger=auth t=2025-04-09T11:22:46.66264372Z level=error msg="Failed to lock and execute cleanup of expired auth token" error="context canceled"
logger=http.server t=2025-04-09T11:22:46.665312086Z level=info msg="HTTP Server Listen" address=[::]:3000 protocol=http subUrl= socket=
logger=serviceaccounts t=2025-04-09T11:22:46.677729954Z level=error msg="Failed to lock and execute the migration of API keys to service accounts" error="context canceled"
logger=plugin.signature.key_retriever t=2025-04-09T11:22:46.679031477Z level=error msg="Error downloading plugin manifest keys" error="kv get: context canceled"
Error: ✗ *provisioning.ProvisioningServiceImpl run error: Datasource provisioning error: open /etc/grafana/provisioning/datasources/loki.yaml: permission denied
```

PR #2706, opened 2025-04-09
fix: give grafana root access

tedim52:
## Description
A user reported the grafana container not having access to the bindmounted datasource yaml within the container. This was due to varying permissions from the original tmp file created by the CLI and inherited by the bindmounted file. To get around this, this change gives the grafana container root user which fixed their issue.

YES

### References
https://github.com/kurtosis-tech/kurtosis/pull/2696#issuecomment-2789352402


cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -146,6 +148,7 @@ func createGrafanaAndLokiContainers(ctx context.Context, dockerManager *docker_m
 		WithRestartPolicy(docker_manager.RestartOnFailure).
 		WithNetworkMode(bridgeNetworkName).
 		WithLabels(grafanaContainerLabels).
+		WithUser(root).
```
mieubrisse: This is definitely a permission escalation which - while common in the Docker ecosystem - is best avoided if possible. It's 98% probably fine, but if you want to do it the secure way you could consider:
- Using Bash file embedding (treats strings as if they were file, a la `<(echo "foobar")`).
- Configure using envvars, if possible

mieubrisse (approved):

PR #2707, opened 2025-04-09
feat: add grafloki to kurtosis config

tedim52:
## Description
This PR create KurtosisConfig v5 which makes the following configurable:

- Grafloki Configurations (should start before engine, graf/loki images)
- Turning off persistent volume logs collection - so that if a user configures other sinks and wants to save storage, kurtosis engine wont store duplicate logs

Additionally, it adds back the `kurtosis_service_name` label into the log stream to make labeling by service name possible in grafana.

YES


cli/cli/commands/grafloki/start/start.go
```diff
@@ -54,44 +52,18 @@ func run(
...
-			},
-			"labels": map[string]string{
-				"job": "kurtosis",
-			},
-		},
+	lokiSink, grafanaUrl, err := grafloki.StartGrafloki(ctx, clusterConfig.GetClusterType(), clusterConfig.GetGraflokiConfig())
+	if err != nil {
+		return err // already wrapped
 	}
+	logrus.Infof("Grafana running at %v", grafanaUrl)
```
mieubrisse: Is there a difference between logrus.Infof and out.PrintOutLn?

logrus seems more standard to me, but not sure why the previous was out.PrintOutLn

cli/cli/commands/grafloki/stop/stop.go
```diff
@@ -34,21 +32,9 @@ func run(
...
-		if err != nil {
-			return stacktrace.Propagate(err, "An error occurred stopping Grafana and Loki containers in Kubernetes.")
-		}
-	default:
-		return stacktrace.NewError("Unsupported cluster type: %v", clusterConfig.GetClusterType().String())
+	if err = grafloki.StopGrafloki(ctx, clusterConfig.GetClusterType()); err != nil {
+		return err // already wrapped
 	}
 
-	out.PrintOutLn("Successfully stopped Grafana and Loki containers.")
```
mieubrisse, line 52: Should there be a logrus log message?

cli/cli/helpers/engine_manager/engine_existence_guarantor.go
```diff
@@ -95,6 +95,9 @@ type engineExistenceGuarantor struct {
 
 	// Destinations the logs aggregator will deliver to
 	sinks logs_aggregator.Sinks
+
+	// If set to true, engine will not store logs in a persistent volume
+	shouldTurnOffPersistentVolumeLogsCollection bool
```
mieubrisse: I find that negatives make things harder to reason about, and increase the risk of bugs that someone will use the bool in the opposite way intended. Suggest instead, `shouldEnablePersistentVolumeLogsCollection`

cli/cli/helpers/engine_manager/engine_manager.go
```diff
@@ -247,6 +260,18 @@ func (manager *EngineManager) StartEngineIdempotentlyWithCustomVersion(ctx conte
 	}
 
 	clusterType := manager.clusterConfig.GetClusterType()
+
+	var lokiSink logs_aggregator.Sinks
+	var grafanaUrl string
+	if manager.clusterConfig.GetGraflokiConfig().ShouldStartBeforeEngine {
```
mieubrisse, line 268: Odd that you have to repeat yourself here and above. Shouldn't StartEngineManagerIdempotentlyWith(Current|Custom)Version share a codepath?

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -18,23 +21,24 @@ import (
 )
 
 const (
-	LokiContainerName    = "kurtosis-loki"
-	GrafanaContainerName = "kurtosis-grafana"
-	lokiReadinessPath    = "/ready"
+	LokiContainerNamePrefix    = "kurtosis-loki"
```
mieubrisse: Oh, is it expected to have a suffix? If so, I'd expect the prefix to end with something like `-`

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -67,7 +71,17 @@ func createGrafanaAndLokiContainers(ctx context.Context, dockerManager *docker_m
...
+	lokiImage := defaultLokiImage
+	if graflokConfig.LokiImage != "" {
+		lokiImage = graflokConfig.LokiImage
+	}
+
+	lokiUuid, err := uuid_generator.GenerateUUIDString()
+	if err != nil {
+		return "", stacktrace.Propagate(err, "An error occurred generating a uuid for Loki.")
+	}
+	lokiContainerName := fmt.Sprintf("%v-%v", LokiContainerNamePrefix, lokiUuid)
```
mieubrisse: Same message above - I'd expect the prefix to be the full prefix, no `-` being included here

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -78,7 +92,7 @@ func createGrafanaAndLokiContainers(ctx context.Context, dockerManager *docker_m
 		Build()
 	lokiContainerId, _, err := dockerManager.CreateAndStartContainer(ctx, lokiArgs)
 	if err != nil {
-		return "", stacktrace.Propagate(err, "An error occurred creating '%v' container.", LokiContainerName)
+		return "", stacktrace.Propagate(err, "An error occurred creating '%v' container.", LokiContainerNamePrefix)
```
mieubrisse: Shouldn't this be `lokiContainerName` instead?

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -107,7 +121,7 @@ func createGrafanaAndLokiContainers(ctx context.Context, dockerManager *docker_m
 		ApiVersion: int64(1),
 		Datasources: []GrafanaDatasource{
 			{
-				Name:      LokiContainerName,
+				Name:      LokiContainerNamePrefix,
```
mieubrisse: Think this is a bug - should be `lokiContainerName`, else it won't pick up any container

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -119,7 +133,6 @@ func createGrafanaAndLokiContainers(ctx context.Context, dockerManager *docker_m
 	if err != nil {
 		return "", stacktrace.Propagate(err, "An error occurred serializing Grafana datasource to yaml: %v", grafanaDatasourceYaml)
 	}
-	logrus.Infof("Grafana data source yaml %v", string(grafanaDatasourceYaml))
```
mieubrisse, line 122: Why removing this? Seems like it'd be good to leave as a Debug if it's too noisy, in the vein of https://github.com/kurtosis-tech/style-guide?tab=readme-ov-file#dont-add-preemptive-debug-statements-but-do-leave-your-debug-statements-in

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -130,7 +143,17 @@ func createGrafanaAndLokiContainers(ctx context.Context, dockerManager *docker_m
...
+	grafanaImage := defaultGrafanaImage
+	if graflokConfig.GrafanaImage != "" {
+		grafanaImage = graflokConfig.GrafanaImage
+	}
+	root := service_user.NewServiceUser(rootUserUid)
+	grafanaUuid, err := uuid_generator.GenerateUUIDString()
+	if err != nil {
+		return "", stacktrace.Propagate(err, "An error occurred generating a uuid for Grafana.")
+	}
+	grafanaContainerName := fmt.Sprintf("%v-%v", GrafanaContainerNamePrefix, grafanaUuid)
```
mieubrisse: Same comment as above, about the prefix being the full prefix

cli/cli/helpers/grafloki/docker_grafloki.go
```diff
@@ -215,6 +239,35 @@ func checkGrafanaAndLokiContainerExistence(ctx context.Context, dockerManager *d
 	return existsLoki && existsGrafana, fmt.Sprintf("http://%v:%v", lokiBridgeNetworkIpAddress, lokiPort), nil
 }
 
+func StopGrafLokiInDocker(ctx context.Context) error {
+	dockerManager, err := docker_manager.CreateDockerManager(EmptyDockerClientOpts)
```
mieubrisse, line 247: Question: will the outer function already have a DockerManager (thereby allowing us to take in a DockerManager rather than recreating a new one)?

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -89,6 +90,10 @@ func createGrafanaAndLokiDeployments(ctx context.Context, k8sManager *kubernetes
 		}
 	}()
 
+	lokiImage := defaultLokiImage
+	if graflokiConfig.LokiImage != "" {
+		lokiImage = defaultLokiImage
```
mieubrisse: Bug; should be `lokiImage = graflokiConfig.LokiImage`

cli/cli/helpers/grafloki/kubernetes_grafloki.go
```diff
@@ -254,6 +259,10 @@ func createGrafanaAndLokiDeployments(ctx context.Context, k8sManager *kubernetes
 		}
 	}()
 
+	grafanaImage := defaultGrafanaImage
+	if graflokiConfig.GrafanaImage != "" {
+		grafanaImage = defaultGrafanaImage
```
mieubrisse: Same species of bug

cli/cli/helpers/grafloki/shared.go
```diff
@@ -31,3 +39,59 @@ type GrafanaDatasources struct {
...
+	case resolved_config.KurtosisClusterType_Kubernetes:
+		err := StopGrafLokiInKubernetes(ctx)
+		if err != nil {
+			return stacktrace.Propagate(err, "An error occurred stopping Grafana and Loki containers in Kubernetes.")
+		}
+	default:
+		return stacktrace.NewError("Unsupported cluster type: %v", clusterType.String())
+	}
+
+	out.PrintOutLn("Successfully stopped Grafana and Loki containers.")
```
mieubrisse: Still confused about why out.PrintOutLn vs logrus. Seems like should be `logrus` always, so that we can capture CLI output by redirecting logrus' output source (if the CLI is using `out.PrintOutLn` then it's printing directly to STDOUT and tests can't capture and inspect that)

cli/cli/kurtosis_config/config_version/config_version.go
```diff
@@ -8,6 +8,7 @@ const (
 	ConfigVersion_v0 ConfigVersion = iota
 	ConfigVersion_v1
 	ConfigVersion_v2 // Fixed a typo in Kubernetes config, `enclave-size-in-Megabytes` -> `enclave-size-in-megabytes`
-	ConfigVersion_v3
-	ConfigVersion_v4
+	ConfigVersion_v3 // adds logs aggregator config
+	ConfigVersion_v4 // adds engine-node-name to KubernetesClusterConfig
+	ConfigVersion_v5 // adds GraflokiConfig to KurtosisClusterConfig
```
mieubrisse: Grafloki seems fine for naming variables in the code, but for user-facing stuff I'd explicitly call it GrafanaLokiConfig (no reason to shorthand where clarity is much more important than speed). Will wait to see if the actual config object below calls it this

cli/cli/kurtosis_config/overrides_migrators/config_overrides_migrators.go
```diff
@@ -26,13 +27,87 @@ type configOverridesMigrator = func(uncastedOldConfig interface{}) (interface{},
...
+			}
+
+			var newLogsAggregatorConfig *v5.LogsAggregatorConfigV5
+			if oldLogsAggregatorConfig != nil {
+				newLogsAggregatorConfig = &v5.LogsAggregatorConfigV5{
+					Sinks: oldLogsAggregatorConfig.Sinks,
+				}
+			}
+
+			newGraflokiConfig := &v5.GraflokiConfig{
```
mieubrisse: Hm yep, looks like it's getting called GraflokiConfig. Strikes me as it should be GrafanaLokiConfig (or maybe even better: InHouseLoggingConfig, or something like that to represent its purpose rather than its implementation??)

cli/cli/kurtosis_config/overrides_migrators/config_overrides_migrators.go
```diff
@@ -26,13 +27,87 @@ type configOverridesMigrator = func(uncastedOldConfig interface{}) (interface{},
...
+				ShouldStartBeforeEngine: false,
```
mieubrisse, line 73: Seems weird that this is configurable. If Grafana can grab the engine logs, wouldn't we always want to start it before the engine so it can capture engine logs too?

cli/cli/kurtosis_config/overrides_migrators/config_overrides_migrators.go
```diff
@@ -26,13 +27,87 @@ type configOverridesMigrator = func(uncastedOldConfig interface{}) (interface{},
...
+				GrafanaImage:            "",
+				LokiImage:               "",
+			}
+
+			defaultShouldTurnOffDefaultLogsSink := false
```
mieubrisse: 1. Same comment about negatives. Suggest making this `shouldEnableDefaultLogsSink`
2. the leading `default` confused me; maybe just `shouldEnableDefaultLogsSink`?

cli/cli/kurtosis_config/overrides_objects/v5/grafloki_config_v5.go
```diff
@@ -0,0 +1,7 @@
+package v5
+
+type GraflokiConfig struct {
```
mieubrisse: Same comment - IMO this would be better as `GrafanaLokiConfig`, or maybe even better, `InbuiltLoggingConfig` or something explaining the purpose

cli/cli/kurtosis_config/overrides_objects/v5/grafloki_config_v5.go
```diff
@@ -0,0 +1,7 @@
...
+	ShouldStartBeforeEngine bool   `yaml:"should-start-before-engine,omitempty"`
```
mieubrisse: If there's a good reason for this config to exist, maybe comments explaining why one would ever pick one or the other?

cli/cli/kurtosis_config/overrides_objects/v5/kurtosis_cluster_config_v5.go
```diff
@@ -0,0 +1,19 @@
...
+    Instead, to make changes, you will need to add a new version of the config
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! WARNING !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+*/
+
+type KurtosisClusterConfigV5 struct {
+	Type *string `yaml:"type,omitempty"`
+	// If we ever get another type of cluster that has configuration, this will need to be polymorphically deserialized
+	Config                       *KubernetesClusterConfigV5 `yaml:"config,omitempty"`
+	LogsAggregator               *LogsAggregatorConfigV5    `yaml:"logs-aggregator,omitempty"`
+	GraflokiConfig               *GraflokiConfig            `yaml:"grafloki-config,omitempty"`
```
mieubrisse: Same comment above about the name of this

cli/cli/kurtosis_config/overrides_objects/v5/logs_aggregator_config_v5.go
```diff
@@ -0,0 +1,5 @@
+package v5
+
+type LogsAggregatorConfigV5 struct {
```
mieubrisse, line 3: Thought while reading: you'll very likely want to update user-facing documentation with the latest info on how to configure the various new things added in this PR

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
```diff
@@ -24,17 +24,25 @@ const (
...
 type KurtosisClusterConfig struct {
-	kurtosisBackendSupplier     kurtosisBackendSupplier
-	engineBackendConfigSupplier engine_server_launcher.KurtosisBackendConfigSupplier
-	clusterType                 KurtosisClusterType
-	logsAggregator              LogsAggregatorConfig
+	kurtosisBackendSupplier                     kurtosisBackendSupplier
+	engineBackendConfigSupplier                 engine_server_launcher.KurtosisBackendConfigSupplier
+	clusterType                                 KurtosisClusterType
+	logsAggregator                              LogsAggregatorConfig
+	graflokiConfig                              GrafanaLoki
```
mieubrisse: Same comment as above about naming

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
```diff
@@ -73,11 +81,27 @@ func NewKurtosisClusterConfigFromOverrides(clusterId string, overrides *v4.Kurto
...
+	var grafloki GrafanaLoki
+	if overrides.GraflokiConfig != nil {
+		grafloki = GrafanaLoki{
+			ShouldStartBeforeEngine: overrides.GraflokiConfig.ShouldStartBeforeEngine,
+			GrafanaImage:            overrides.GraflokiConfig.GrafanaImage,
+			LokiImage:               overrides.GraflokiConfig.LokiImage,
+		}
+	}
+
+	shouldTurnOffDefaultLogsSink := defaultShouldTurnOffDefaultLogsSink
```
mieubrisse: 1. Same mention about negatives vs positives
2. Ohhh wait, so `defaultShouldTurnOffDefaultLogsSink` is used here as well as above?? (EDIT: wait no it's not... we seemingly have the `defaultShouldTurnOffDefaultLogsSink` constant, and we also have a variable called `defaultShouldTurnOffDefaultLogsSink` that I commented on earlier. Do these need to be unified?)

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config.go
```diff
@@ -101,12 +125,20 @@ func (clusterConfig *KurtosisClusterConfig) GetLogsAggregatorConfig() LogsAggreg
 	return clusterConfig.logsAggregator
 }
 
+func (clusterConfig *KurtosisClusterConfig) GetGraflokiConfig() GrafanaLoki {
+	return clusterConfig.graflokiConfig
+}
+
+func (clusterConfig *KurtosisClusterConfig) ShouldTurnOffPersistentVolumeLogsCollection() bool {
```
mieubrisse: Comment about negatives vs positives

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config_test.go
```diff
@@ -144,22 +160,104 @@ func TestNewKurtosisClusterConfigLogsAggregatorFullConfig(t *testing.T) {
...
+		EnclaveSizeInMegabytes: &kubernetesEnclaveSizeInMB,
+		EngineNodeName:         &kubernetesEngineNodeName,
+	}
+	kurtosisClusterConfigOverrides := v5.KurtosisClusterConfigV5{
+		Type:   &kubernetesType,
+		Config: &kubernetesFullConfig,
+		LogsAggregator: &v5.LogsAggregatorConfigV5{
+			Sinks: map[string]map[string]interface{}{
+				"elasticsearch": {
+					"type": "elasticsearch",
```
mieubrisse, line 175: Nice job adding tests 💪

cli/cli/kurtosis_config/resolved_config/kurtosis_cluster_config_test.go
```diff
@@ -144,22 +160,104 @@ func TestNewKurtosisClusterConfigLogsAggregatorFullConfig(t *testing.T) {
...
+		Config:         &kubernetesFullConfig,
+		LogsAggregator: nil,
+		GraflokiConfig: &v5.GraflokiConfig{
+			ShouldStartBeforeEngine: false,
+			GrafanaImage:            "grafana:1.32",
+			LokiImage:               "loki:1.32",
+		},
+		ShouldTurnOffDefaultLogsSink: nil,
+	}
+	_, err := NewKurtosisClusterConfigFromOverrides("test", &kurtosisClusterConfigOverrides)
```
mieubrisse: Prob want to have asserts that your overrides are getting picked up (to guard against the bug earlier in this PR, with overrides not being used on the Kubernetes version)

cli/cli/kurtosis_config/resolved_config/kurtosis_config.go
```diff
@@ -20,8 +20,9 @@ const (
...
+	defaultShouldTurnOffDefaultLogsSink = false
```
mieubrisse: 1. Same comment about negation
2. Same question about if this value should be used when doing the config parsing (maybe not; I'm not sure)

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -41,7 +41,8 @@ import (
 )
 
 const (
-	isResourceInformationComplete = true
+	isResourceInformationComplete                      = true
+	defaultShouldTurnPersistentVolumeLogsCollectionOff = false
```
mieubrisse: negation comment

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -389,6 +392,7 @@ func (backend *DockerKurtosisBackend) CreateLogsAggregator(
 		logsAggregatorContainer,
 		httpPortNum,
 		sinks,
+		defaultShouldTurnPersistentVolumeLogsCollectionOff,
```
mieubrisse: Point of confusion: CreateEngine has a new param about whether to use persistent volume (makes sense). Shouldn't `DockerKurtosisBackend.CreateLogsAggregator` have a similar one, since presumably it's being called by the Engine which has the should-use-persistent-volume value? Otherwise, why even have the parameter to `logs_aggregator_functions.CreateLogsAggregator` if it's always a static value?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
```diff
@@ -107,6 +108,7 @@ func CreateEngine(
 		logsAggregatorContainer,
 		defaultHttpLogsAggregatorPortNum,
 		sinks,
+		shouldTurnOffPersistentVolumeLogsCollection,
```
mieubrisse: Huh now I'm confused - why is DockerKurtosisBackend even exposing CreateLogsAggregator if the engine (which is the only thing that creates it to my knowledge) is bypassing it and going directly to `logs_aggregator_functions.CreateLogsAggregator`? I'd expect that the engine would use `DockerKurtosisBackend.CreateLogsAggregator` or it wouldn't exist; as-is, there are now two ways to create a logs aggregator and I think I'd be confused which I should be using.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/create_logs_aggregator.go
```diff
@@ -22,6 +22,7 @@ func CreateLogsAggregator(
 	logsAggregatorContainer LogsAggregatorContainer,
 	logsAggregatorHttpPortNumber uint16,
 	sinks logs_aggregator.Sinks,
+	shouldTurnOffPersistentVolumeLogsCollection bool,
```
mieubrisse: comment about negatives

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/vector_config.go
```diff
@@ -23,9 +23,12 @@ func newVectorConfig(
 	listeningPortNumber uint16,
 	httpPortNumber uint16,
 	sinks logs_aggregator.Sinks,
+	shouldTurnOffPersistentVolumeLogsCollection bool,
 ) *VectorConfig {
-	reconciledSinks := map[string]map[string]interface{}{
-		logs_aggregator.DefaultSinkId: {
+	reconciledSinks := map[string]map[string]interface{}{}
+
+	if !shouldTurnOffPersistentVolumeLogsCollection {
```
mieubrisse: this is where the positives vs negatives thing makes it easier to reason about.

e.g. I parse this as, "I'm _not_ turning _off_ persistent volume logs collection"

compare with `if shouldEnablePersistentVolumeLogsCollection`

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/logs_aggregator_functions/implementations/vector/vector_configuration_creator_factory.go
```diff
@@ -8,7 +8,8 @@ func createVectorConfigurationCreatorForKurtosis(
 	listeningPortNumber uint16,
 	httpPortNumber uint16,
 	sinks logs_aggregator.Sinks,
+	shouldTurnOffPersistentVolumeLogsCollection bool,
```
mieubrisse: Btw if you do choose to go for positive vs negative, be careful doing the refactor. If you just start renaming the variable and don't trace down where it gets used, the IDE won't throw an error (b/c it's still a bool variable) but you'll accidentally be using the inverted version

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go
```diff
@@ -245,6 +245,12 @@ func (provider *dockerEnclaveObjectAttributesProviderImpl) ForUserServiceContain
 	}
 	labels[docker_label_key.LogsServiceUUIDDockerLabelKey] = serviceUuidLabel
 
+	serviceNameLabel, err := docker_label_value.CreateNewDockerLabelValue(serviceNameStr)
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred creating a Docker label value from GUID string '%v'", serviceNameStr)
```
mieubrisse: Looks like a bad copy/paste; don't think this is a GUID string

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/enclave_object_attributes_provider.go
```diff
@@ -245,6 +245,12 @@ func (provider *dockerEnclaveObjectAttributesProviderImpl) ForUserServiceContain
...
+	}
+	labels[docker_label_key.LogsServiceUUIDDockerLabelKey] = serviceNameLabel
```
mieubrisse: Bug; key is wrong and will overwrite `LogsServiceUUIDDockerLabelKey`

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/kubernetes_kurtosis_backend.go
```diff
@@ -36,9 +36,10 @@ import (
 )
 
 const (
-	isResourceInformationComplete = false
-	noProductionMode              = false
-	anyNodeEngineNodeName         = "" // engine can be scheduled by k8s on any node
+	isResourceInformationComplete                      = false
+	noProductionMode                                   = false
+	anyNodeEngineNodeName                              = "" // engine can be scheduled by k8s on any node
+	defaultShouldTurnOffPersistentVolumeLogsCollection = false
```
mieubrisse, line 42: Negative vs positive

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_aggregator_functions/implementations/vector/vector_config.go
```diff
@@ -23,9 +23,12 @@ func newVectorConfig(
 	listeningPortNumber uint16,
 	httpPortNumber uint16,
 	sinks logs_aggregator.Sinks,
+	shouldTurnOffPersistentVolumeLogsCollection bool,
 ) *VectorConfig {
-	reconciledSinks := map[string]map[string]interface{}{
-		logs_aggregator.DefaultSinkId: {
+	reconciledSinks := map[string]map[string]interface{}{}
+
+	if !shouldTurnOffPersistentVolumeLogsCollection {
```
mieubrisse: negative vs positive

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
```diff
@@ -472,6 +472,7 @@ func generateFluentBitConfigStr(
 		CheckpointDbMountPath  string
 		LogsEnclaveUUIDLabel   string
 		LogsServiceUUIDLabel   string
+		LogsServiceNameLabel   string
```
mieubrisse, line 475: This is `LogsServiceNameLabel`; maybe a name you could rename `GraflokiConfig` to `InbuiltLogsServiceConfig`?

container-engine-lib/lib/backend_impls/kubernetes/object_attributes_provider/enclave_object_attributes_provider.go
```diff
@@ -188,6 +188,12 @@ func (provider *kubernetesEnclaveObjectAttributesProviderImpl) ForUserServicePod
 	}
 	labels[kubernetes_label_key.LogsServiceUUIDKubernetesLabelKey] = uuidLabelValue
 
+	nameLabelValue, err := kubernetes_label_value.CreateNewKubernetesLabelValue(string(serviceName))
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred creating a short GUID Kubernetes label value from GUID string '%v'", string(serviceName))
```
mieubrisse: Careful with the copy-pasted log messages; this looks like it wasn't modified from the line above

container-engine-lib/lib/backend_impls/metrics_reporting/metrics_reporting_kurtosis_backend.go
```diff
@@ -57,6 +57,7 @@ func (backend *MetricsReportingKurtosisBackend) CreateEngine(
 	shouldStartInDebugMode bool,
 	githubAuthToken string,
 	sinks logs_aggregator.Sinks,
+	shouldTurnOffPersistentVolumeLogsCollection bool,
```
mieubrisse: negatives vs positives

engine/launcher/engine_server_launcher/engine_server_launcher.go
```diff
@@ -52,6 +52,7 @@ func (launcher *EngineServerLauncher) LaunchWithDefaultVersion(
 	domain string,
 	logRetentionPeriod string,
 	sinks logs_aggregator.Sinks,
+	shouldTurnOffPersistentVolumeLogsCollection bool,
```
mieubrisse: negatives vs positives (think you get the idea so I'll stop commenting)

mieubrisse (approved):
Couple bugs, and some suggestions on renames to make things easier to understand

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: ah good point

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: agreed

cli/cli/commands/grafloki/start/start.go
tedim52: using logrus from here on out - as mentioned in our style guide - only going to use `out.Print..` for text written to stdout that should be piped into other commands or processed downstream

cli/cli/commands/grafloki/stop/stop.go
tedim52: added this back in

cli/cli/helpers/grafloki/docker_grafloki.go
tedim52: nah it doesn't

cli/cli/kurtosis_config/overrides_migrators/config_overrides_migrators.go
tedim52: Grafana is only capturing logs from services in enclaves right now - altho it could one day capture logs.

Grafana / Loki is not a default, its an available option if users want it. The default is the simpler persistent volume /kurtosis service logs setup. 

We give the users the ability to use Grafana / Loki - `grafloki start` `grafloki stop` AND let them make it sticky by having kurtosis always do `grafloki start` before `engine start` when configured.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/engine_functions/create_engine.go
tedim52: ye - tbh KurtosisBackend shouldn't expose `CreateLogsAggregator` since its not being used by anyone - but I'm going to leave that for a separate PR.

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/logs_collector_functions/implementations/fluentbit/fluentbit_logs_collector_daemonset.go
tedim52: referring back to some of our conversations from last PR - not viewing Grafana/Loki as part of Kurtosis core so gonna avoid naming it to anything that indicates such - viewing it as an external tool provided as an option by the kurtosis platform - but i will rename it `GrafanaLokiConfig`

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
tedim52: addressed in the other comment - but `CreateLogsAggregator` should be taken off KurtosisBackend interface but leaving that for another change

cli/cli/kurtosis_config/overrides_objects/v5/grafloki_config_v5.go
tedim52: will add a comment

cli/cli/kurtosis_config/overrides_objects/v5/grafloki_config_v5.go
tedim52: renaming to GrafanaLokiConfig

cli/cli/kurtosis_config/config_version/config_version.go
tedim52: renaming to `GrafanaLokiConfig`

cli/cli/kurtosis_config/overrides_migrators/config_overrides_migrators.go
tedim52: renaming to `GrafanaLokiConfig` (from discussions of last PR - grafana loki is an additional / optional extension to Kurtosis but not in House)

cli/cli/kurtosis_config/overrides_objects/v5/logs_aggregator_config_v5.go
tedim52: nice will add some docs

cli/cli/helpers/engine_manager/engine_manager.go
tedim52: agreed - could probably be refactored to use the same code path but going to leave this for a separate change

cli/cli/helpers/engine_manager/engine_existence_guarantor.go
tedim52: ye i agree with this - i went with the negative that it was clear that the default is that `PersistentVolumeLogsCollection` is turned of - if it's not the negative then `shouldEnablePersistentVolumeLogsCollection` seems like it's not the default

tedim52: i think the right move is to go with the positive + set default to true + comment why thats the default , warn against changing. not a big fan that the the default is set at the cluster config level

Issue #2622, opened 2025-01-02
`kurtosis` doesn't clean `Ingress` objects in k8s


0x416e746f6e:
### What's your CLI version?

1.4.3

### Description & steps to reproduce

After stopping the enclave in k8s, `kurtosis` does not delete the `Ingress` objects (even though the `Service` objects referred by these `Ingress`es _are_ removed).

### Desired behavior

Remove `Ingress` objects just as well.

### What is the severity of this bug?

Painful; this is causing significant friction in my workflow.

### What area of the product does this pertain to?

CLI: the Command Line Interface

PR #2715, opened 2025-04-17
fix: remove k8s ingress on stop user services

tedim52:
## Description
This was leaving ingresses around. Additionally, it was preventing service updates in k8s. Service updates involve removing the registered user, then recreating its resources. Because the ingress wasn't being removed a naming conflict would occur when the ingress resource was being recreated.
image

NO

## References
https://github.com/kurtosis-tech/kurtosis/issues/2622


praetoriansentry (approved):
lgtm

PR #2716, opened 2025-04-17
fix: return empty deployment if not found

tedim52:
## Description
Restarts on grafloki start over k8s were not working because of the way the not found error was being handled. This change returns a nil object if a deployment isn't found by k8s instead of returning an error.


praetoriansentry (approved):
lgtm

PR #2730, opened 2025-04-23
fix: task edits on k8s

tedim52:
## Description
Enclave edits were not working for packages with tasks. This was because of the way that the service network service registrations and service k8s objects for tasks were being handled. This fix makes two changes:

image

1. Copies `add_service` for tasks where if a task exists, it should only be updated on an enclave edit instead of added again.
2. To get the above to work, a check for the tasks service registration is made. However, `serviceNetwork.RemoveService` would remove the tasks pods, ingress (as desired) after execution but also the tasks service registration. This change uses `serviceNetwork.StopService` which will also remove the tasks pods, ingress but not the tasks service registration. This way a subsequent edit will check the service registration, sees it exists, and update, instead of add.

This does have two effects:
- kt enclave inspect will now show the status of task as stopped, whereas before the service wouldn't appear in enclave inspect
- edits will not work when using `--non-blocking-mode` because `serviceNetwork.StopService` will never get called

TODO:

- [ ] Expose DestroyService on service network
- [ ] Use serviceNetwork.DestroyService instead of serviceNetwork.RemoveService
- [ ] Document what the differences are

### DIFF
--- a/container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/start_user_services.go
+++ b/container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/start_user_services.go
@@ -3,8 +3,9 @@ package user_services_functions
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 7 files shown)


barnabasbusa:

PR #2760, opened 2025-06-15
fix: in `GetFullServiceConfigStarlark` use `MinMillicpus` instead of `MaxMemory`

nalepae:
## Description
`GetFullServiceConfigStarlark`: Use `MinMillicpus` instead of `MaxMemory`
Also fixed a few other mistakes.

## REMINDER: Tag Reviewers, so they get notified to review
@barnabasbusa 

YES


tedim52:
looks like `bridgeNetworkIpAddressAddr` change broke smth, lemme check out y

tedim52:
Ah I found the issue with bridge network address change. Mind undoing that change in this PR so I can make a separate fix for it (don't want to blow up the scope of this fix)? @nalepae

tedim52 (approved):

PR #2768, opened 2025-06-23
fix: refactor `HttpRequestService ` and `Recipe` to remove unnecessary calls to container engine

tedim52:
## Description
This change significantly reduces the number of calls made to the Docker daemon or Kubernetes API server by avoiding redundant `GetService` invocations.

Previously, every call to `serviceNetwork.HttpRequestService` triggered a call to `serviceNetwork.GetService`, which looked like this:
```
func (network *DefaultServiceNetwork) HttpRequestService(ctx context.Context, serviceIdentifier string...
	userService, getServiceErr := network.GetService(ctx, serviceIdentifier)
	....
```

Each `GetService` call internally performed a `ContainerList`, followed by a `ContainerInspect` for every container:
```
...
dockerContainers, err := manager.dockerClient.ContainerList(ctx, opts)

	dockerContainersDetails := []types.ContainerJSON{}
	for _, dockerContainer := range dockerContainers {
		dockerContainerDetails, err := manager.InspectContainer(ctx, dockerContainer.ID)
                 ...
		dockerContainersDetails = append(dockerContainersDetails, dockerContainerDetails)
	}
...
```
This leads to a large number of redundant Docker API calls, especially in scenarios where `HttpRequestService` is invoked frequently — e.g., during runs like kt run github.com/ethpandaops/ethereum-package. In cases with >10 services and many readiness or recipe checks, this causes unnecessary CPU load.

This PR refactors the `HttpRequestService` and `Recipe` APIs to accept a `service` object directly instead of a `serviceName`. `GetService` is now called once, getting rid of thousands of redundant Docker daemon/k8s api server calls.

NO


core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -629,12 +629,9 @@ func (network *DefaultServiceNetwork) RunExecs(ctx context.Context, userServiceC
 	return successfulExecs, failedExecs, nil
 }
 
-func (network *DefaultServiceNetwork) HttpRequestService(ctx context.Context, serviceIdentifier string, portId string, method string, contentType string, endpoint string, body string, headers map[string]string) (*http.Response, error) {
+func (network *DefaultServiceNetwork) HttpRequestService(ctx context.Context, userService *service.Service, portId string, method string, contentType string, endpoint string, body string, headers map[string]string) (*http.Response, error) {
+	serviceIdentifier := userService.GetRegistration().GetName()
```
mieubrisse, line 633: Just verifying - the `identifier` being the name is indeed correct, right? We don't want the UUID?

core/server/api_container/server/service_network/default_service_network.go
```diff
@@ -629,12 +629,9 @@ func (network *DefaultServiceNetwork) RunExecs(ctx context.Context, userServiceC
...
 	logrus.Debugf("Making a request '%v' '%v' '%v' '%v' '%v' '%v'", serviceIdentifier, portId, method, contentType, endpoint, body)
-	userService, getServiceErr := network.GetService(ctx, serviceIdentifier)
```
mieubrisse, line 634: ahhhhhahahaha... perf tuning do be like this. remove one line and everything is 10x faster

core/server/api_container/server/startosis_engine/kurtosis_instruction/exec/exec.go
```diff
@@ -185,9 +186,14 @@ func (builtin *ExecCapabilities) Validate(_ *builtin_argument.ArgumentValuesSet,
 }
 
 func (builtin *ExecCapabilities) Execute(ctx context.Context, _ *builtin_argument.ArgumentValuesSet) (string, error) {
-	result, err := builtin.execRecipe.Execute(ctx, builtin.serviceNetwork, builtin.runtimeValueStore, builtin.serviceName)
+	serviceNameStr := string(builtin.serviceName)
+	service, err := builtin.serviceNetwork.GetService(ctx, serviceNameStr)
```
mieubrisse, line 190: Had a thought: the serviceNetwork caching information on behalf of Docker/k8s could also be interesting. Now that you're reducing the number of calls it might be a moot point, but I don't _think_ (?) there's any reason not to cache a name -> container_id lookup.

Runtime info would still need to be live; maybe that could be a separate call in ServiceNetwork?

Orrrrr maybe none of this is necessary and you can ignore

core/server/api_container/server/startosis_engine/kurtosis_instruction/shared_helpers/service_helpers.go
```diff
@@ -45,14 +45,17 @@ func ExecuteServiceAssertionWithRecipe(
...
-	serviceName service.ServiceName,
+	service *service.Service,
 	recipe recipe.Recipe,
 	valueField string,
 	assertion string,
 	target starlark.Comparable,
 	interval time.Duration,
 	timeout time.Duration,
 ) (map[string]starlark.Comparable, int, error) {
+	if service == nil || service.GetRegistration() == nil {
```
mieubrisse, line 57: Nice sanity check

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/test_engine/add_services_framework_test.go
```diff
@@ -68,16 +70,16 @@ func (suite *KurtosisPlanInstructionTestSuite) TestAddServices() {
 		mock.Anything,
 	).Times(1).Return(
 		map[service.ServiceName]*service.Service{
-			testServiceName:  service.NewService(service.NewServiceRegistration(testServiceName, testServiceUuid, testEnclaveUuid, nil, string(testServiceName)), nil, nil, nil, container.NewContainer(container.ContainerStatus_Running, testContainerImageName, nil, nil, nil)),
-			testServiceName2: service.NewService(service.NewServiceRegistration(testServiceName2, testServiceUuid2, testEnclaveUuid, nil, string(testServiceName2)), nil, nil, nil, container.NewContainer(container.ContainerStatus_Running, testContainerImageName, nil, nil, nil)),
+			testServiceName:  testService1,
```
mieubrisse, line 73: Thanks for breaking these out!

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/test_engine/exec_named_args_framework_test.go
```diff
@@ -11,20 +13,29 @@ import (
...
+var execWithNamedArgsTestService *service.Service = getService(execServiceName)
+
 type execWithNamedArgsTestCase struct {
 	*testing.T
 	serviceNetwork    *service_network.MockServiceNetwork
 	runtimeValueStore *runtime_value_store.RuntimeValueStore
 }
 
 func (suite *KurtosisPlanInstructionTestSuite) TestExecWithNamedArgs() {
+	suite.serviceNetwork.EXPECT().GetService(
```
mieubrisse, line 31: Niiice job thoroughly adding tests.. you were always great at this; one of your strongest points as an engineer I feel!

core/server/api_container/server/startosis_engine/kurtosis_starlark_framework/test_engine/shared_functions.go
```diff
@@ -51,3 +57,25 @@ func getCustomReadyConditionsScripPart(
 		timeoutStr,
 	)
 }
+
+func getService(serviceName service.ServiceName) *service.Service {
+	return service.NewService(
+		service.NewServiceRegistration(
+			serviceName,
+			service.ServiceUUID(""),
```
mieubrisse: Worth making this and the enclave UUID a random UUID, just to be more realistic?

core/server/api_container/server/startosis_engine/recipe/exec_recipe.go
```diff
@@ -124,14 +125,11 @@ func (recipe *ExecRecipe) Execute(
 		commandWithRuntimeValue = append(commandWithRuntimeValue, maybeSubCommandWithRuntimeValues)
 	}
 
-	serviceNameStr := string(serviceName)
-	if serviceNameStr == "" {
-		return nil, stacktrace.NewError("The service name parameter can't be an empty string")
-	}
+	serviceNameStr := string(service.GetRegistration().GetName())
 
 	execResult, err := serviceNetwork.RunExec(ctx, serviceNameStr, commandWithRuntimeValue)
```
mieubrisse, line 130: Question: wouldn't this `serviceNetwork.RunExec` suffer from the same bug? Shouldn't this be passing in a `service` instead? (seems like if we're fixing some methods to avoid this bug, we should fix all of them right?)

mieubrisse (approved):

core/server/api_container/server/startosis_engine/kurtosis_instruction/exec/exec.go
tedim52: completely agree I think a `cache` is the next step for optimization here - I think before adding a cache, I want to get rid of as many unnecessary calls to the container engine as possible at this layer, and then from there measure the improvement a cache would offer to justify the overhead/coordination of maintaining it (maybe there's not that much though)?

core/server/api_container/server/service_network/default_service_network.go
tedim52: fr

core/server/api_container/server/service_network/default_service_network.go
tedim52: yep, it's purely being used for logging out in this function

core/server/api_container/server/startosis_engine/recipe/exec_recipe.go
tedim52: `RunExec` does checks the enclave db every time it's called but doesn't make the expensive GetService calls that `HttpServiceRequest` does so it's not as bad. 

Also - it's rare `ExecRecipe`'s are used in `ReadyCondition` (although more common with waits), so I don't think changing the interface here is as important. Maybe for consistency but then the other methods taking in `serviceIdentifier` (e.g. `StartService`, `StopService`, `RemoveService`) should be changed as well and that refactor is out of scope of this change.

barnabasbusa (approved):

PR #2773, opened 2025-07-07
fix: docker auth bugs

skylenet:
There were 2 problems:

1. We were just propagating the `Auth` field to the `X-Registry-Auth` header, via `ImagePullOptions.RegistryAuth`. 
We do need to also set the `Username`, `Password` and `ServerAddress`. The docs are not very clear, https://docs.docker.com/reference/api/engine/version/v1.45/#section/Authentication , but not having those seemed to affect the authentication. It's also weird that you don't get an error when you provide uncomplete authentication headers... anyway.

2. Images like `alpine:3.17` didn't get the right registry address. They did point to `alpine:3.17` , isntead of using `https://index.docker.io/v1/`. This made it impossible to get the credentials when pulling images that were using `.` in their tags. I made sure to strip out the tag part. 

I've also updated the tests so that these 2 situations are covered. 

I've also tested this locally, using a simple auth config file, like we use in CI runs that only looks like this:

```json
{
        "auths": {
                "https://index.docker.io/v1/": {
                        "auth": "xxxxx"
                }
        }
}
```

I also did some manual test by doing the following in a loop after removing my docker auth config:
- Run kurtosis 
- Run a package
- Cleanup enclave + stop engine
- Docker remove all images that were pulled
- Repeat

After some minutes I hit the 100 pull limit for unauthenticated requests. I kept it running like this for a couple of minutes.

Then I did a docker login which populated my ~/.docker./config.json . After restarting the engine. There were no more rate limit messages.

### DIFF
--- a/container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
+++ b/container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
@@ -238,6 +238,8 @@ func GetAuthFromDockerConfig(repo string) (*registry.AuthConfig, error) {
 				ac.Username = string(decodedAuth[:usernamePasswordSeparatorIndex])
 				ac.Password = string(decodedAuth[usernamePasswordSeparatorIndex+1:])
 			}
+		} else {
+			return nil, stacktrace.NewError("no username or password or auth found for registry '%s'", registryHost)
 		}
 
 		return &ac, nil
--- a/container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
+++ b/container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
@@ -223,9 +223,12 @@ func GetAuthFromDockerConfig(repo string) (*registry.AuthConfig, error) {
 			RegistryToken: auth.RegistryToken,
 		}
 
-		// If the base64 encoded auth field is set, decode it and also set the Username and Password
-		if auth.Auth != "" {
-			// Base64 decode it and set the Username and Password
+		// If the username or password fields are set, set them in the AuthConfig (Overrides the decoded auth)
+		if auth.Username != "" && auth.Password != "" {
+			ac.Username = auth.Username
+			ac.Password = auth.Password
+		} else if auth.Auth != "" {
+			// If the base64 encoded auth field is set, decode it and also set the Username and Password
 			decodedAuth, err := base64.StdEncoding.DecodeString(auth.Auth)
 			if err != nil {
 				return nil, stacktrace.Propagate(err, "error decoding auth for registry '%s'", registryHost)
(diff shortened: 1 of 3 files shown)


container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
```diff
@@ -206,7 +209,41 @@ func GetAuthFromDockerConfig(repo string) (*registry.AuthConfig, error) {
...
+		// If the base64 encoded auth field is set, decode it and also set the Username and Password
+		if auth.Auth != "" {
+			// Base64 decode it and set the Username and Password
+			decodedAuth, err := base64.StdEncoding.DecodeString(auth.Auth)
+			if err != nil {
+				return nil, stacktrace.Propagate(err, "error decoding auth for registry '%s'", registryHost)
+			}
+			ac.Username = string(decodedAuth[:strings.IndexByte(string(decodedAuth), ':')])
+			ac.Password = string(decodedAuth[strings.IndexByte(string(decodedAuth), ':')+1:])
+			ac.Auth = auth.Auth
```
tedim52: nit: is it necessary to set `auth.Auth` again here? seems like it's already happening above

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
```diff
@@ -206,7 +209,41 @@ func GetAuthFromDockerConfig(repo string) (*registry.AuthConfig, error) {
...
+		}
+
+		// If the base64 encoded auth field is set, decode it and also set the Username and Password
+		if auth.Auth != "" {
+			// Base64 decode it and set the Username and Password
+			decodedAuth, err := base64.StdEncoding.DecodeString(auth.Auth)
+			if err != nil {
+				return nil, stacktrace.Propagate(err, "error decoding auth for registry '%s'", registryHost)
+			}
+			ac.Username = string(decodedAuth[:strings.IndexByte(string(decodedAuth), ':')])
```
tedim52: nit: store `strings.IndexByte(string(decodedAuth), ':')` in a variable like `usernamePasswordSeparatorIndex` or `indexOfUsernamePasswordSeparator` (or whatever you like) and reuse variable to make this more slightly readable?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
```diff
@@ -206,7 +209,41 @@ func GetAuthFromDockerConfig(repo string) (*registry.AuthConfig, error) {
...
+			ac.Password = string(decodedAuth[strings.IndexByte(string(decodedAuth), ':')+1:])
+			ac.Auth = auth.Auth
+		}
+
+		// If the username or password fields are set, set them in the AuthConfig (Overrides the decoded auth)
+		if auth.Username != "" {
```
tedim52: it seems like the `Username` and `Password` from `auth` takes precedence over decoded `Username` and `Password` from auth? 

if thats the case, you could avoid decoding `auth` entirely by putting this check before the above check?

container-engine-lib/lib/backend_impls/docker/docker_manager/docker_auth.go
```diff
@@ -206,7 +209,41 @@ func GetAuthFromDockerConfig(repo string) (*registry.AuthConfig, error) {
...
+			ac.Username = auth.Username
+		}
+		if auth.Password != "" {
+			ac.Password = auth.Password
+		}
```
tedim52, line 243: unlikely, but is it ever possible to run into a case where `Username` is non empty but `Password` is empty? 
in which case you'd have the `Password` from decoded auth and the `Username` from `auth`? Probably unlikely but maybe worth adding a test for

tedim52 (approved):

PR #2774, opened 2025-07-07
fix: speed up enclave inspect (and other CLI commands)

tedim52:
### Description
Previously, `enclave inspect` became slow when more than 3-4 enclaves were running, mainly because `KurtosisContext.GetEnclave` internally called the slower `engineClient.GetEnclaves` method and post filtered.

This PR improves performance by:

- Updates a direct `GetEnclavesByUuids` endpoint to the engine client
- Updating `KurtosisContext.GetEnclave` to use this new endpoint
- Removing 2 unnecessary `GetEnclave` calls

With 5 empty enclaves running locally, inspect time drops from `~2.4s` to less than `100ms`. Additionally, the performance of any CLI command that calls `KurtosisContext.GetEnclave` will improve with the updated implementation.

Note: Performance still scales with the number of enclaves due to how `KurtosisBackend.GetAPIContainers` works. A future improvement would be to cache a map of container ids to enclave ids in `KurtosisBackend`. More details on this have been commented in code.

YES


cli/cli/commands/enclave/inspect/files_artifacts.go
```diff
@@ -17,7 +18,7 @@ const (
 )
 
 func printFilesArtifacts(ctx context.Context, kurtosisCtx *kurtosis_context.KurtosisContext, enclaveInfo *kurtosis_engine_rpc_api_bindings.EnclaveInfo, showFullUuids bool, _ bool) error {
-	enclaveContext, err := kurtosisCtx.GetEnclaveContext(ctx, enclaveInfo.GetName())
+	enclaveContext, err := kurtosisCtx.GetEnclaveContextFromEnclaveInfo(ctx, enclaveInfo)
```
tedim52, line 21: `GetEnclaveContext` makes a call to `GetEnclave` which is unnecessary since `enclaveInfo` is already provided in the params to this function. using `GetEnclaveContextFromEnclaveInfo` avoid the extra call

cli/cli/kurtosis_gateway/server/engine_gateway/engine_gateway_service_server.go
```diff
@@ -129,26 +130,58 @@ func (service *EngineGatewayServiceServer) GetEnclaves(ctx context.Context, in *
...
 	for enclaveId, enclaveInfo := range responseEnclaves {
 		var runningApiContainerGateway *runningApiContainerGateway
 		runningApiContainerGateway, isRunning := service.enclaveIdToRunningGatewayMap[enclaveId]
 		// If the gateway isn't running, start it
 		if !isRunning {
 			runningApiContainerGateway, err = service.startRunningGatewayForEnclave(enclaveInfo)
+			if err != nil {
+				return nil, stacktrace.Propagate(err, "Expected to be able to start a local gateway for enclave '%v', instead a non-nil error was returned", enclaveId)
+			}
 			defer func() {
```
tedim52, line 173: fixed the defer undo logic here to follow style guide (err check comes before defer undo) https://github.com/kurtosis-tech/style-guide/blob/1fdac40514204f3971f23d48f8a2eb1c73aad12d/guides/picking-up-your-toys.md#deferred-cleanup-must-come-after-error-checking

api/protobuf/engine/engine_service.proto
```diff
@@ -151,6 +153,11 @@ message GetEnclavesResponse {
   map<string, EnclaveInfo> enclave_info = 1;
 }
 
+message GetEnclaveResponse {
+  //All the enclave information inside this object
+  EnclaveInfo enclave_info = 1;
```
mieubrisse: Haven't read full PR yet, but I'm curious why GetEnclave would be meaningfully faster than GetEnclaves. Intuitively, I'd expect this would only save the call for "filter for the enclave that matches our requested identifier" (which should be negligible unless the user has 1000s of enclaves), but will read more.

api/rust/src/engine_api.rs
```diff
@@ -110,6 +110,13 @@ pub struct GetEnclavesResponse {
         EnclaveInfo,
     >,
 }
+#[allow(clippy::derive_partial_eq_without_eq)]
```
mieubrisse: Assuming this code is generated; not going to review

engine/server/engine/enclave_manager/enclave_manager.go
```diff
@@ -644,7 +659,33 @@ func (manager *EnclaveManager) getEnclavesWithoutMutex(
...
+func (manager *EnclaveManager) getEnclaveWithoutMutex(
+	ctx context.Context,
+	enclaveIdentifier string,
+) (*types.EnclaveInfo, error) {
+	enclaves, err := manager.kurtosisBackend.GetEnclaves(ctx, getEnclaveFilter(enclaveIdentifier))
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred getting enclave '%v'", enclaveIdentifier)
+	}
+
+	enclaveObj, existsEnclave := enclaves[enclave.EnclaveUUID(enclaveIdentifier)]
```
mieubrisse: Is there a risk that the `enclaves` map that comes back has the enclave, but it's keyed by a different ID (rather than `enclaveIdentifier`)?

engine/server/engine/enclave_manager/enclave_manager.go
```diff
@@ -644,7 +659,33 @@ func (manager *EnclaveManager) getEnclavesWithoutMutex(
...
+	if !existsEnclave {
+		return nil, stacktrace.NewError("Enclave '%v' not found", enclaveIdentifier)
+	}
+
+	// filter idle enclaves because these were not created by users
+	if isIdleEnclave(*enclaveObj) {
```
mieubrisse: Hmm? What's an "idle enclave", not created by users? My gut says I'd be surprised if the enclave manager is hiding enclaves that exist

engine/server/engine/enclave_manager/enclave_manager.go
```diff
@@ -671,6 +712,15 @@ func getAllEnclavesFilter() *enclave.EnclaveFilters {
 	}
 }
 
+func getEnclaveFilter(enclaveIdentifier string) *enclave.EnclaveFilters {
```
mieubrisse: Is this technically "createGetEnclaveFilter"?

engine/server/engine/enclave_manager/enclave_manager.go
mieubrisse: EDIT: It looks like the `enclaveIdentifier` is being treated strictly as a UUID when the filter is created. I'm a little rusty on our terminology, but I seem to remember "identifier" = (uuid|short_uuid|name), so I _think_ calling it identifier here is incorrect perhaps?

engine/server/engine/enclave_manager/enclave_manager.go
```diff
@@ -253,6 +253,21 @@ func (manager *EnclaveManager) GetEnclaves(
 	return enclaveMapKeyedWithUuidStr, nil
 }
 
+func (manager *EnclaveManager) GetEnclave(
+	ctx context.Context,
+	enclaveIdentifier string,
```
mieubrisse: See comment below about the name "enclaveIdentifier" and what this semantically means

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
```diff
@@ -423,6 +423,9 @@ func (backend *DockerKurtosisBackend) getMatchingApiContainers(ctx context.Conte
 		docker_label_key.ContainerTypeDockerLabelKey.GetString(): label_value_consts.APIContainerContainerTypeDockerLabelValue.GetString(),
 		// NOTE: we do NOT use the enclave UUID label here, and instead do postfiltering, because Docker has no way to do disjunctive search!
 	}
+
+	// TODO: dockerManager.GetContainersByLabels lists all APIContainers and inspects them so the performance of any upstream call to retrieve even a single APIContainer will always scale with the number of enclaves
```
mieubrisse, line 427: Seems like this is the heart of the problem. Inspecting any API container requires listing all API containers, and then throwing away.

But I'm curious: why is this linear time? In the worst-case, shouldn't Docker do an O(n) operation over all containers regardless?

So my gut is saying that, rather than adding API surface area with a `GetEnclave` call (and requiring users to know the difference, and that GetEnclaves might be slower), a transparent performance improvement would be better here.

E.g. some ideas:
- Figuring out why GetContainersByLabels gets slower
- If the user asks for just 1 enclave ID, then we just go and fetch that directly (e.g. using a different codepath than with multiple)

mieubrisse (commented):
I have some questions/thoughts about this guy; didn't want to burn the requested 30m with those outstanding so pausing here with comments added

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
tedim52: `GetContainersByLabels` scales with n because under the hood it does a container list and inspect on all api containers. Ideally, we leverage filtering and ask Docker to only list and inspect the api container matching the enclave uuid we want to retrieve - but Docker doesn't have a way to do disjunctive search, hence the comment above `// NOTE: we do NOT use the enclave UUID label here, and instead do postfiltering, because Docker has no way to do disjunctive search!`

tedim52: More importantly, that comment actually isn't the heart of the problem. The heart of the problem is at `enclaveManager.getEnclavesWithoutMutex` which gets called every time `enclave inspect` happens. 

```
func (manager *EnclaveManager) getEnclavesWithoutMutex(
	ctx context.Context,
) (map[enclave.EnclaveUUID]*types.EnclaveInfo, error) {
	enclaves, err := manager.kurtosisBackend.GetEnclaves(ctx, getAllEnclavesFilter())
	if err != nil {
		return nil, stacktrace.Propagate(err, "Error thrown retrieving enclaves")
	}

	result := map[enclave.EnclaveUUID]*types.EnclaveInfo{}
	for enclaveId, enclaveObj := range enclaves {
		// filter idle enclaves because these were not created by users
		if isIdleEnclave(*enclaveObj) {
			continue
		}

		enclaveInfo, err := getEnclaveInfoForEnclave(ctx, manager.kurtosisBackend, enclaveObj)
		if err != nil {
			return nil, stacktrace.Propagate(err, "An error occurred getting information about enclave '%v'", enclaveId)
		}
		result[enclaveId] = enclaveInfo
	}
	return result, nil
}
```
The call `getEnclaveInfoForEnclave`here  - under the hood calls `kurtosisBackend.GetAPIContainers` - which as mentioned above retrieves all api containers. This means that if there are 5 enclaves, for each call we get all the 5 API containers from the docker backend, giving us O(n^2) behavior. The reason we make all 5 calls is because we only post filter for the single enclave we want at the api layer, after the engine has returned all enclaves.

The key fix in this PR was to avoid looping over `getEnclaveInfo` for all enclaves unnecessarily by adding `enclaveManager.getEnclaveWithoutMutex`.

tedim52: Regarding not increasing the api surface area/using a different codepath than with multiple:

The only api call that was added here was to the engine client - `engineClient.GetEnclave`. The api exposed to users + clients at the api layer didn't change at all. To get an enclave users still interface with `kurtosisCtx.GetEnclave` so I don't think users need to reason about anything differently, and ideally before or after this change users shouldn't have to reason about the performance.

It is possible to avoid adding `engineClient.GetEnclave` by changing `engineClient.GetEnclaves` to accept an extra "filters" arg like `engineClient.GetEnclaves(ctx, filters)` - if filters only requests one enclave then use the `enclaveManager.getEnclaveWithoutMutex(ctx, filters)` etc... Do you think thats a preferred api over?

tedim52: Discussed offline - we'll go with `engineClient.GetEnclaves(ctx, filters)` approach

engine/server/engine/enclave_manager/enclave_manager.go
tedim52: `create` is probably a better verb to describe whats happening here but I want to be consistent with all the other getFilter functions that already exist in this code e.g. `getEnclaveByEnclaveIdFilter`, `getApiContainerByEnclaveIdFilter`, etc.

engine/server/engine/enclave_manager/enclave_manager.go
tedim52: that check was copied from the previous get enclaves flow but it isn't needed anymore because in the new api users should only provide enclave uuids from enclaves they've created. 

will take the opportunity to improve the original comment for clarity though

engine/server/engine/enclave_manager/enclave_manager.go
tedim52: this is fixed by forcing only uuids to be provided to `GetEnclaves(ctx, uuids)`

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
tedim52: To avoid making a breaking change I kept the original `GetEnclaves` endpoint and added `GetEnclavesByUuids` endpoint

api/golang/engine/lib/kurtosis_context/kurtosis_context.go
```diff
@@ -193,8 +193,18 @@ func (kurtosisCtx *KurtosisContext) GetEnclaveContext(ctx context.Context, encla
...
+	if err != nil {
+		return nil, stacktrace.Propagate(err, "An error occurred creating an enclave context from the provided enclave info")
+	}
+	return enclaveCtx, nil
+}
+
 func (kurtosisCtx *KurtosisContext) GetEnclaves(ctx context.Context) (*Enclaves, error) {
-	response, err := kurtosisCtx.engineClient.GetEnclaves(ctx, &emptypb.Empty{})
+	response, err := kurtosisCtx.engineClient.GetEnclavesByUuids(ctx, &kurtosis_engine_rpc_api_bindings.GetEnclavesByUuidsArgs{
+		EnclaveUuids: []string{}, // retrieves all enclaves
```
mieubrisse, line 206: noicee

api/protobuf/engine/engine_service.proto
```diff
@@ -179,6 +181,14 @@ message StopEnclaveArgs {
   string enclave_identifier = 1;
 }
 
+// ==============================================================================================
+//                                       Get Enclaves
+// ==============================================================================================
+message GetEnclavesByUuidsArgs {
```
mieubrisse, line 187: There's already a `Get Enclaves` section above; is it expected that this isn't a part of it?

cli/cli/command_framework/highlevel/enclave_id_arg/enclave_id_arg.go
```diff
@@ -104,41 +105,45 @@ func getValidationFunc(argKey string, _ string, isGreedy bool) func(context.Cont
...
-				return stacktrace.Propagate(err, "Expected a value for non-greedy arg '%v' but didn't find one", argKey)
-			}
-			enclaveIdentifiersToValidate = []string{enclaveIdentifier}
 		}
 
-		enclaves, err := kurtosisCtx.GetEnclaves(ctx)
-		if err != nil {
-			return stacktrace.Propagate(err, "An error occurred getting enclaves, which is necessary to check if the enclaves exist")
-		}
+		// only validate enclave identifer if this is a greedy enclave identifier arg
```
mieubrisse: oh huh, why are we only doing it if there are more than 1 enclave??

cli/cli/command_framework/highlevel/enclave_id_arg/enclave_id_arg.go
```diff
@@ -104,41 +105,45 @@ func getValidationFunc(argKey string, _ string, isGreedy bool) func(context.Cont
...
+					if len(matches) > validShortenedUuidOrNameMatches {
+						return stacktrace.NewError("Found multiple matching uuids '%v' for name '%v' which is ambiguous", enclaveInfosToUuidsStr(matches), enclaveIdentifier)
+					}
+					continue
+				}
+				return stacktrace.NewError("No enclave found for identifier '%v'", enclaveIdentifier)
+			}
+		} else if len(enclaveIdentifiersToValidate) == 1 {
+			// enclave identifier is validated in GetEnclave, so checking no err is enough
+			_, err = kurtosisCtx.GetEnclave(ctx, enclaveIdentifiersToValidate[0])
```
mieubrisse: wait question - I thought `GetEnclave` takes in UUIDs. so wouldn't shortened UUIDs and names (which are valid identifiers but not valid UUIDs) fail here?

cli/cli/command_framework/highlevel/enclave_id_arg/enclave_id_arg.go
```diff
@@ -104,41 +105,45 @@ func getValidationFunc(argKey string, _ string, isGreedy bool) func(context.Cont
...
-				continue
+				if matches, found := enclavesByName[enclaveIdentifier]; found {
+					if len(matches) > validShortenedUuidOrNameMatches {
+						return stacktrace.NewError("Found multiple matching uuids '%v' for name '%v' which is ambiguous", enclaveInfosToUuidsStr(matches), enclaveIdentifier)
+					}
+					continue
+				}
+				return stacktrace.NewError("No enclave found for identifier '%v'", enclaveIdentifier)
+			}
+		} else if len(enclaveIdentifiersToValidate) == 1 {
```
mieubrisse: ah, you answer this here. I'd recommend putting this case first with a comment, so it's clearer what's going on

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
```diff
@@ -445,7 +444,9 @@ func (backend *DockerKurtosisBackend) getMatchingApiContainers(ctx context.Conte
 			apiContainer.GetLabels(),
 			apiContainer.GetStatus(),
 			apiContainer.GetHostPortBindings(),
-			bridgeNetworkIpAddress,
+			// Bridge network IP address will be filled in later if this APIContainer passes the post-filter.
```
mieubrisse, line 447: should this param even exist anymore then, if we're expecting that that info will get filled in later?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
```diff
@@ -466,7 +467,16 @@ func (backend *DockerKurtosisBackend) getMatchingApiContainers(ctx context.Conte
 			}
 		}
 
-		allMatchingApiContainers[containerId] = apicObj
+		bridgeNetworkIpAddress, err := backend.dockerManager.GetContainerIP(ctx, backend.dockerManager.GetBridgeNetworkName(), containerId)
```
mieubrisse, line 470: we already have the bridge network IpAddress from earlier; why do we need to get it again?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
```diff
@@ -551,6 +561,23 @@ func getApiContainerObjectFromContainerInfo(
...
+	bridgeNetworkIpAddress string,
+) (*api_container.APIContainer, error) {
+	result := api_container.NewAPIContainer(
+		apicObj.GetEnclaveID(),
+		apicObj.GetStatus(),
+		apicObj.GetPrivateIPAddress(),
+		apicObj.GetPrivateGRPCPort(),
+		apicObj.GetPublicIPAddress(),
+		apicObj.GetPublicGRPCPort(),
+		net.ParseIP(bridgeNetworkIpAddress),
```
mieubrisse: ParseIP has a silent error condition:
```
		bridgeNetworkIpAddr = net.ParseIP(bridgeNetworkIpAddrStr)
		if bridgeNetworkIpAddr == nil {
			return nil, stacktrace.NewError("Couldn't parse bridge network IP address string '%v' to an IP", bridgeNetworkIpAddrStr)
		}
```

enclave-manager/server/server.go
```diff
@@ -298,6 +298,31 @@ func (c *WebServer) GetEnclaves(ctx context.Context, req *connect.Request[emptyp
 	}
 	return resp, nil
 }
+
+func (c *WebServer) GetEnclavesByUuids(ctx context.Context, req *connect.Request[kurtosis_enclave_manager_api_bindings.GetEnclavesByUuidsRequest]) (*connect.Response[kurtosis_engine_rpc_api_bindings.GetEnclavesResponse], error) {
```
mieubrisse, line 302: a bit confused: I thought the result of our last chat was that you wanted to add a filters parameter to GetEnclaves, rather than adding a new endpoint? but I trust you know what you're doing

engine/server/engine/enclave_manager/enclave_manager.go
```diff
@@ -253,6 +253,27 @@ func (manager *EnclaveManager) GetEnclaves(
 	return enclaveMapKeyedWithUuidStr, nil
 }
 
+func (manager *EnclaveManager) GetEnclavesByUuid(
```
mieubrisse, line 256: Same question as above - I thought the outcome of our last call was adding a filters param on GetEnclaves, rather than adding more endpoints? (I could be totally misremembering here)

mieubrisse (approved):

engine/server/engine/enclave_manager/enclave_manager.go
tedim52: discussed offline

cli/cli/command_framework/highlevel/enclave_id_arg/enclave_id_arg.go
tedim52: `GetEnclave` will call `GetExistingAndHistoricalIdentifiers` which resolves enclave uuids to names and shortened identifiers

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
tedim52: I considered this but `getApiContainerObjectFromContainerInfo` is used by two other functions that actually do populate the `bridgeNetworkIpAddr`. Instead of changing those functions, I went with documenting why it wasn't being populated here.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
tedim52: AFAICT its not being retrieved earlier in this function - can you direct me to what lines are you are referring to where its being called earlier?

cli/cli/command_framework/highlevel/enclave_id_arg/enclave_id_arg.go
tedim52: answered in other comment

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend_api_container_functions.go
tedim52: okay for this to be nil here, as the api container can be in states where its not connected to the bridge network. added a comment clarifying in the api container object

PR #2777, opened 2025-07-09
fix: reduce logs agg container timeout

tedim52:
## Description
`clean -a` was taking a long time because of this time out. 

YES


leovct (approved):
lgtm

PR #2806, opened 2025-09-18
fix: use rootless docker for starting the engine

qu0b:
## Description
Fix Kurtosis engine startup failure when using rootless Docker with custom socket path.

When Docker is configured to use a rootless user socket location (e.g., /run/user/1000/docker.sock for rootless Docker), the Kurtosis engine container fails to start because it attempts to mount the hardcoded /var/run/docker.sock path, which doesn't exist. This causes the engine to fail with a "permission denied" error when trying to connect to the Docker daemon.

This change detects if the DOCKER_HOST environment variable is set to a Unix socket path, validates that the socket file exists, and uses it for mounting into the engine container instead of the hardcoded default. The socket is still mounted to /var/run/docker.sock inside the container to maintain compatibility with the existing engine code.

## REMINDER: Tag Reviewers, so they get notified to review
@tedim52


tedim52:
Hey @qu0b was OOO towards end of last week. Will try and get around to review this this week.

qu0b:
@tedim52 I ended up removing my docker rootless setup so im not sure if this will work across devices.

If you want to reproduce this install docker on Ubuntu with docker-rootless-setup

barnabasbusa:
are we good to close this pr?

PR #2818, opened 2025-10-22
fix: don't publish udp ports

tedim52:
## Description
Users on Docker Desktop 4.42.0> been experiencing this error: `Caused by: No host machine port binding was specified for Docker port '9001/udp' which corresponds to port spec with num '9001' and protocol 'UDP'`. because Docker Desktop is unsuccessfully publishing UDP ports since 4.41.2 https://github.com/docker/desktop-feedback/issues/191. 

Users have had to resort to downgrading [Docker Desktop or switching to Orbstack](https://github.com/ethpandaops/ethereum-package/issues/1075#issuecomment-3238121736). This PR creates a workaround inside by opting to not publish UDP ports. UDP ports will still be exposed on the container and accessible by other services in the network, but they will not be published to the host machine. 

This should be okay for users of the https://github.com/ethpandaops/ethereum-package as it's not common to access a UDP port from localhost however it could cause a breaking change for any users depending on a UDP port exposed to localhost for access to a service inside an enclave.

YES

## References:
https://github.com/docker/desktop-feedback/issues/191
https://github.com/ethpandaops/ethereum-package/issues/1075#issuecomment-3238121736


barnabasbusa (approved):
send it

PR #2819, opened 2025-10-23
fix: kurtosis service update with wait=None

### LINKED ISSUE #2813: PortSpec issue on service update
### What's your CLI version?

1.11.1

Having a service with a `PortSpec` that includes `wait=None` allows the service to start even if the specified port is not available.
However, this setting seems to be ignored when performing a `kurtosis service update` command, and the service will fail to start if the port is not available.

Here a really small PoC so you can see the exact issue:
https://github.com/xavier-romero/scripts-pub/tree/main/kurtosis-poc/update-port-issue

### Desired behavior

The update should honor the `wait=None` on the PortSpec and the service should be able to be updated.

leovct:
## Description

Closes #2813 

Before the fix - the port spec doesn't contain `wait=None` and the operation fails:

```bash
$ kt --cli-log-level=debug service update test service-1 --image nginx:latest
...
DEBU[2025-10-23T14:39:59Z] Update service starlark:
def run(plan):
	plan.add_service(name = "service-1", config = ServiceConfig(image="nginx:latest", ports={"test": PortSpec(number=2000)}, entrypoint=["/docker-entrypoint.sh"], cmd=["nginx", "-g", "daemon off;"], env_vars={"PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin","NGINX_VERSION": "1.28.0","NJS_VERSION": "0.8.10","NJS_RELEASE": "1~bookworm","PKG_RELEASE": "1~bookworm","DYNPKG_RELEASE": "1~bookworm"}))
...
# service update fails
```

After the fix - the port spec contains `wait=None` and the operation succeeds:

```bash
$ kt --cli-log-level=debug service update test service-1 --image nginx:latest
...
DEBU[2025-10-23T14:59:31Z] Update service starlark:
def run(plan):
	plan.add_service(name = "service-1", config = ServiceConfig(image="nginx:latest", ports={"test": PortSpec(number=2000,wait=None)}, entrypoint=["/docker-entrypoint.sh"], cmd=["nginx", "-g", "daemon off;"], env_vars={"NGINX_VERSION": "1.28.0","NJS_VERSION": "0.8.10","NJS_RELEASE": "1~bookworm","PKG_RELEASE": "1~bookworm","DYNPKG_RELEASE": "1~bookworm","PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}))
...
# service update succeeds
```

Tested against Xavi's PoC:

```bash
kt run --enclave test .
kt --cli-log-level=debug service update test nginx --image nginx:latest
```

YES

### DIFF
--- a/api/golang/core/lib/services/service_config_builder.go
+++ b/api/golang/core/lib/services/service_config_builder.go
@@ -60,8 +60,12 @@ func portToStarlark(port *kurtosis_core_rpc_api_bindings.Port) string {
 	if port.GetTransportProtocol() != kurtosis_core_rpc_api_bindings.Port_TCP {
 		starlarkFields = append(starlarkFields, fmt.Sprintf(`transport_protocol="%s"`, port.GetTransportProtocol().String()))
 	}
-	if port.GetMaybeWaitTimeout() != "" {
-		starlarkFields = append(starlarkFields, fmt.Sprintf(`wait="%s"`, port.GetMaybeWaitTimeout()))
+	waitTimeout := port.GetMaybeWaitTimeout()
+	if waitTimeout == "" {
+		// When wait=None is set in Starlark, the proto field MaybeWaitTimeout is an empty string.
+		starlarkFields = append(starlarkFields, `wait=None`)
+	} else {
+		starlarkFields = append(starlarkFields, fmt.Sprintf(`wait="%s"`, waitTimeout))
 	}
 	return fmt.Sprintf("PortSpec(%s)", strings.Join(starlarkFields, ","))
 }
--- a/docs/docs/cli-reference/service-update.md
+++ b/docs/docs/cli-reference/service-update.md
@@ -37,3 +37,6 @@ kurtosis service update my-enclave test-service \
 This command replaces the existing service with a new container using the updated configuration. The service will be briefly stopped and restarted as part of this process.
 :::
 
+:::note Port wait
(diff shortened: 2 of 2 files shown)


tedim52 (approved):

tedim52:
Thanks for the fix! I verified it works on my end. I actually noticed a separate issue while reviewing your PR

Right now, if `PortSpec(number=80, wait="25s")` is on an `add_service` - on a corresponding `service update`, the wait="25s" isn't respected. This is bc we store and retrieve Port information from a service container via container labels on  and we don't store the wait information currently on the label.

https://github.com/kurtosis-tech/kurtosis/blob/f87a91ead5306196818ff3c39d2bcfbacc0826cd/container-engine-lib/lib/backend_impls/docker/object_attributes_provider/docker_port_spec_serializer/port_spec_serializer.go#L213

What happened before is on a service update is `PortSpec(number=80, wait="25s")` just got turned into PortSpec(number=80) and the default wait would get used wait="15s".

What will happen now is PortSpec(number=80, wait="25s") will get turned into PortSpec(number=80, wait=None). 

Your change is correct though but while you're at it do you mind highlighting in the `service update `documentation: https://docs.kurtosis.com/service-update/? Something like a note that calls out that wait for PortSpec doesn't get respected on a service update/the original wait check is removed/turned into a `wait=None`. 

The proper fix would be too store wait information on the container and and update the serialize / deserialize logic but that's out of scope for this PR : )

tedim52 (approved):

PR #2845, opened 2025-11-27
feat: Add support for mounting devices

dzobbe:
## Description

This PR adds device support to the Docker backend, bringing feature parity with the Kubernetes backend. Previously, the `devices` field in `ServiceConfig` was only supported when running on Kubernetes. This change enables services running in Docker containers to access host devices (e.g., `/dev/tpm0`, `/dev/gpu0`, etc.).
It can be configured via an extended ethereum-package repo 
https://github.com/ethpandaops/ethereum-package/pull/1251/files

### Changes Made

1. **Added `devices` field** to `CreateAndStartContainerArgs` and `CreateAndStartContainerArgsBuilder` in the Docker manager
2. **Added `WithDevices()` method** to the builder pattern for setting device paths
3. **Updated `getContainerHostConfig()`** to accept and process device paths, converting them to Docker's `DeviceMapping` format
4. **Added `convertDevicesToDockerDeviceMapping()` helper function** that maps device paths to Docker's device format with read-write-mknod permissions
5. **Updated `start_user_services.go`** to extract devices from service config and pass them through to the Docker container creation

### How It Works

When a service config includes a `devices` list (e.g., `devices = ["/dev/tpm0"]`), the Docker backend now:
- Extracts the device paths from the service config
- Converts them to Docker's `container.DeviceMapping` format
- Mounts each device with the same path in the container as on the host
- Grants read-write-mknod permissions (`rwm`) to the container

### Motivation

This change enables services that require access to host hardware devices (such as TPM chips for attestation, GPUs, or other specialized hardware) to function correctly when running in Docker-based Kurtosis enclaves. Previously, such services would fail with "device not found" errors even when the device was specified in the Starlark service config.

### Example Usage

service_config = ServiceConfig(
    name = "tpm-service",
    image = "my-image",
    devices = ["/dev/tpm0"],  # Device is now accessible in Docker containers
    # ... other config
)## REMINDER: Tag Reviewers, so they get notified to review

YES


barnabasbusa:
were you able to compile it locally? 

Looks like the CI fails with build errors. Can you try to debug those?

dzobbe:
Looks like now the checks have passed

tedim52:
Thanks for the contribution @dzobbe !

metrics-library/golang/go.mod
```diff
@@ -4,6 +4,8 @@ go 1.23.0
 
 toolchain go1.23.7
 
+replace github.com/kurtosis-tech/kurtosis/enclave-manager/api/golang => ../../enclave-manager/api/golang
```
tedim52: Think we can remove this?

container-engine-lib/lib/backend_impls/kubernetes/kubernetes_kurtosis_backend/user_services_functions/start_user_services.go
```diff
@@ -372,6 +373,65 @@ func createStartServiceOperation(
...
 			}
 		}
+
+		// Add device volumes if devices are specified
+		if len(devices) > 0 {
+			for _, devicePath := range devices {
+				deviceType := apiv1.HostPathCharDev
+				// Try to detect if it's a block device (common patterns: /dev/sd*, /dev/nvme*, /dev/vd*)
+				// For TPM devices (/dev/tpm*), they are character devices
+				volumeName := "device-" + strings.ReplaceAll(strings.TrimPrefix(devicePath, "/dev/"), "/", "-")
```
tedim52: can we prefix the volume name with `kurtosis`? 

Still need to test this locally - if multiple services mount the same device path - the volume mounted here would get reused / won't conflict with other volume mounts? (i'd assume this is the case for docker but just making sure)

container-engine-lib/lib/backend_interface/objects/service/service_config.go
```diff
@@ -293,6 +297,10 @@ func (serviceConfig *ServiceConfig) GetTtyEnabled() bool {
 	return serviceConfig.privateServiceConfig.TtyEnabled
 }
 
+func (serviceConfig *ServiceConfig) GetDevices() []string {
```
tedim52: very nit but can we move this function to be located where the other `With...` functions are below the `Build` and constructor for consistency?

tedim52 (changes requested):
Looks great! Just a few minor comments.

dzobbe:
I think I addressed the comments

tedim52 (approved):

PR #2890, opened 2026-02-16
fix: replace netstat with /proc/net/tcp for port availability checks

barnabasbusa:
The fluent-bit debug container (distroless:debug) does not include netstat, causing the logs collector port check to always fail on k8s. Use /proc/net/tcp which works on any Linux container without tool deps.

When it broke: Commit be932f74a ("impl k8s fluentbit collector availability check", Feb 18, 2025), merged via PR #2653 at commit 05e38135c ("feat:
  fluent bit logs collector in k8s backend", Feb 22, 2025). This has been broken since day one of the k8s logs collector feature - it was never
  working.

  Root cause: The port availability check runs netstat -anp tcp | grep LISTEN | grep 9713 via kubectl exec inside the fluent/fluent-bit:4.0.0-debug
  container. That image is based on Google's distroless:debug, which has a minimal busybox that does not include netstat. The engine and API
  containers are Alpine-based and have busybox netstat, so their checks work fine.

  The fix: Replaced the netstat command with a /proc/net/tcp check in both the Kubernetes and Docker backends:
  grep -i ':25F1 ' /proc/net/tcp /proc/net/tcp6 2>/dev/null | grep -qi ' 0A '

  - Reads the Linux proc filesystem directly (always available, zero tool dependency)
  - 25F1 is the hex representation of the port number
  - 0A is the TCP LISTEN state in /proc/net/tcp

### DIFF
--- a/container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/shared_helpers/shared_helpers.go
+++ b/container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/shared_helpers/shared_helpers.go
@@ -382,10 +382,10 @@ func WaitForPortAvailabilityUsingNetstat(
 	maxRetries uint,
 	timeBetweenRetries time.Duration,
 ) error {
+	portHex := fmt.Sprintf("%04X", portSpec.GetNumber())
 	commandStr := fmt.Sprintf(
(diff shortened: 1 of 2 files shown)


barnabasbusa:
Closing as it is now part of #2891

PR #2928, opened 2026-03-05
feat: add --resource-check CLI flag to skip resource validation

mattevans:
## Description

- Adds a `--resource-check` CLI flag (default `true`) that controls whether `GetAvailableCPUAndMemory` is called during the validation phase.
- Disable with `--resource-check=false` to skip the Docker API call which takes multiple seconds.
- Plumbs the flag through proto, binding constructors, enclave context, runner, and validator.

YES


core/server/api_container/server/api_container_service.go
```diff
@@ -160,6 +160,7 @@ func (apicService *ApiContainerService) RunStarlarkScript(args *kurtosis_core_rp
 	nonBlockingMode := args.GetNonBlockingMode()
 	downloadMode := convertFromImageDownloadModeAPI(ApiDownloadMode)
 	shouldExecuteInParallel := args.GetParallel()
+	scriptResourceCheck := args.ResourceCheck == nil || args.GetResourceCheck()
```
tedim52: curious about the naming scriptResourceCheck here? I'm assuming this is variable indicates whether or not the resource check will happen. If so can we do  `shouldCheckResources` or something similar to match other variable names.

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -558,6 +558,9 @@ func (backend *DockerKurtosisBackend) DisconnectReverseProxyFromNetwork(ctx cont
 }
 
 func (backend *DockerKurtosisBackend) GetAvailableCPUAndMemory(ctx context.Context) (compute_resources.MemoryInMegaBytes, compute_resources.CpuMilliCores, bool, error) {
+	if !isResourceInformationComplete {
```
tedim52: what's this check doing/is it needed? It's always set to true up above so this branch would never fire:

https://github.com/mattevans/kurtosis/blob/c1e9c72b1f416f4da9cf5f4b597d3eea07c4761b/container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go#L44

tedim52 (approved):
approved but 2 minor comments

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
mattevans: @tedim52 Good catch, thank you! `isResourceInformationComplete` is hardcoded to `true` in the Docker backend so the branch was dead code. I've removed the constant and the unreachable check, and now just return `true` directly.

core/server/api_container/server/api_container_service.go
mattevans: Updated, thanks!

PR #2937, opened 2026-03-11
fix: fallback to registry key without https:// prefix in docker auth lookup

kayano:
## Description
  Problem

  When using `docker login ghcr.io`, docker stores the registry credentials in `~/.docker/config.json` using the bare hostname as the key:

```json
  {
    "auths": {
      "ghcr.io": { "auth": "..." }
    }
  }
```

  However, `GetAuthFromDockerConfig` normalizes the registry hostname by prepending `https://` before looking it up in the auths map (e.g. "https://ghcr.io"). This causes a key mismatch — credentials are present but never found, and the
  image pull proceeds unauthenticated and fails.

  This affects any registry where docker login stores the key without a protocol prefix (GHCR, ECR, GCR, etc.). Docker Hub is unaffected because it is stored as "https://index.docker.io/v1/" which already includes the prefix.

  The same bug affects `docker_config_storage_creator` — when building the credentials volume for the engine container, it calls `GetAuthFromDockerConfig` for each registry key, so GHCR credentials are silently skipped and never written to
   the volume.

  Fix

  Add a fallback in `GetAuthFromDockerConfig` that retries the auths map lookup without the `https://` prefix if the initial lookup fails. This is a minimal, targeted change that doesn't affect the existing behaviour for Docker Hub or any
  registry already stored with the `https://` prefix.

  Testing

  Added `TestGetAuthConfigForRepoBase64AuthWithoutHttpsPrefix` which covers the exact failing scenario: config has "ghcr.io" as the key (as produced by docker login ghcr.io), pulling `ghcr.io/my-org/my-image:latest`.

@kurtosis-tech/engineers, @tedim52, @mieubrisse

NO


tedim52:
tested locally and works well! LGTM thank you for the change

tedim52 (approved):

PR #3122, opened 2026-05-21
feat(cli): add `kurtosis otel` command + stamp enclave name on logs

samcm:
Adds `kurtosis otel start/stop` (Docker-only), modeled on `grafloki`: it runs an OpenTelemetry collector and a single-node ClickHouse as engine-managed side containers and reconfigures the engine's Vector logs-aggregator to also export logs to the collector, exposing OTLP and ClickHouse on the host so they're reachable from inside every enclave. It also stamps the enclave name onto user-service logs (previously only the enclave UUID was present) so logs and traces can be tenanted and queried by enclave name.


container-engine-lib/lib/backend_impls/docker/object_attributes_provider/object_attributes_provider.go
```diff
@@ -37,6 +37,7 @@ type DockerObjectAttributesProvider interface {
 		restAPIPortSpec *port_spec.PortSpec,
 	) (DockerObjectAttributes, error)
 	ForEnclave(enclaveUuid enclave.EnclaveUUID) (DockerEnclaveObjectAttributesProvider, error)
```
tedim52, line 39: doesn't seem like `ForEnclave` is being used anymore with this change, if so lets remove it?

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/backend_creator/backend_creator.go
```diff
@@ -236,6 +238,7 @@ func getDockerKurtosisBackend(
 	// If running within the API container context, detect the network that the API container is running inside
 	// so, we can create the free IP address trackers
 	enclaveFreeIpAddrTrackers := map[enclave.EnclaveUUID]*free_ip_addr_tracker.FreeIpAddrTracker{}
+	enclaveNames := map[enclave.EnclaveUUID]string{}
```
tedim52: it seems like the only purpose of `enclaveNames` is for `backend.getEnclaveName` is to look up the enclave name but we can just make a direct call to enclave network to grab enclave name instead of using this cache? it doesnt seem that expensive and we're also not updating this cache on the addition of new enclaves so we there's a fallback to check the network anyways in the current diff

container-engine-lib/lib/backend_impls/docker/docker_kurtosis_backend/docker_kurtosis_backend.go
```diff
@@ -266,6 +280,22 @@ func (backend *DockerKurtosisBackend) StartRegisteredUserServices(ctx context.Co
 	return successfullyStartedService, failedService, nil
 }
 
+func (backend *DockerKurtosisBackend) getEnclaveName(ctx context.Context, enclaveUuid enclave.EnclaveUUID) (string, error) {
```
tedim52: can we get rid of enclaveName and just always use backend.getEnclaveNetworkByEnclaveUuid?

tedim52 (approved):

container-engine-lib/lib/backend_impls/docker/object_attributes_provider/object_attributes_provider.go
samcm: Have replaced with `ForEnclaveWithName` in this PR, but `ForEnclave` seems to be used elsewhere :)

barnabasbusa (approved):

